PR #5, opened 2017-06-08
Removing distribution policy text from README.md

eclare108213:


eclare108213:
I'm not doing things in the right order...
LANL policy for copyright statements seems pretty stringent; we need the Distribution Policy in a separate document from the copyright and license.

PR #7, opened 2017-06-08
Testscriptswolf

eclare108213:
There are two different things in this pull request, changes to the testing scripts for wolf and the change to the README.md file, which I pushed (on the wrong branch) by mistake.  There's an artificial conflict in the README.md file (might be whitespace?) which shouldn't matter; I haven't figured out how to remove the commit or make the pull request only include the changes for wolf.  Fixing this looks pretty complicated.  So maybe just let it all go through together?  And then maybe we should provide instructions on what NOT to do...  (I'm learning)


apcraig:
So, I had to resolve the conflict before I could pull.  I resolved it on the github page directly and did the commit there.  It seems to have worked.  Will now execute the PR.

PR #8, opened 2017-06-15
Rm restart flags from icepack

eclare108213:
This pull request removes references to restart flags from Icepack.  In doing so, the order of initialization for the bgc was changed.  Nicole tested these changes with the bgc turned on and suggested additional mods to ice_init_column.F90, which are included here.  I tested the standard configuration (bgc turned off) using the smoke test and also a manual restart test.

Note:  Nicole provided the compile flags and ice_in namelist for testing bgc in an email 6/14/2017.  I've invited Nicole to CICE-Consortium -- I'd like her to review these changes, since they are for the bgc.  We could also invite Dave, either as reviewer or assignee, so that he can also experience submodule-land...

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "icepack"]
 	path = icepack
-	url = https://github.com/eclare108213/Icepack.git
-	branch = icepack_initial_driver
+	url = https://github.com/cice-consortium/Icepack.git
+	branch = master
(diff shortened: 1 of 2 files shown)


eclare108213:
I realized in the middle of the night that I probably also need to do a separate pull request for Icepack, and that one probably needs to be satisfied first.  What's the process for this?  I'll make a pull request now.

apcraig:
When there is a dependence between Icepack and CICE, the proper order is probably to do the pull request for Icepack, then update the CICE pull request to point to the new version of Icepack.  I will note that in my review.

eclare108213:
Are the changes to .gitmodules in the  99b50db commit what's needed for this PR?

PR #10, opened 2017-07-11
Putting overview info in README.md.  Also updating my fork to the upstream Consortium version of CICE.

eclare108213:
README.md:  The original content was moved (separately) into LICENSE.pdf.  New content is a draft with the kinds of info suitable for a 'front page'.  @dabail10 @duvivier please comment on whether this is the right info, what's missing, etc.  

Icepack:  I'm not at all sure whether we want to merge this change into the consortium repo.  @apcraig are the differences here what you would expect to see?  Note that the consortium version of icepack is c1996be but the one in my fork is c0582fb, in spite up my having updated my fork from the upstream repo and doing 'git add icepack', commit, push.  I'm worried that the Icepack linkage in my fork is messed up.

and this should have been done on a branch rather than master...


duvivier:
Elizabeth,
I think that this mostly looks good and I'm happy for you to commit it. I have a few comments:

Eventually I think we want the documentation link to point to here:
https://cice-consortium.github.io/CICE/
We would also provide this link in the wiki. Alternately, from the READMD.md we could just have the links for the documentation, tests, data, etc. go to the wiki page where the information can be found rather than directly to the link.

The link won't work until I have finished the conversion of the tex files to sphinx and we've completed the pull request. However, in the https://github.com/CICE-Consortium/CICE/tree/master/doc directory you currently are pointing to we will have the sphinx *.rst files under the source directory but since we want the CICE documentation to be an easily changeable document that we update as we modify the master branch, I was under the impression that we wouldn't frequently make pdfs of the documentation. However, we *will* make a pdf of the current documentation with each release so we get a "snapshot" of the documentation at the time of release. 

One thing to consider if if we want to keep all the PDFs of supporting literature in the doc/PDF/ directory. I've been making links to each paper in the sphinx documentation so I'm not sure this PDF directory is necessary. 

Cheers,
Alice

eclare108213:
I agree it's better to send users to the wiki for documentation rather than pointing to the doc directory, so I removed that reference.

I definitely think we should remove all of the pdfs of papers in doc/PDF that are easily obtainable.  Let's keep those that are technical documents, i.e. not part of the publicly available literature, but they could be put in our data repository with links from the other documentation, rather than keeping them in the code repos.  I won't remove the pdfs as part of this pull request, but I'd like us to do this soon.

I will not merge this pull request, especially since the Icepack part of it is questionable.  I'd like @apcraig to weigh in on that.
Thank you!

PR #13, opened 2017-07-26
updating Icepack version to c1996be

eclare108213:


apcraig:
I think I understand why we're showing a diff when there isn't one.  See

https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

I will issue the PR and then see what happens.

eclare108213:
Autosquash. I like it maybe we could do that as a matter of course?
e

apcraig:
I was thinking more rebase before doing PRs.

I think all autosquash does is turn multiple actual commits into a 
single commit in the PR.  I think the user needs to rebase their branch, 
review changes, and test before doing the PR.  To rebase, I think it's

either

git pull --rebase upstream master

or

git fetch upstream
git rebase upstream/master

PR #20, opened 2017-08-10
PR-template for consortium

duvivier:
[Enter Description of the changes in this Pull Request. It should be 
enough information for someone not following this development to 
understand. Please also update the documentation, if necessary, in 
doc/source/rst and indicate below if you need to have the gh-pages html regenerated.]

Summary:  Adding a template for future pull requests for the consortium
PR Creator: Alice DuVivier
Developer(s): Alice DuVivier
Results (bit for bit, roundoff, climate changing): n/a
Details:
Code review:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -8,5 +8,6 @@ PR Creator:
 Developer(s):
 Updated documentation (Y/N):
 Results (bit for bit, roundoff, climate changing):
-Details:
 Code review: 
+Other Relevant Details:
+
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,10 +1,4 @@
-[Enter Description of the changes in this Pull Request. It should be 
-enough information for someone not following this development to 
-understand. Please also update the documentation, if necessary, in 
-doc/source/rst and indicate below if you need to have the gh-pages html regenerated.]
-
-Summary:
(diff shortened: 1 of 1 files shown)


apcraig:
Alice, you should be able to update your PR-template branch in your fork and those changes will automatically become part of the PR.  I suggest the following, switch the order of details and code review in the template (details should be last as that can be several paragraphs), specify bit-for-bit under results (as that is the case for this PR), specify "none" under Code Review unless you feel someone contributed, and then add one or two sentences (enough for this PR) regarding details.  You can say in details, for instance, something like "Added PR template under .github/PULL_REQUEST_TEMPLATE for general use in the future.  This template will evolve as needed over time" or something like that.  If we are going to have a template, we should almost require that all fields are filled in.  thanks.

duvivier:
Summary: Adding a template for future pull requests for the consortium
PR Creator: Alice DuVivier
Developer(s): Alice DuVivier
Code review: none
Other relevant details: Added PR template under .github/PULL_REQUEST_TEMPLATE for general use in the future. This template will evolve as needed over time.

apcraig:
Thinking about this a little more, I have a couple more comments.  

To make this work best with github, I think we should remove the first paragraph and then change "Summary:" to [Remove this and add a short summary line].  The point is that the first line of the PR template is how the PR is defined.  

Also, I think we can remove "PR Creator:".  github takes care of that automatically.

apcraig:
Thanks Alice!  Sorry for the extra iterations.

PR #21, opened 2017-08-25
HTML code for CICE-Consortium

duvivier:
The initial conversion of code from the CICE LaTeX document to HTML code for github. Has not yet been broken into CICE and Icepack versions.

Developer(s): Alice DuVivier
Updated documentation (Y/N): Y
Results (bit for bit, roundoff, climate changing): b4b
Code review: 
Other Relevant Details:


eclare108213:
It looks like these changes will be put in the gh-pages branch of the CICE repo rather than master; is that right?  Also, I think we will link to them from the wiki page or README.md.  What would that link be?  I'm clicking around in the gh-pages branch of https://github.com/duvivier/CICE and not finding a starting place that gives me the table of contents in readable form, for instance.  Where is that?

duvivier:
Yes, these changes (the html only) will go in the gh-pages branch. The gh-pages branch is only for the html code and does not contain the documentation source code (rst files) and those will be in the master branch. 

I did modify the README.md file in the second pull request I submitted for the master branch in which I link to the documentation. If you look at that second pull request you should see the line I added to the README.md that should work with the master branch as long as these html pages are in the gh-pages branch. 

I agree we may also want to link from the wiki, but that will also be through the master branch. I did not include it in my second pull request because we hadn't discussed that yet. 

If you want to see what the documentation pages will look like this is the version from my fork: https://duvivier.github.io/CICE/
It won't look like this from the consortium page until we commit the consortium URL to the README.md on the master branch and have the gh-pages linked.

eclare108213:
This is a good starting place.  We will modify/fix as needed.

PR #24, opened 2017-09-14
Add Hooks to CDash

mattdturner:
This pull request includes commits that allow for a regression test to be run, and the results to be submitted to the CDash website.  
Developer(s): Matthew Turner
Results (bit for bit, roundoff, climate changing): N/A; no changes to code or how tests are run.
Code review: 
Other Relevant Details:
- The current setup submits the CTest results to the myCICE project on CDash.  Once we feel that this setup is ready to be used, we can create a new CICE project.
- CTest has the ability to hook into git and checkout a specific branch and pull the latest updates.  However, on the servers that I used to test this, the version of cmake/ctest was too old to include this feature.
- In order to wait for all test jobs to complete, I have the script setup to monitor for the .o* files that are generated by PBS once a job completes.  I am not sure if other queue managers create similar files (or what the filenames are).  As such, the current setup might be PBS-specific, but she be easily adaptable for other job managers.
- I have not updated the documentation yet.  I was not sure where to document this, as I doubt we want all users to be running the regression and submitting to the CDash dashboard.
- To test this out, all that needs to be done is to run the "run.regression" script.  This will do the following:
1. Check out the master branch
2. Pull any code changes (my thoughts were, if running nightly tests we always want the most recent master)
3. Run create.case for the base_suite.  This compiles and submits all tests
4. Monitor for the completion of all jobs
5. Parse results
6. Run CTest and submit to CDash.
7. Delete the regression suite directory.


apcraig:
I'm a little concerned about the reliance on cmake and specific versions.  Is cmake readily available on most platforms?  If not, do we have alternatives.  It sounds like we need cmake to push to cdash.

I think we need to find another way to check whether the job is complete.  What's there is OK for now, but it's not ideal.  Lets continue to think about that.

We might want to hide the run.regression file from users.  Maybe it should live under scripts right now?  Also, I think we should have a different name, run.regression implies many things.  How about run.suite.  I will suggest that in the PR review.

apcraig:
I'm also wondering about the CTestTestfile.cmake file.  We do not want to hardwire the tests that are reported.  We might have different suites on different machines to improve coverage and we might have different levels of tests (nightly, weekly, prerelease, etc) and we want to be able to publish all test results that are generated at any given time.  Is the CTestTestfile.cmake hardwiring the test report?  If so, then we should come up with a way to specify the test list such that our scripts and CTest are automatically on the same page all the time via a single set of lists or some scripting.  We don't want to have to maintain our test lists in two places.

mattdturner:
I renamed run.regression and moved to run.suite.  I agree that we do not want it visible to most users.

Regarding the fact that CTestTestfile.cmake is hardcoded.  I am not sure of a better way to do it.  I understand that we will eventually have different suites that run on different servers.  Right now, the only way that I can think of to handle the variability is to have CTestTestfile.cmake built dynamically when running create.case.  I thought about having it just go through the results.log file line-by-line, but that could potentially miss some tests that did not run (e.g., if a run fails, no information will be given about 'compare' or 'bfbcomp').  I can continue to think about this, but would be open to suggestions as well.

PR #26, opened 2017-09-18
Update scripts 

apcraig:
Updated the scripts to fix a few problems and make consistent with first version of icepack scripts
Developer(s): tcraig
Code review: 
Other Relevant Details:

Added new pond namelist script option
Removed old cice5 scripts
Refactored some env variable names for better consistency and reuse with icepack scripts
Fixed $status ($?) checks
Added ICE_MACHINE_MAKE variable
Updated sed usage to make them POSIX compliant for use on OSX
Added testmachine machine option that should be usable on any platform to test scripts


mattdturner:
The base suite runs to completion on Conrad.  So I am marking this as approved.

apcraig:
@eclare108213, are you ok if I merge these to master?  These changes align the cice and icepack scripts a bit and fix some of the bugs that were found in cice and icepack thru testing.

eclare108213:
Yes, I think this is okay to merge.  I haven't tested it.  My one suggestion is to make sure that the documentation is up to date with what's available in the scripts.  Thx!

apcraig:
I made some minor updates to the documentation in google docs and in the README.  I think it would be good to clarify how the documentation is going to be organized.  Do we want both the google doc and README, should they be the same or how should they reference each other, how does the user guide align with the other documentation being put in place.

I will merge this now.

eclare108213:
Documentation that is downloaded with the code for offline reference should be considered "primary", in my opinion, meaning that other other references such as the wiki point to it.  The exception will be the sphinx scientific docs, which will be "primary" online with downloadable pdfs with the code.  I view the google docs more for planning and background info.  cc to @duvivier for further comment.

duvivier:
My opinion is that we should move away from using both google docs and sphinx since I think it will be easy to lose track of which have been updated and could cause confusion for users/developers about where to update. It seems to me it would be pretty easy to incorporate all the info in the google doc into sphinx and just include it as an additional page in the general documentation. I'm happy to go ahead and do this if you guys think it makes sense.

If we want to keep them separate I'd suggest we name the google docs version "Quick Start User's guide" vs. the sphinx docs "Scientific user's guide" or something like that.

eclare108213:
I agree that we shouldn't use both places as 'final' or 'primary' user references.  I do think we should continue using google docs in general, for planning, design docs, etc but not point users to them (and label them as working or draft documents).  That puts all of the "official" reference material/documentation on github.  The next question is where those docs go, in our git space.  In the June meeting at NCAR we agreed to use sphinx for the scientific documentation and user guides, and point to them from the wiki or readme pages.  We could consider putting user guides directly in as readme or other (e.g. pdf) files.  My main requirement is that the critical info is downloaded when the code is cloned.  Beyond that, it's just a matter of being able to maintain it easily as things evolve.  Sphinx is a bit more of a hurdle for that, I think, but we can document how to do that too...  Other opinions welcome.

e

duvivier:
Right now on the wiki there are these links for guides:
"Quick Start" instructions are available in README_v6. 
Online and searchable documentation of the current version of CICE   (links to sphinx)
CICE User Guide  (links to google docs)
CICE Git and Workflow Guide   (links to google docs)

Is there a reason the "Quick Start" and User Guide are separate? It looks to me like they cover the same sorts of materials. My feeling is they should be merged, probably into the README_v6 document so that we can track those changes easily and it would be downloaded with the code. Alternately, we could just make a separate wiki page with that info, but that wouldn't be downloaded with the code so I am not sure that's what we want.

The sphinx documentation link is for the "Scientific guide" and we can link the pdf that I will create as well that will be downloaded with the code. 

I'm not sure how we should address the git workflow guide. I think we need all that info available, but I'm not sure we need it to be downloaded with the code. The google doc is in nice shape, but would we want to put this info on the wiki, another sphinx page? I don't have a clear solution/path but we might want a pdf version of this as well as a working copy.

eclare108213:
Yes, I agree that the user guide info that is on google docs should be moved into github. Maybe we should get rid of the README_v6 and have something called "User Guide" (or similar) instead.  The README_v6 file is left over from the svn repo, where we didn't have the README.md front page.   README.md should be enough to convey the basic info and point to the location of 'quick start' instructions, if not show them explicitly.


In my opinion, we could keep the git workflow guide on google docs for the time being, because we're still working on it.   We should probably have a guinea pig go through everything in there and make sure it's right; the last bit about changing Icepack in CICE is especially tricky and I'm not sure we settled on the right way to do it.  Also, I think we've more or less decided against using tags but that decision hasn't been finalized.  Eventually the git workflow guide could go on a wiki page, not downloaded, in my opinion.

e


In

eclare108213:
We also have README.test, which I think ought to be part of the user guide too, rather than a separate file.


apcraig:
I agree with this strategy.  I personally think we should have a user 
guide and a science guide that's in github, in sphinx.  It could also be 
a single document, maybe that's actually better.  the repo should 
include a pdf of it/them, and the guides should be available via a clear 
web link on README.md and/or github wiki.  I think the git workflow 
guide could remain a google doc.  I think that is important to publish 
but it's important only to a subset of the community.  I think it could 
be migrated to the wiki at some point, but almost think it's easier to 
work with in google docs, at least for now.  A link should be available 
via the README.md and/or the wiki.  We should get rid of other 
references when we are ready, that is at least the README_v6 and the 
google doc user guide.  There is also a readme for testing that needs to 
be migrated into the user guide.  QuickStart could be a first chapter in 
the user guide.

Anyway, I agree that we should start the consolidation process and that 
the science/user guide info needs to be available via pdf and web and 
sphinx seems to be the way to do that.  The git workflow can remain 
separate via wiki or google doc.

eclare108213:
That sounds like the final word to me!  Let's go with it.

e

duvivier:
Sounds good to me too. I will start consolidating the readme_v6, readme.test, and google doc into new sections in the sphinx document. I'll make sure that the Readme.md and wiki files point to that document directly as well. 

I am also happy to be the guinea pig for the github workflow stuff. I'm currently working on the cice testing and trying to get a gx1 grid going with some help from Dave, so that will be a good way to test this all out.

PR #30, opened 2017-10-05
Additional files for gx1 testing.

dabail10:
Added some files for gx1 testing.
Developer(s): David Bailey
Results (bit for bit, roundoff, climate changing): bfb
Code review: Tony Craig, Matt Turner
Other Relevant Details: I tested these changes on the NCAR machine, cheyenne. Does require a couple new things:

1. A standard input directory for the gx1 forcing files. This is specified in the env.${MACHINE} file.
2. I added the restart/initial file for the gx1 grid. I have gzipped it, so it fits on github.
3. Added an ice_in namelist for the gx1 test. This uses paths from cheyenne and assumes the forcing starts from 2005.


eclare108213:
I believe we decided to not put any (big) data files directly into the repository, but rather keep them on NCAR's ftp site with instructions for how to download them and where to put them (or where to put the appropriate path to wherever they are).  Since the forcing files are already there, maybe the restart (initial) file could be housed in the same place, or similarly.

Also, rather than having a gx1-specific ice_in file, I think we will have a set of options in a namelist settings script that represents all the differences between the gx1 and gx3 namelists.  The scripts are pretty slick and I haven't completely absorbed them, but I suspect that everything here can be done in the scripts!

dabail10:
That is fine. The compressed version of the restart is 20MB, so I figured that might be ok. I agree that I did not want a gx1 specific ice_in file, but I wanted to get one there so that we could begin testing. I agree that the script should set these.

dabail10:
I will go ahead and close it.

PR #31, opened 2017-10-13
Add support for gx1 in addition to gx3.

apcraig:
Add support for gx1 in addition to gx3.
Refactor way input datasets are set.
A few scripts improvements.

Developer(s): tcraig
Code review: 
Other Relevant Details:

This will impact the directory structure and location of the input datasets moving forward.  Two things to note, in env.machine, the value of ICE_MACHINE_INPUTDATA now points to a top level directory of CICE inputdata.  In set_nml.gx3 and set_nml.gx1, the actual paths and filenames are specified.  Then when the namelist gets "written" to ice_in, the ICE_MACHINE_INPUTDATA and ICE_SANDBOX strings are replaced by their local values.


apcraig:
Please have a look at these changes @eclare108213, @dabail10, @duvivier .  This provides support for multiple resolutions and a formal inputdata area moving forward.  We should finalize the inputdata directory structure and input data filenames, revise the implementation, test, and update this PR before merging the PR.  

This has been tested on cheyenne and it should work on yellowstone, including for gx1 IF the gx1 restart file is dropped into scripts/configuration/data/gx1.  Other machines will need modifications to the ICE_MACHINE_INPUTDATA setting and the location of the input filesnames for this PR to work.  Again, this demonstrates the capability but does not necessarily reflect the final input data filenames or paths.

apcraig:
OK, I think this is ready.  I have tested on cheyenne, yellowstone, conrad (from input data tar files), and thunder (from input data tar files) with mostly successful results.  There may be some outstanding issues that we want to resolve, but overall, I think this PR accomplishes what was needed to support different grids with the new input dataset directory structure.  If others don't disagree, I will execute the PR in the next day or two.  If someone wants to do the PR sooner, go ahead.

eclare108213:
I'm testing this now.  Since the grid and restart files are in the data tar files and will be kept within that directory structure, can we now remove the configuration/data directory from the main repo?  Most things in there have other homes now, but I'm not sure about ice.restart_file.

apcraig:
Yes, we should remove the data file from the repo.  I was just waiting for agreement before doing that, but I can do that on my branch later today.

apcraig:
The entire data directory has been removed from the repo now.

PR #33, opened 2017-10-17
Autogen CTestTestfile.cmake

mattdturner:
Update CTest scripts to have the Testfile be automatically built based on the test suite being run.
Developer(s): Matt Turner
Results (bit for bit, roundoff, climate changing): N/A
Code review: 
Other Relevant Details:
- This should resolve the concern about the previous implementation of CTest/CDash implementation having the individual tests card-coded.  The scripts now determine which tests to post to CDash based on the tests being run for the given test suite.


mattdturner:
@apcraig This pull request should address the concern that you had about the list of tests being hard-coded.

mattdturner:
It detects failed tests.  All gen_ctestfile.csh does is generate the CTestTestfile.cmake file.  This file checks the results.log file for PASS for each test.  If it is unable to find PASS for the given tests, then it reports a failure.

I can change the script to accept platform as an argument.

PR #42, opened 2017-12-01
Update CICE documentation with Compliance Testing Code from Andrew.

dabail10:
Developer(s): D. Bailey, A. Roberts
Updated documentation (Y/N): Y
Results (bit for bit, roundoff, climate changing): bfb
Code review: E. Hunke, A. Roberts
Other Relevant Details:


dabail10:
Added Matt's documentation as well. Still have some references that I need to talk to Matt  about.

Dave

_sources/cice_3_user_guide.rst.txt
```diff
@@ -2262,7 +2262,80 @@ hemispheres, and must exceed a critical value nominally set to
 test and the Two-Stage test described in the previous section are
 provided in :cite:`Hunke2018`.
 
-
+In applying equations :eq:`t-distribution` through :eq:`short-means`,
```
proteanplanet: I suggest that lines 2265 to 2300 need to be removed from this, and the corresponding html files also needs to be changed.

proteanplanet: Please note also, in reviewing this pull request, I have found some isses with the documentation of the documentation that are not actually in this request, but are important for its implementation. 

First about-sphinx-documentation.txt  contains a link to https://github.com/NCAR/CICE that is not accessible to the outside world, at least not to me.  Please revise this link.  In addition, this file includes a bunch of funny characters when using a text editor (vim or emacs) to read it that look like the text has been cut and pasted. Characters such as apostrophes are unreadable.   This is also true of the file sphinx-documentation-workflow.txt. 

Second, the Makefile as currently exists under /doc does not work, instead giving the error message: "Makefile:12: *** commands commence before first target.  Stop.". This is because of the following lines: 
# User-friendly check for sphinx-build
ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
        $(error The '$(SPHINXBUILD)' command was not found. Make sure you have Sphinx installed, then set the SPHINXBUILD environment variable to point to the full path of the '$(SPHINXBUILD)' executable. Alternatively you can add the directory with the executable to your PATH. If you don\'t have Sphinx installed, grab it from http://sphinx-doc.org/)
endif

dabail10: I thought that text was from Matt in section 3? I'll work on the other stuff, but the Makefile works if you have python/2.7.11 installed as well as sphinx. For the citations, you also need sphinxcontrib-bibtex installed. I will fix that link. That is our user's guide for the CESM-CICE. Thanks for the feedback Andrew.

proteanplanet: Just the lines I've highlighted are from my original write-up and have changed.  Matt has added everything after the "Practical Testing Procedure" header. My version of python was 2.7.10.

dabail10: Got it.

eclare108213:
There will be more changes to bring in;  this gets the documentation closer.

PR #51, opened 2017-12-15
update PR template

eclare108213:
Make the PR template less cryptic.  Is this too wordy?


eclare108213:
@duvivier Please take a look at this.  If it makes sense to you, I think you can merge this PR.  Or I can merge it, or you can directly edit the file yourself and I can close this PR.  Either way, we should make the same changes to the template for Icepack.  Thanks!

duvivier:
I think that the changes make sense. I agree it's better to be explicit, especially as we work out the sphinx documentation and users get used to the new workflow. I'll update the Icepack PR template.

Issue #49, opened 2017-12-15
capture timing results


eclare108213:
Record timing results as part of the testing procedure, or at least flag them if changes are substantial in comparison tests.  Preferably include at least these:
Timer   2:  TimeLoop
Timer   3:  Dynamics
Timer   5:    Column

mattdturner:
In order to post the timings to CDash, we can only have 1 timer logged.  We can include the timer information for multiple timers elsewhere (although I'm not sure where).

For example, http://my.cdash.org/viewTest.php?onlypassed&buildid=1338795 is what I have been working on today.  I've been putting the 'Timer 1: Total' timer into the Time field for the *_run tests for smoke tests only.  I could easily change to another timer, but we still have the issue of only a single timing being able to be posted to CDash.

eclare108213:
If we can only choose one, then the best one is TimeLoop.  If that one is very different between runs, then we can look more closely at Dynamics, Column, etc. manually.

mattdturner:
I think this was addressed in #125 .  The wiki is now showing timings for tests (https://github.com/CICE-Consortium/Test-Results/wiki/3eed537e29.conrad.intel.180418.000150).

eclare108213:
Cool, thank you.

PR #54, opened 2017-12-18
Ctest timings

mattdturner:
Add timing information to the CDash online dashboard
Developer(s): Matt Turner
Results (bit for bit, roundoff, climate changing): N/A
Code review: 
Other Relevant Details:

- Timing information is posted for every *_run* test

### DIFF
--- a/configuration/scripts/tests/CTest/parse_timings.csh
+++ b/configuration/scripts/tests/CTest/parse_timings.csh
@@ -17,17 +17,32 @@ foreach line ("`cat $testfile`")
     endif
   endif
   if ( "$line" =~ *"FullName"* ) then
-    if ( "$line" =~ *"_run<"* && $save_time == 1) then
-      set save_time=1
+    if ( $save_time == 1 ) then
+      if ( "$line" =~ *"_run<"* ) then
+        set save_time=1
+      else if ("$line" =~ *"_run-initial<"*) then
+        set save_time=2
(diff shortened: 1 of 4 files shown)


mattdturner:
This PR addresses Issue 49 (https://github.com/CICE-Consortium/CICE/issues/49)

eclare108213:
http://my.cdash.org/viewTest.php?buildid=1340131
I managed to run this branch and upload the results!  I had changed the proxy settings for downloading python packages, and that also fixed this problem.  The results are not stellar because I told run.suite to compare with a baseline that does not exist (oops).   I suggest that if a user makes this particular mistake, that the results not be uploaded to CDash.  Also, CDash is not showing the timings because my test runs create.case and the other scripts from the repo, which do not yet have the commands to be merged in this PR.

eclare108213:
Matt, I wasn't very clear in my last comment.  Is it possible to have the scripts not upload results to CDash when the error is that a baseline doesn't exist?  That's a user problem, not a code problem.

mattdturner:
@eclare108213 Do you want it to not upload anything to CDash if the baseline for a test doesn't exist?  Or not upload data for the specific tests that do not have a baseline?  (The former would be easier to implement)

eclare108213:
The former!  Just write a message saying nothing will be uploaded because the baseline doesn’t exist, but the results can be looked at locally [where].  Thanks!

PR #57, opened 2017-12-21
T-Test to CDash

mattdturner:
This PR updates the QC script to post the results to CDash.

Developer(s): Matt Turner
Are the code changes bit for bit, different at roundoff level, or more substantial? N/A
If not, does the documentation need to be updated separately? (Y/N) N (I don't think it needs to be updated, since all of the additions are automated).
Other Relevant Details:


apcraig:
Matt, sorry to do this, but this PR is showing conflicts now.  Can you have a look and see if you can resolve the problem.  Looking at the conflict, it's not totally clear why this is happening.  It looks like all the conflicted code needs to be included, but please take a look, then we can merge.  thanks.

mattdturner:
I'm not sure why the 'merge' had to be done twice, but I resolved the conflict via the online conflict resolution tool

PR #68, opened 2018-01-10
Move set_nml.ttest to set_nml.qc

mattdturner:
Rename the QC namelist, and set it to only output ice thickness daily (remove the other variables from daily history files)
Developer(s):  Matt Turner
If not, does the documentation need to be updated separately? (Y/N) Y
Other Relevant Details:


mattdturner:
I actually need to add another commit to update the documentation...

mattdturner:
Documentation has been updated.  I also included a line in the documentation about the Python modules necessary for plotting QC failures (assumes https://github.com/CICE-Consortium/CICE/pull/55 will be merged).

PR #69, opened 2018-01-10
Fix earlier yellowstone removal.

dabail10:
Fixed a problem with the earlier yellowstone removal.
Developer(s): D. Bailey
Other Relevant Details:


eclare108213:
Fixed in PR #70

PR #70, opened 2018-01-10
End to end Documentation (+ bugfix)

mattdturner:
Add documentation for an end-to-end testing procedure.  Also include a bugfix for cice.batch.csh and cice.launch.csh that showed up when Yellowstone was removed from machines.
Developer(s): Matt Turner
Is the documentation being updated with this PR? (Y/N) Y
If not, does the documentation need to be updated separately? (Y/N)
Other Relevant Details:

### DIFF
--- a/configuration/scripts/cice.batch.csh
+++ b/configuration/scripts/cice.batch.csh
@@ -36,7 +36,7 @@ EOF0
 
 #==========================================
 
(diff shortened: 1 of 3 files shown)


eclare108213:
This also fixes the issue in PR #69 so I will close that one.

PR #76, opened 2018-02-07
Update CICE scripts with new arguments

apcraig:
Update the CICE scripts with new arguments, add report_results capability, new machine/compiler files, rename create.case to cice.setup,
Developer(s): tcraig
Are the code changes bit for bit, different at roundoff level, or more substantial? bit-for-bit
If not, does the documentation need to be updated separately? (Y/N) Y
Other Relevant Details:

See https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks, hash 17f346b4150 for initial results.


apcraig:
Good question.  I had to change how ICE_RUNLENGTH was working in CICE.  In Icepack, it's a time, but all runs are short and fast, so they can always run in the fast/debug/short/serial queue.

With CICE, it's more complicated.  Some runs might be small and short and can run in a debug queue.  Others might need to run the in regular queue.  So I am using ICE_RUNLENGTH as an integer to trigger what batch time and batch queue to use.  That is done in cice.batch.setup.  We need some documentation for this feature.  Basically, 0 means set time to 15 minutes and then to appropriate queue.  As that integer increases, cice.batch.setup can set longer times and different queues.  

I agree that we need to review the nml and env settings, decide on defaults, and think about test coverage.

eclare108213:
I'll add an issue for documentation of the scripts

PR #99, opened 2018-03-13
Modif3mar2018

JFLemieux73:
Developer: JF Lemieux, ECCC

1) Modified basal_stress_coeff to remove an if statement (following a suggestion by Till Rasmussen). I got BFB results.

2) There was a minus sign missing in the calculation of taubx and tauby for the outputs. This is just for the outputs. It does not affect the prognostic variables (e.g. uvel, vvel, vice, aice...).

3) Added internal pressure as an output (calculated in principal_stress). Possible names were sigI, internal_pressure, Pint, Pinternal and sigP. I decided to call it sigP. We could find a better name later if someone has an idea...

4) Normalization of sig1 and sig2 is now done with the ice strength for both evp and eap (instead of prs_sig for the evp). I removed prs_sig as it is not needed anymore. This does not change the prognostic variables (e.g. uvel, vvel, vice, aice...). This is just for diagnostic. I compared sig1 and sig2 obtained this way with the previous ones and they are the same where both are calculated (they are not calculated at all the same locations because of the new condition 'if (strength(i,j) > puny) then' in principal_stress compared to the old one 'if (prs_sig(i,j) > puny) then' .

Is the documentation being updated with this PR? N

### DIFF
--- a/cicecore/cicedynB/analysis/ice_history.F90
+++ b/cicecore/cicedynB/analysis/ice_history.F90
@@ -5,7 +5,7 @@
 !
 ! The following variables are currently hard-wired as snapshots 
 !   (instantaneous rather than time-averages):
-!   divu, shear, sig1, sig2, trsig, mlt_onset, frz_onset, hisnap, aisnap
+!   divu, shear, sig1, sig2, sigI, trsig, mlt_onset, frz_onset, hisnap, aisnap
 !
 ! Options for histfreq: '1','h','d','m','y','x', where x means that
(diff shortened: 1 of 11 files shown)


eclare108213:
I tested this in the gx1 configuration, and it was BFB identical to #41158c1 for 3.5 years, when I stopped the run.  Although the principal stress normalization is being done differently, the ellipse plots also appear identical (from iceh.2003-05.nc).

Documentation for this and other changes to the dynamics still needs to be updated.  I created an issue for that task.

ellipse_jf

PR #103, opened 2018-03-22
V6 alpha prep

duvivier:
Getting documentation ready for v6alpha release.

Developer(s): Alice 
Are the code changes bit for bit, different at roundoff level, or more substantial? bit for bit
Is the documentation being updated with this PR? Yes, version number only
Other Relevant Details:

Removing some files about using sphinx that we don't need, removing the cicedoc.pdf (might want to do more removal of files still...), simplifying Readme.md, and updating version number

We need to make some wiki changes along with this, but since I don't know how to do those as part of a PR, they aren't included.


eclare108213:
Also need to remove CICE/doc/CICE-Consortium.pdf
I'm not sure why that's there...

duvivier:
I removed the PDF that ECH requested. Still need to work on the Wiki links.

PR #105, opened 2018-03-23
Dummy argument and unused variable cleanup.

dabail10:
Ran the main CICE code through the NAG compiler and cleaned up dummy arguments and unused variables. Also added support for Hobart (NCAR) NAG and Intel. Still doing some work on uninitialized variables.
Developer(s): D. Bailey
Other Relevant Details:


dabail10:
I ran the base_suite on cheyenne with intel and compared to 6.0.0. Everything is bfb with the prior version. I fixed the ifdef logic around the flush statements. I'm confused about what you are saying on ice_init_column.F90 though. Is there ifdef logic when the BGC is turned on? I commented out stuff that I thought might be used eventually. Similarly with ice_dyn_eap.F90. In terms of nu_diag, it is easy to add it when debugging.

eclare108213:
[eclare@pi-fe1 base_suite.dabail]$ results.csh | grep FAIL
FAIL pinto_intel_smoke_col_1x1_bgcNICE_debug compare v1.0.2 different-data
FAIL pinto_intel_restart_col_1x1_bgcNICE exact-restart
FAIL pinto_intel_restart_col_1x1_bgcNICE compare v1.0.2 different-data
3 of 82 tests FAILED

Fortunately one of the different-data fails is with the debugging flags on, so maybe I can use the debugger...  @apcraig or @mattdturner could one of you tell me quickly what data is being compared?  i.e. history or restart output after n dt/days, or diagnostic output, or what?  I should know that......   Thx

eclare108213:
The diagnostic output files in my tests are identical.  This is Icepack so those ARE the history output files.  And there are no restart files written at all.  So one problem is with how the test is set up.   In these tests, dumpfreq = 'y', and the runs are only 1032 hours long.  In other restart tests in the base_suite, dumpfreq = 'm'.   Try changing that for bgcNICE and bgcISPOL...  might still fail though.  This might require changing when restarts are checked in the test suites.  

Anyhow, this seems to be a problem with the test, not with your changes to the code, @dabail10.  There is a file conflict that needs to be fixed.

dabail10:
So, I ran the base_suite as it is on cheyenne with intel. This passed all but four tests:

FAIL cheyenne_intel_decomp_gx3_4x2x25x29x4_spacecurve run
FAIL cheyenne_intel_decomp_gx3_4x2x25x29x4_spacecurve test
FAIL cheyenne_intel_decomp_gx3_4x2x25x29x4_rakeX1 run
FAIL cheyenne_intel_decomp_gx3_4x2x25x29x4_rakeX1 test

Note that the rakeX1 test also fails in CICE6.0. Travis is failing on the following:

PASS travisCI_gnu_restart_gx3_2x1 build
PASS travisCI_gnu_restart_gx3_2x1 initialrun
PASS travisCI_gnu_restart_gx3_2x1 run 37.51 5.00 21.61
FAIL travisCI_gnu_restart_gx3_2x1 test 
#---
PASS travisCI_gnu_restart_gx3_1x2 build
PASS travisCI_gnu_restart_gx3_1x2 initialrun
PASS travisCI_gnu_restart_gx3_1x2 run 38.24 4.18 20.90
FAIL travisCI_gnu_restart_gx3_1x2 test 

Is this known?

Dave

dabail10:
That is correct, although Cheyenne was acting weird today, so I would like to rerun the base_suite.

apcraig:
Travis should be passing.  Separately, I will try to run the full CICE suite on cheyenne and post results.  Unless you want to do that Dave.  We should run and post results for the current master.  That's also what we should be comparing to.

dabail10:
I am rerunning the base_suite again on cheyenne.

Dave

dabail10:
So, the rakeX1 is failing, but the decomp is not valid. The message is:

ice: no. blocks exceed max: increase max to 5

This is the same for the CICE6.0 basecode and my code. The spacecurveX1 is failing for a strange message in only my modified code:

Error in `./cice`: free(): invalid next size (fast): 0x00000000052dc930

It is dying in create_distrb_spacecurve. I need to do more digging here.

apcraig:
That makes sense, we can easily increase the maxblocks for the rakeX1 test.  To be honest, I still don't understand what is being raked and how that works, even though I've looked at the code a few times.  Anyway, we can fix that test, I think we just need to change

decomp gx3 4x2x25x29x4

to 

decomp gx3 4x2x25x29x5

in the base_suite.ts file.  I will make that change in my PR and/or you can in yours as well.

eclare108213:
Presumably the blocks are being raked from one processor to another.  The gx3 grid might be too small for it to really manifest itself, but if it's asking for maxblocks to increase to 5, then it sounds like at least one block is moving to another processor.   You have to dig deep in the code to diagnose the block layout -- I think there's a print statement somewhere that's commented out.

apcraig:
So the rake is trying to load balance using an estimate of the cost of each block?  That makes sense and it also makes sense that the outcome would be some tasks with more blocks than others and the need to increase maxblocks compared to other decompositions.  Thanks for providing some insight.

dabail10:
I guess I'm confused about how these decomp tests work. The code is compiled with the same block size in all cases and the number of processors is always 4. Then the algorithm for laying out the blocks is changed in each case. The number of blocks is always four, so why is it asking for 5 for max blocks? I think as Elizabeth highlights, these are not the best for the gx3 grid.

eclare108213:
The philosophy behind the grid decomposition is that we have a collection of blocks that cover the global grid (minus some that are all-land cells), and we have a certain number of processors, and we need to figure out which blocks to collect on eah processor for optimal load balancing.  So we generally start out thinking about a cartesian decomposition, and moving blocks around from that nice orderly arrangement onto nearby processors that we think of as "assigned" to regions of the grid, but it doesn't have to be like that at all.  The space-filling curves, for instance, assign segmented lines ('curves') of blocks to processors, and it doesn't matter at all what the processor layout is to begin with.  My advice is to not think of the processors as being associated with the grid, but as being 'bins' into which we throw work.  We do try to keep the work contiguous on the grid, to reduce communication, but that's not required.  The decomp tests on gx3 have 4 processors, and work from anywhere on the grid can be put on them; the decomposition choices define how that happens. The rake algorithm is deciding that it's better to have 5 blocks on one of the processors, and fewer on another processor.  @apcraig added some of the newer decompositions, like roundrobin, and we have figures illustrating them in the documentation:
http://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_implementation.html#performance

apcraig:
Just another comment about the decomp test.  That test is the first test that leverages a single binary to execute multiple tests.  We hope to leverage that more and more, so each run doesn't require a different case and it's own build, where possible.  But the only way to do that is to have the same grid, blocksize, and maxblocks.  The decomp test simply takes the blocks defined and runs a different decomposition of those blocks with the same binary, and it compares that results are identical.  This is all quite fast because it's reusing the same binary and runs all the tests as one job, and it works nicely in that respect.

There is another test suite, decomp_suite.ts.  If you look at that, it is running different blocksizes, decompositions, and so forth for each test and comparing against each other.  That is testing padded blocks and some other stuff.  That is also nice, but each case needs to be built independently, so it actually takes a bunch more wall time.  In many cases, the build takes longer than the run and the builds are very serial in our test suites whereas runs could be done concurrently (if we could do the builds fast enough).  I think we're still working on setting up the appropriate test suites.  Maybe we should move away from decomp_suite.ts and just use the decomp test but do it with padded blocks and other things.  But the other problem is that some decompositions only work with certain blocksizes (cartesian for instance) so we can't really test any blocksize with all decompositions, so that limits what we can do with the decomp test.  Again, I think we're still trying to figure out what will work best.

apcraig:
Just for reference, I have run the current CICE master for the base_suite, travis_suite, decomp_suite,and quick_suite and have posted results here, https://github.com/CICE-Consortium/Test-Results/wiki/3007d8288e.cheyenne.intel.180502.232321.

eclare108213:
I'm not sure where we are with this PR.  It's been sitting here long enough that there are conflicts needing to be resolved, but we've built out our test suites quite a bit, so I would feel more confident about these code changes if those all pass.  @dabail10 could you try to resolve the conflicts and run the base_suite and decomp_suite tests again?

dabail10:
I was kind of waiting on figuring out a few of these restart issues. I hope to resolve this next week.

Dave

eclare108213:
Which restart issues?

anders-dc:
@eclare108213 From the [raw travis log](https://api.travis-ci.org/v3/job/375061106/log.txt):

```
PASS travisCI_gnu_smoke_gx3_1x1_debug_run1day build
PASS travisCI_gnu_smoke_gx3_1x1_debug_run1day run 58.56 18.02 23.28
PASS travisCI_gnu_smoke_gx3_1x1_debug_run1day test 
#---
PASS travisCI_gnu_smoke_gx3_2x2_debug_run1day build
PASS travisCI_gnu_smoke_gx3_2x2_debug_run1day run 67.81 35.98 17.64
PASS travisCI_gnu_smoke_gx3_2x2_debug_run1day test 
#---
PASS travisCI_gnu_smoke_gx3_1x2_run2day build
PASS travisCI_gnu_smoke_gx3_1x2_run2day run 19.38 2.11 10.90
PASS travisCI_gnu_smoke_gx3_1x2_run2day test 
#---
PASS travisCI_gnu_smoke_gx3_2x1_run2day_thread build
PASS travisCI_gnu_smoke_gx3_2x1_run2day_thread run 16.40 1.80 9.97
PASS travisCI_gnu_smoke_gx3_2x1_run2day_thread test 
PASS travisCI_gnu_smoke_gx3_2x1_run2day_thread bfbcomp travisCI_gnu_smoke_gx3_1x2_run2day.travisCItest
#---
PASS travisCI_gnu_restart_gx3_2x1 build
PASS travisCI_gnu_restart_gx3_2x1 initialrun
PASS travisCI_gnu_restart_gx3_2x1 run 42.12 5.04 25.36
FAIL travisCI_gnu_restart_gx3_2x1 test 
#---
PASS travisCI_gnu_restart_gx3_1x2 build
PASS travisCI_gnu_restart_gx3_1x2 initialrun
PASS travisCI_gnu_restart_gx3_1x2 run 45.69 5.06 25.54
FAIL travisCI_gnu_restart_gx3_1x2 test
```

eclare108213:
Ah, got it, thanks.

dabail10:
I am closing this and I will reissue later.

Issue #77, opened 2018-02-07
Document CICE scripts


eclare108213:
especially ICE_RUNLENGTH

eclare108213:
The CICE documentation still has create.case instead of cice.setup, with the old flags.  That should be fixed soon!

duvivier:
Oops, I just changed that all. Will do a PR soon.

duvivier:
Just adding to this (as a list for my reference): 
- The quickstart, user's guide, and developer's guides all need to be thoroughly tested and updated with the correct documentation in the *rst files.

eclare108213:
If ICE_RUNLENGTH has been documented, this issue can be closed.  I don't remember seeing that.  ??

apcraig:
ICE_RUNLENGTH is documented and I will close this, but we will need to review documentation before every release.

PR #106, opened 2018-03-23
Updating CICE doc structure

duvivier:
Updating CICE documentation to have same structure as Icepack.

Developer(s): Alice
Are the code changes bit for bit, different at roundoff level, or more substantial? bit for bit
Is the documentation being updated with this PR? (Y/N) Y
Other Relevant Details:

I wanted to organize this roughly the same way as Icepack, so I've broken up the documentation into sections the way we did for icepack. There still needs to be significant changes to the documentation (e.g. figure labeling, section referencing), but now we can all work on the same documents the way they should be done.

I also updated all the script to cice.setup instead of create.case as per Issue #77


duvivier:
@apcraig Also, it doesn't look like we have Travis-CI working yet with CICE, is that correct?

apcraig:
Correct, travis is not yet setup with CICE.  There are many things we need to review and do with CICE at this point to "get it up to date".

duvivier:
Okay, thanks @apcraig . I just wanted to be sure since I did the PR and there was no automatic testing that could test if things are broken (though I only touched the documentation *rst files, so this shouldn't be an issue).

apcraig:
The documentation definitely needs a thorough review with respect to using the scripts.  I think the CICE scripts were updated but the documentation hasn't been yet.  I can take that on if you like.  We could migrate the Icepack documentation to CICE then revise as needed.  Or we can just rewrite what we've got in CICE.

eclare108213:
Here are some additional items I noticed that need to be changed, but not necessarily before this PR is merged:

> user_suide/ug_performance.rst:  the test setup information should be put somewhere else.  Is this duplicated in user_guide/ug_testing.rst?  If so, just delete it here.

> user_guid/ug_running.rst:  this is very old info, from cice v5!  Maybe this is where to put setup info that's currently in ug_performance?  We need to keep the table of environment variables.

> user_guide/ug_testing.rst:   update the cice.setup flags and remove the version/date line at the top

> user_guide/ug_troubleshooting.rst:  this is v5 info — update to v6

> Many of the new files contain text specific to Icepack, which needs to be changed for CICE.  E.g.
> developer_guide/dg_about.rst, dg_driver.rst, dg_scripts.rst

> remove all references to drupal

apcraig:
Also, I think this could be a few step process and doesn't all need to happen in one PR.  There are several tasks that need to be done before CICE is "more ready" including travis, documentation, further testing, and so forth.  We need to look thru the outstanding issues as well.

duvivier:
@apcraig @eclare108213 I have made the changes that Elizabeth lists as necessary but *not* the additional items yet. I just wanted to get this structure into the repo so that we are all working with the same format. Once we have that in there I think we will need to go through it with a fine tooth comb and re-do it. Basically what was there before is just what we decided was part of "CICE" as opposed to "Icepack" when Elizabeth and I did the initial separation. There will be lots of tweaking to do to this as we get ready for the real release so I think there will be a number of PRs to finish it all. I've made note of Elizabeth's comments above to include soon in a PR update.

eclare108213:
Thanks @duvivier !

PR #108, opened 2018-03-27
QC Script Debug

mattdturner:
Update the QC script to properly handle the case where the effective sample size is greater than 30.

Developer(s): Matt Turner
Other Relevant Details:
- I also commented out the call to the post_to_cdash function since we are no longer using cdash.


eclare108213:
Thanks, Matt!

Issue #82, opened 2018-02-18
version number in netcdf output


eclare108213:
Have the code read the current version number from version.txt (or whatever is most appropriate) and use it in the netcdf metadata, e.g.
>    :source = "Los Alamos Sea Ice Model (CICE) Version 5" ;

Check other output formats to see if it can be used there also.

apcraig:
Resolved in #121

PR #121, opened 2018-04-06
Version number in netcdf

mattdturner:
These edits now include the CICE version number (pulled from version.txt) in the netcdf output files.

- Developer(s): Matt Turner

- Other Relevant Details:
This addresses https://github.com/CICE-Consortium/CICE/issues/82
The attribute in the netcdf file is currently:
`:source = "Los Alamos Sea Ice Model, CICE_6.0.0.alpha" ;`


eclare108213:
Question:
> if a user sets --version in cice.setup, then the version.txt is overwritten.

Usually, a --version flag only returns what the version number is, doesn't it?  Why would version.txt be overwritten?

apcraig:
This is a good point.  The --version feature was added at some point to allow the scripts to update the version via the scripts, rather than have the user have to do it manually.  That feature probably needs to be reviewed to make sure it's working as it should, and we can decide whether we want it or not.  It's also true that --version may not be the best name.  Maybe --version should return the version number and --setversion should set it.

eclare108213:
Let's have --version return the current version number.  I didn't know we had an option to use the scripts to set the version number.  Maybe that could be helpful, since we have version numbers in numerous places.

apcraig:
It might also be good to implement the CICE version not as a cpp but as a namelist input.  Generally, cpps should be avoided except for this that have to be defined at compile time.  The version name is not one of those things.

apcraig:
@mattdturner , I would not worry about the --version argument at this point.  I'll fix that when we make some progress on version naming process.

mattdturner:
Ok, I updated the Pull Request to remove the CPP flag for version and replace it with a `version_name` variable in the setup_nml namelist.  `version_name` is defined in ice_history_shared.F90, and assigned a default value.

eclare108213:
My preference is to leave it as it is.  Considerations:  (1) "Los Alamos Sea Ice Model, CICE" is the official, legal name of the model.  (2) The Consortium might eventually have more than one sea ice model.  (3) Some people have turned CICE into the acronym for Community Ice CodE, but I never adopted that, partly for funding-agency (DOE) reasons and partly because I'm just not that fond of it.  

I'm open to other suggestions, but perhaps this should be discussed in a separate issue rather than this PR.

PR #126, opened 2018-04-20
Update documentation

apcraig:
Update the documentation including reorganization consistent with the Icepack release.

- Developer(s): tcraig

- Other Relevant Details:

This is a step forward, documentation needs additional review and needs to continue to evolve.
Also removed the CTest directory and scripts, these are no longer needed.


apcraig:
Built documentation can be viewed here,

http://apcraig-cice.readthedocs.io/en/doca/

PR #141, opened 2018-05-15
Doc for grounding scheme and tensile strength

JFLemieux73:
[Remove this and add a short summary line]:

- Developer(s): JF Lemieux

- Are the code changes bit for bit, different at roundoff level, or more substantial? N.A. only changes to te doc.

- Other Relevant Details: I think the doc was a bit misleading for the replacement pressure (the replacement pressure is only the term on the LHS of eq. 53). I have made some modifications to clarify this but it would be good if Elizabeth could have a look to make sure this is ok.


eclare108213:
Closing this PR, replaced with #142

PR #143, opened 2018-05-17
Add "--queue" as an option to cice.setup

mattdturner:
This PR adds the ability to pass a batch-system queue to the cice.setup script.

- Developer(s): Matt Turner

- Other Relevant Details:
- Default queues are defined in each individual env.<machine>_<compiler>
- This should resolve an issue where some tests within the suites would require a larger walltime than the hard-coded queues.


mattdturner:
> Second, it looks like we no longer have a queue dependence on runlength where short jobs might be debug and longer jobs might be in another queue.

That was my original thought with having the queue passed via `--queue` to only be used for jobs that don't fit in the debug queues.  That is a simple thing to do.  However, having the queue be defined based on runlength when `--queue` is not passed would be a bit more difficult.

I forgot to set the default queue.  I'll do that now.

Issue #138, opened 2018-05-11
boxrestart test is failing exact restart


apcraig:
The boxrestore test does not restart exactly.

apcraig:
this is fixed in #163 with restart_ext mod.

Issue #137, opened 2018-05-11
frz_onset not bit-for-bit with different decompositions


apcraig:
In recent testing, the restart field frz_onset is not bit-for-bit with different decompositions in the decomp_suite.  This error occurs at 1 gridcell only and needs to be further debugged.  Some thoughts as to what might be causing this are missing non-initialization, gx3 forcing issues, or an issue on a halo.  But those are all speculation at this point.

eclare108213:
I tried to run the decomp suite on pinto and it fails because the maximum number of tasks per node is 16, which makes taskpernode=16/20=0 in cice.batch.csh.  Not sure what to do about this kind of issue.

[eclare@pi-fe1 CICE.ech]$ cice.setup -m pinto --grid gx3 --test restart --pes 1x20x5x29x80 --set dsectrobin --testid e2
 
cice.setup:
ICE_SANDBOX  = /turquoise/usr/projects/climate/eclare/CICE.v6.0/github/CICE.ech
ICE_CASENAME = pinto_intel_restart_gx3_1x20x5x29x80_dsectrobin.e2
ICE_CASEDIR  = /turquoise/usr/projects/climate/eclare/CICE.v6.0/github/CICE.ech/pinto_intel_restart_gx3_1x20x5x29x80_dsectrobin.e2
ICE_MACHINE  = pinto
ICE_COMPILER = intel
ICE_PES      = 1x20
ICE_GRID     = gx3 (100x116) blocksize=5x29x80
adding namelist mods set_nml.gx3
adding namelist mods set_nml.dsectrobin
running parse_settings.sh
running parse_namelist.sh
running parse_namelist_from_settings.sh
running cice.run.setup.csh
cice.batch.csh cice.run
1 20 16 climateacme    (added line: echo ${ntasks} ${nthrds} ${maxtpn} ${acct})
Division by 0.
/turquoise/usr/projects/climate/eclare/CICE.v6.0/github/CICE.ech/pinto_intel_restart_gx3_1x20x5x29x80_dsectrobin.e2/casescripts/cice.run.setup.csh: ERROR cice.batch.csh aborted
cice.setup: ERROR, cice.run.setup.csh aborted

eclare108213:
This is not obvious to me.  I'm getting 4 failures (commenting out the problematic 20-thread case):
[...]

PR #146, opened 2018-06-01
fix decomp test failure by broadcasting calendar fields

eclare108213:
Fix nonBFB decomposition test comparisons.

- Developer(s):  E. Hunke

BFB except for the tests that were failing.

- Other Relevant Details:

This PR addresses issue #137.   All of the ranks/tasks need calendar information for the frz_onset and mlt_onset variables, but nyr was not broadcast when using restart calendar information.  This broadcast has been added for all three I/O types (only netcdf is tested here).  We might also need to broadcast month, mday and sec too, but I’ve elected to wait on those until we actually see a problem with them.

I also updated drivers/cice/CICE_RunMod.F90_debug to match the current module.  This allows helpful printing of many pertinent fields throughout the timestepping for a grid point specified in ice_diagnostics.F90 (I added frz_onset to the printing routine temporarily but am not committing that change here).  

I have tested base_suite and decomp_suite, but using pes=1x8x5x29x80 in decomp_suite instead of 1x20x5x29x80 (see issue #137).  The boxrestart test fails (known issue, #138) in base_suite, otherwise everything passes.

### DIFF
--- a/cicecore/cicedynB/infrastructure/io/io_binary/ice_restart.F90
+++ b/cicecore/cicedynB/infrastructure/io/io_binary/ice_restart.F90
@@ -36,7 +36,7 @@
 
       subroutine init_restart_read(ice_ic)
 
-      use ice_calendar, only: istep0, istep1, time, time_forc, npt
+      use ice_calendar, only: istep0, istep1, time, time_forc, npt, nyr
       use ice_communicate, only: my_task, master_task
       use ice_dyn_shared, only: kdyn
       use ice_read_write, only: ice_open, ice_open_ext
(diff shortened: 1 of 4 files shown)


eclare108213:
Yes, this just brings CICE_RunMod.F90_debug up to date.  If you diff that with CICE_RuMod.F90, the only differences are the lines needed for printing out diagnostic info.

eclare108213:
And I used it to track down this bug, so consider it working...

dabail10:
Is this also a problem in CICE5? I noticed in our version that the io_pio version was reading in nyr on all tasks. Here the broadcast is not necessary. Our io_netcdf and io_binary have this issue, but we never run with these in the CESM.

eclare108213:
I'm not sure whether we had the namelist option use_restart_time in CICE5, but even if we didn't, this might still have been a problem.  It's rare though, not something I would have noticed even if I were paying very close attention to frz_onset and mlt_onset, which are just kludgey diagnostics.

PR #148, opened 2018-06-07
Icepack interface for binary restarts

eclare108213:
Added Icepack interface calls to binary restart module and isolated netcdf bgc code.

- Developer(s):  E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB

- Other Relevant Details:

The modal_aero option requires netcdf input, so I added an abort if modal_aero=T and netcdf is not available.  @njeffery is it possible to include this data directly in the code or provide binary for it?

The change to io/netcdf/ice_restart.F90 is cosmetic.

This PR is not yet fully tested, just trying to keep Rob going...

### DIFF
--- a/cicecore/cicedynB/general/ice_forcing_bgc.F90
+++ b/cicecore/cicedynB/general/ice_forcing_bgc.F90
@@ -848,6 +848,7 @@
     if (icepack_warnings_aborted()) call abort_ice(error_message="subname", &
        file=__FILE__, line=__LINE__)
 
+#ifdef ncdf
     if (modal_aero) then
        diag = .true.   ! write diagnostic information 
        optics_file =  &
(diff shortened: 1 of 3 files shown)


eclare108213:
Smoke and restart tests using binary I/O work.  These options must be set in ice_in to run them:

    ice_ic         = 'default' or 'none'
    restart        = .false.

Ran the base_suite; results are at https://github.com/CICE-Consortium/Test-Results/wiki/e203dcbf71.pinto.intel.180608.172808

pinto_intel_restart_gx1_40x4_debug_droundrobin failures are because this case asks for 10 nodes, but the maximum number of nodes that can be requested is 5, for the standard qos on pinto.  This will always fail on pinto.  The other failure is for a box configuration, which still needs to be debugged.

This PR should be ready to merge.

njeffery:
@eclare:  Yes.  I can create binary versions.

PR #149, opened 2018-06-09
mask restoring variables

eclare108213:
The restoring variables needed to be masked by land, and restart_ext must be true when using open boundaries, so that the halo cells are saved in the restart files.

- Developer(s): E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?  BFB except for the test that was failing.

- Other Relevant Details:
Addresses issue #138.  I still need to run the base_suite.


eclare108213:
travis is failing on the builds because I'm multiplying real arrays by tmask, which is a logical array.  I need to use hm instead, or convert tmask to real (double) locally.

eclare108213:
@apcraig, this should be ready to merge now

PR #157, opened 2018-07-03
Allocatable version

mhrib:
- Developer(s):
Mads Hvid Ribergaard, DMI

- Other Relevant Details:
Code, that makes the following configurable via. ice_in namelist:
        max_blocks  , & ! max number of blocks per processor
        block_size_x, & ! size of block in first horiz dimension
        block_size_y, & ! size of block in second horiz dimension
        nx_global   , & ! i-axis size
        ny_global       ! j-axis size

- Files modified:
cicecore/shared/ice_domain_size.F90
cicecore/shared/ice_arrays_column.F90
cicecore/drivers/cice/CICE_InitMod.F90
cicecore/drivers/hadgem3/CICE_InitMod.F90
cicecore/drivers/cesm/CICE_InitMod.F90
cicecore/cicedynB/infrastructure/ice_grid.F90
cicecore/cicedynB/infrastructure/ice_restoring.F90
cicecore/cicedynB/general/ice_flux.F90
cicecore/cicedynB/general/ice_state.F90
cicecore/cicedynB/general/ice_forcing.F90
cicecore/cicedynB/general/ice_forcing_bgc.F90
cicecore/cicedynB/general/ice_flux_bgc.F90


eclare108213:
@mhrib Can you update your fork to eliminate (most of) the code conflicts?  Should we expect these changes to be bit-for-bit?  Did you try to run any of the test suites?  I realize that you may have created the PR so that Rob can get the code easily, but if making the code arrays allocatable is desirable more generally (@dabail10 @rallard77 @JFLemieux73 @proteanplanet @apcraig please weigh in), then we should prioritize getting it merged.  The documentation will need to be updated too.

apcraig:
I like the idea of moving to dynamics arrays.  That way the same binary can be reused for different grids and decompositions.  Also, it probably ultimately allows for more flexibility in coding.

But, I think we need to demonstrate bit-for-bit and if not, then understand why.  We also need to evaluate performance.  I think a 10% or 20% degradation (which I don't expect) would be OK, but we should look at the performance when we have it and then discuss.

PR #161, opened 2018-07-18
Update qc per roberts2018

mattdturner:
Update the QC script to include modifications defined in Roberts, A. F., E. C. Hunke, R. Allard, D. A. Bailey, A. P. Craig, J-F. Lemieux, M. D. Turner (2018):  Quality Control for Community Based Sea Ice Model Development, Philos. Trans. Royal Soc. A, accepted, doi:10.1098/rsta.2017.0344.

- Developer(s):  Matt Turner

- Other Relevant Details:
Modifications included:
- Include the new files that have critical t-value lookup tables.
- Add new function that plots CICE data on a map.
- Update the mask to properly ignore points with no sea ice
- Add a function to create area-weighted fraction of failure points, and compare to a critical fraction
- Add additional comments to the script.
- Remove all CDash-related code


eclare108213:
@proteanplanet  Andrew, could you clarify this point for Matt, please?  thx

proteanplanet:
Hi Matt,  

To pass the 2SPT test, less than half the domain has H1 confirmed in a grid cell.  That is, t>=t_crit in either the first stage or second stage test in more grid cells, weighted by their area, than not. 

In the first stage test, the lag-1 autocorrelation, r1, then the effective sample size n_eff is calculated, and t_crit comes from the standard t lookup table.  In the second test, the sample size n is used, and t_crit comes from our computed table that also needs a value of r1 if n_eff is less than 30 and H0 is confirmed. Once these results have been tallied for the entire domain, providing the area-weighted fraction of the sea ice zone with H1 confirmed is less than 0.5, then  2SPT passes.  

It somewhat muddies the waters if we talk about the test passing or failing at each grid point, so that is why I talk in terms of H0 or H1 being confirmed in each grid cell, and then determine whether or not the 2SPT test passes or fails for the entire (hemispheric) domain after first calculating the answer for each grid point.

Please let me know if that is as clear as mud.

- Andrew

mattdturner:
@proteanplanet , thanks for the clarification.  I just updated the qc script per your comments.  I believe that the script is now ready for testing.  

So far, I have tested it using 2 identical simulations with the data from 1 being scaled by a random value.  I'd be happy to run further tests, unless somebody else would like to.  Are there any suggestions on simulations that I should run in order to further test the script?

eclare108213:
It seems that we need a test to test the test... 
Would it be possible to create a test for the QC scripts which pulls earlier versions (hashes) of the code for which we know the script worked in the past, and does the QC comparisons and produces output from them?  E.g. there could be 3 hashes from the master branch, 2 BFB and a third one which produces a non-BFB but non-climate-changing simulation, and then run one of those 3 again with a namelist change that produces a climate-changing result.  Is this getting to complicated?  I was thinking of creating these runs myself and then thought maybe we ought to automate that.

mattdturner:
I think it would be easier to automate if we could have 2 BFB runs and 1 non-BFB run for a given hash by changing the namelist options.  I'm not (yet) sure how we would automatically scroll through recent hashes to find changes that were non-BFB but non-climate-changing.  I also do not know which namelist configurations we would use for the 3 runs, but it would be easiest to run tests with varying namelist options instead of varying hashes.

eclare108213:
Good point.  It's easy to find alternative namelist configurations that change the 'climate' of the model, e.g. evp and eap.  For BFB I guess we could just run the same hash twice, with a different case name.  It's harder to find namelist parameters that change the answers but not the climate.  The most obvious one for that would be the time step...  Could change dt from 3600 to 1800 and npt from N to 2N (and fingers crossed that it actually works correctly).  If we also compare log files then diagfreq would also need to change with dt.

apcraig:
I'm happy to help create a test for this or @mattdturner probably can too.  It should not be difficult.  How much would this cost?  We would have to run 3 x 5 years on gx3?  

Another thing we could think about doing is providing a more complete qc test script for actually carrying out the qc test on some code changes.  That could either be a separate script or a test on it's own.  In that case, we need to have a way to specify the baseline code version, have the test check it out, run it first, then run the "new" code, and do the qc checks.   It might do something like the "end to end" testing described in https://cice-consortium-cice.readthedocs.io/en/latest/user_guide/ug_testing.html#end-to-end-testing-procedure.  I think these are two different things, testing the qc test, and making the qc test easier to run for a developer.

eclare108213:
I think both of those things would be helpful.  The second one (easier QC testing) could actually cover the first one (testing the QC scripts), with just a few suggestions from us for how to apply the scripts.  This part of our testing suite is pretty daunting for me, so anything we can do to make it easier for the user (and me!) is most welcome.

eclare108213:
Regarding cost, it would be 3x5 years on gx3 (if we copy the output from one of the runs into a different case for the BFB comparison, otherwise 4x5 years), but we'd only need to run it when we change the QC scripts, not as part of our regular testing.

mattdturner:
To start, I have done 4 runs on Conrad:

1. Base configuration with the `-s qc` option and `--testid qc_base`
2. BFB test with `-s qc` and `--testid qc_bfb`
3. Non-BFB but not climate changing test with `-s qc_nonbfb` (i created the new `set_nml.qc_nonbfb`, where dt was halved and npt was doubled) and `--testid qc_test`
4. Non-BFB but climate changing test with `-s qc,alt02` and `--testid qc_fail`

I then ran the QC script on the following combinations:

- `qc_base` and `qc_bfb`:  The test passed, reporting that the data is bit-for-bit
- `qc_base` and `qc_nonbfb`:  The test passed, reporting that the two-stage test passed and the skill test passed in both hemispheres
- `qc_base` and `qc_fail`:  The test failed, reporting that the 2-stage test passed and the skill test failed (skill score of 0.93 in northern hemisphere and 0.95 in southern hemisphere)

So, it looks like the QC test is working properly.  

I think that the work towards automating the validation of the QC test script should be included in a separate pull request.  However, I think creating a script to go through the processes I just went through wouldn't be difficult.  It might need to be setup as 2 separate scripts, 1 to build the tests and submit them, and 1 to run the QC script for the 3 combinations.

eclare108213:
This is great, thank you @mattdturner.  Please add an issue to automate the QC tests (with one script or two, whatever works).  I'll merge this PR.

PR #179, opened 2018-08-24
Docs

duvivier:
Lots of groundwork for documentation to be correct for upcoming release.

- Developer(s): Alice DuVivier

- Are the code changes bit for bit, different at roundoff level, or more substantial? b4b

- Other Relevant Details:
Major fixes:
- Changed authorship of documentation to the CICE Consortium from me personally
- Updated the labeling of tables/figures to be consistent with Icepack and so that PDF will be rendered correctly.
- Updated Readme.md to include language on testing.


eclare108213:
@duvivier did you check that all references on the wikis and in the documentation to the pdf have been changed?  We might have gotten them all when we redesigned and created the 'Resources' web page.

duvivier:
@eclare108213 We should not have to change any references on the wiki or documentation to point the new PDF since we created the "Resources" page. 

If you look at the current "master" PDF build of our documentation on Readthedocs you can see the main issues I'm fixing. https://media.readthedocs.org/pdf/cice-consortium-cice/master/cice-consortium-cice.pdf
- I'm the author (cover page)
- The index is all funky (pg.i)
- Table and Figure numbering is odd (e.g. pg.7; pg.15)

I checked on my local machine and it looks like all the in text references are being done correctly still.

duvivier:
@apcraig Yes, you are right. I guess I was confused about the versioning and was trying to get it in line with the Icepack versioning but forgot that when creating a release the cice.setup --setvers script automatically changes conf.py through the set_version_number.csh. I added a note under the release process wiki page because I looked there and still didn't realize it till I delved more into the scripts.

I modified the description of html_title as well. Right now we don't actually use this and the documentation uses the default of: # "<project> v<release> documentation".
I just thought without specifying this people (me?) could get confused again in the future. We should probably do the same thing for Icepack in the conf.py file since I looked there and this is partially why I got confused.

PR #180, opened 2018-08-28
Dummy and unused variables.

dabail10:
This is just the removal or commenting out of a number of dummy arguments and unused variables. I have in some comments some uninitialized variable changes (answer changing) that will come in a later tag. However, everything in this pull request is BFB with the NAG compiler. There is a lot here and I am happy to discuss these.

- Developer(s): D. Bailey

- Are the code changes bit for bit, different at roundoff level, or more substantial? BFB

- Other Relevant Details:


apcraig:
I would like to run a full test suite on conrad with these changes before I approve.  Will set that up now.

eclare108213:
Thanks @apcraig.  Is there a way to set a check-uninitialized flag for most if not all of the runs?

apcraig:
I want to first check whether everything runs fine and is bit-for-bit.  We build and run on 4 compilers on conrad.  Happy to look into uninitialized variables as a separate step.

dabail10:
With the initialization part, I was holding these off for a separate tag. This relies on some Icepack changes too. I agree that commenting out the BGC variables is misleading, but I left them commented so when they are actually used, they can easily be added back.

dabail10:
For the NAG compiler I used the -nan flag. Intel is supposed to use -check uninit, but this did nothing for me.

Dave

dabail10:
So, this is a concern. The idea is that these function calls return an error status, but nothing was done with them. I added error handling for this purpose. Doesn't this highlight an issue in the rake algorithm?

apcraig:
In this case, I think it just means there is a temporary distrb type created that is not necessarily used entirely.  I think it just stores a subset of information temporarily.  We could add an arbitrary allocate on the array in the data structure that is not currently used.  Or we could remove the deallocate istat checks.  Either is acceptable to me.  The former is more robust, the latter is more flexible.  I guess I vote for the former if I have to vote.

dabail10:
Can we merge this yet?

apcraig:
@dabail10, we just need the istat checks in the distributionDestroy routine turned off.  The others in the rake routine should/can remain.  Also, please confirm that you tested the drakeX2 option in your testing.  thanks.

dabail10:
Ok. What is the difference between a drakeX2 and a rakeX2?

Dave

apcraig:
drakeX2 is the test option and will appear in the testname.  It turns on the rakeX2 option in namelist.

dabail10:
So, the base_suite does a rakeX2 in the cheyenne_intel_decomp_gx3_4x2x25x29x5 test. Is this enough?

Dave

apcraig:
I believe a rakeX2 test in the decomp test is adequate.  thanks.

PR #187, opened 2018-09-20
update pgi compiler optimization to address reproducibility problems

apcraig:
Update pgi compiler optimization, switch from O2 to O. 

- Developer(s): tcraig

- Are the code changes bit for bit, different at roundoff level, or more substantial? NOT bit-for-bit for pgi compiler, but this is good.

- Other Relevant Details:

Confirmed that O0, O1, and O all generate results that are reproducible as expected (different block sizes, etc).  O2 does not.  O is between O1 and O2 and seems to actually perform better than O2 on conrad and gordon.

### DIFF
--- a/configuration/scripts/machines/Macros.conrad_pgi
+++ b/configuration/scripts/machines/Macros.conrad_pgi
@@ -14,7 +14,7 @@ FFLAGS_NOOPT:= -O0
 ifeq ($(ICE_BLDDEBUG), true)
   FFLAGS     += -O0 -g -Mbounds -Mchkptr
 else
-  FFLAGS     += -O2
+  FFLAGS     += -O -g
 endif
 
 ifeq ($(ICE_COMMDIR), mpi)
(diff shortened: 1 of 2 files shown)


apcraig:
@eclare108213, I would like to get this onto the master in the next day so it's part of the weekend testing on conrad and gordon.  This will establish a full set of new baselines for the pgi compiler to make it easier to test other change (like dynamic allocation).  This will change answers which I believe is a good thing.

Feel free to review and Pull, this is ready and has been tested, hash 9a08fd81 on https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks

apcraig:
The -g doesn't hurt for these runs.   I would leave -g for now.  pgi is so unreliable as it is.

Also, I have change the pgi settings for all our pgi machines, conrad and gordon.  Those are the only machines we test pgi on right now.

Issue #159, opened 2018-07-04
gravit not defined in ice_dyn_shared?


JFLemieux73:
In ice_dyn_shared we have:

#ifndef coupled
         ! calculate tilt from geostrophic currents if needed
         strtltx(i,j) = -fm(i,j)*vocn(i,j)
         strtlty(i,j) =  fm(i,j)*uocn(i,j)
#else
         strtltx(i,j) = -gravit*umass(i,j)*ss_tltx(i,j)
         strtlty(i,j) = -gravit*umass(i,j)*ss_tlty(i,j)
#endif

I run CICE6 non-coupled so my code compiles and works fine. However, I have the impression it would not work in coupled mode because gravit is not defined. 

Do we need a 'call icepack_query_parameters'?

eclare108213:
@apcraig, would you check this in the version of RASM that includes CICE v6, please?  Maybe you already fixed it.

apcraig:
It looks like neither RASM nor CESM set the "coupled" cpp.  They both use the CESMCOUPLED cpp.  The coupled cpp is used in limited places in CICE5 and CICE6.  We should review that cpp and decide whether it should be removed, changed, or otherwise.  The cpp appears in 

cicecore/cicedynB/infrastructure/comm/serial/ice_communicate.F90
cicecore/cicedynB/dynamics/ice_dyn_shared.F90
cicecore/cicedynB/general/ice_init.F90
cicecore/drivers/cesm/*/*.F90

dabail10:
Actually, we do define "coupled" in the CESM build and use both CESMCOUPLED and coupled. We do the following in ice_init.F90:

#ifdef CESMCOUPLED
#define coupled
#endif 

We need both for various pieces of code. I agree that we do need to add the parameter gravit for the coupled case. I will do this.

Dave

dabail10:
This is addressed in PR #182.

dabail10:
Resolved in PR #191.

Issue #97, opened 2018-02-24
snow, ice evaporation


eclare108213:
Check for potential bug.  This might be easier (or already fixed) as part of the CMIP6/SIMIP history output (#93).  Alex West's original email:

Hello Elizabeth,
 
Hope that you’re well.  I’ve been having a go at adding two new diagnostics to CICE to report ice and snow evaporation separately – apart from being in the SIMIP data request, I think they may be useful for my PhD on the Arctic energy budget.  However, I’ve been having some unexpected problems, and wondered if there are aspects of the evaporation I’m still not aware of (and if I’ve come across a possible reason why this hasn’t been done before!).
 
Basically, I’ve
 
-          defined the new diagnostics, evap_ice_ai and evap_snow_ai, in the history routines in the normal way;
-          defined new aggregate and category fields evap_ice, evapn_ice etc in ice_flux;
-          added code to thickness_changes in ice_therm_vertical to calculate the fields evapn_ice, evapn_snow in the same way as evapn, and divide them by timestep length;
-          added the new evaporation fields to the routine merge_fluxes in in_flux.
 
It looks to my naive eye that the new fields are then calculated in much the same way as the existing evapn field.  However, when I look at the resulting diagnostic files, the new snow and ice evaporation fields are of a completely different order of magnitude to the total evaporation field (about 1.e3 bigger).  Moreover, the sum of the two new fields actually has a different structure to the total evaporation field.
 
Do you have any idea what I might be missing here?  I attach the copies of ice_flux, ice_step_mod, ice_therm_vertical, ice_history and ice_history_shared from my branch; unfortunately the output file with the three evaporation fields is too large to send by email.
 
Because this branch is a general ‘new diagnostics for SIMIP’ branch it also contains Dave Bailey’s new diagnostic changes, so a diff with the 5.1.2 trunk will show these differences also.

dabail10:
I have added two new variables in CESM with the SIMIP request to do this. So, I believe this will be resolved when I bring the CMIP variables in.

dabail10:
This is addressed in PR #182.

dabail10:
Resolved in PR#191.

Issue #93, opened 2018-02-24
CMIP6/SIMIP history output


eclare108213:

dabail10:
This is addressed in PR #182.

dabail10:
Resolved in PR#191.

Issue #63, opened 2017-12-26
lonu_bounds uninitialized


eclare108213:
the NAG compiler reports lonu_bounds is uninitialized for iconrner=1, iblk=1, at line 1912 of ice_grid.f90. 
(issue reported by DMI collaborators)

dabail10:
This is will be fixed in a new PR coming soon. Since these are just history variables, we don't use the ghost cells. None-the-less, it is still good to initialize these.

Dave

dabail10:
This is addressed in PR #180.

dabail10:
Resolved in PR #191.

PR #193, opened 2018-09-24
Cmip2

dabail10:
This PR includes the following. Note these rely on the CMIP updates from Icepack as well.

Update all of the CMIP history variables #93

Update evaporation over snow and bare ice. #97

Also, some uninitialized variables. #63

Adds gravit to ice_dyn_shared.F90 when #coupled is activated. #159

There is a new history namelist flag f_CMIP that activates all of the CMIP history in one go.

Developer(s): D. Bailey

Please suggest code Pull Request reviewers in the column at right.

Are the code changes bit for bit, different at roundoff level, or more substantial? BFB

Other Relevant Details:


apcraig:
This looks the same as the cmip branch.  @dabail10, can you look thru the code carefully and make sure things are what you expect.  I almost wish you would have left the istat calls on the cmip branch.  That was a really obvious mod to check different git processes for differences in PRs.

There are some things that seem either reverts from the dummy2 branch (which we don't want) or something different than the CMIP mods.  The changes in ice_dyn_shared, ice_step_mod lines 80-83, ice_grid, and CICE_RunMod all seem somewhat unrelated to the CMIP changes.  So, are they mods that fix other things?  Are they mods that shouldn't be there?  Should there be equivalent mods in other CICE_RunMod files other than cice (ie. cesm)?

dabail10:
I thought I had detailed everything above in the comments. There some initialization issues and other relatively small changes that I figured I could combine with the CMIP changes. If this is not ok, I can try to add them one at a time. The issue is with all the cumulative changes and PRs sitting around, I just kept dealing with a few of the minor issues.

apcraig:
That's all fine, I'm not saying they shouldn't be combined.  I just want for you to look over the code changes and make sure you don't see anything "extra" or notice anything "missing".  I don't have a good idea whether this is exactly what should be in this PR or whether we still have some dummy2 merge issues or something else.  I am not convinced we've found a robust process to deal with code like this.  That's what I would like to figure out if possible.  I don't like thinking we have to manually check all code and fix missing/extra merge stuff manually.  The first step is to decide whether this PR is correct as best as you can tell.  If not, then we can discuss next steps.

dabail10:
This was the same as the CMIP PR.

PR #191, opened 2018-09-24
Update CICE with CMIP changes.

dabail10:
This PR includes the following. Note these rely on the CMIP updates from Icepack as well.

1. Update all of the CMIP history variables #93 

2. Update evaporation over snow and bare ice. #97 

3. Also, some uninitialized variables. #63 

4. Adds gravit to ice_dyn_shared.F90 when #coupled is activated. #159 

There is a new history namelist flag f_CMIP that activates all of the CMIP history in one go.

- Developer(s): D. Bailey

- Please suggest code Pull Request reviewers in the column at right.

- Are the code changes bit for bit, different at roundoff level, or more substantial? BFB

- Is the documentation being updated with this PR? (Y/N) N
If not, does the documentation need to be updated separately at a later time? (Y/N) Y
Note: "Documentation" includes information on the wiki and .rst files in doc/source/, 
which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/.

- Other Relevant Details:

PR #43, opened 2017-12-02
Testing documentation update.

dabail10:
Developer(s): D. Bailey
Updated documentation (Y/N): Y
Results (bit for bit, roundoff, climate changing): bfb
Code review: E. Hunke, A. Roberts
Other Relevant Details:
The Makefile is still working with my software stack. There are still some tweaks in the txt files
that I will work on, but overall the RST is close here.

PR #195, opened 2018-10-01
adding CMIP variable documentation

duvivier:
Documentation about addition of CMIP6/SIMIP variable inclusion in history variables. Follow up to PR #191 

- Are the code changes bit for bit, different at roundoff level, or more substantial?  bfb

- Please include the link to test results or paste the summary block from the bottom of the testing output below. n/a

- Does this PR create or have dependencies on Icepack or any other models? no

- Is the documentation being updated with this PR? Y - follow up to #191 

- Other Relevant Details:

### DIFF
--- a/doc/source/developer_guide/dg_documentation.rst
+++ b/doc/source/developer_guide/dg_documentation.rst
@@ -142,7 +142,7 @@ Here are some resources for using RST files:
 Building documentation
 ~~~~~~~~~~~~~~~~~~~~~~
 
-Once you've committed and pushed changes to the documentation *.rst files on your personal development fork. 
+Once you've committed and pushed changes to the documentation `*.rst` files on your personal development fork. 
 Go to your readthedocs.org site and then select your project "Overview". Whenever you commit to your fork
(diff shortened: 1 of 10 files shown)


duvivier:
@dabail10 I added the correct reference for SIMIP you suggest. 

@eclare108213 @dabail10 (@proteanplanet ). I noticed some inconsistencies/duplicates in the bibliography while adding the above reference. I have also fixed these and the relevant citations in text. We have two references in master_list.bib that are unused anywhere in the documentation- Roberts2018 and Irving2015. It appears these were added as part of #43 from last November and are related to testing developed by Andrew but was a commit by @dabail10. 
@proteanplanet , if you are able to quickly peruse the section that was added with #43 (https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_testing.html#two-stage-paired-thickness-test) and check if these references are indeed missing and should be added, that would be helpful. 

Finally, I just cleaned up some other formatting bugs that were giving errors in sphinx and causing minor oddities in the documentation.

PR #203, opened 2018-10-10
References and testing in documentation

duvivier:
Updating documentation about testing procedures as well as changing how reference format.

- Developer(s): Alice DuVivier

- Are the code changes bit for bit, different at roundoff level, or more substantial?  bit for bit

- Please include the link to test results or paste the summary block from the bottom of the testing output below. N/A

- Is the documentation being updated with this PR? Y
If not, does the documentation need to be updated separately at a later time? N/A

- Other Relevant Details:

Based on my experience with testing procedures I have modified the user's guide section on testing. Mainly I have tried to add some text at points where I got confused or lost. I also made the examples a list in a subsection instead of each their own subsection because the navigation got really complex that way.

I also changed how the bibliographic items are referenced in text because I noticed the past several documentation pull requests were not consistent and it resulted in some of the references being added twice and referenced differently in the text. I also provided an "example" for how people should add references to the bibliography and emphasized adding references in chronological then alphabetical order so we can better keep track of them and avoid repetition.

@proteanplanet still needs to weigh in on if Irving2015 and Roberts2018 should be cited as they are not at the moment.


apcraig:
Did you verify this builds fine on readthedocs?  Assigned the merge to @duvivier in case she wants to wait for AR to reply before merging.

duvivier:
@apcraig I will wait to see if @proteanplanet replies in the next day or so. It should be a quick check hopefully.

PR #206, opened 2018-10-12
Corrected k1 parameter for grounding scheme

JFLemieux73:
[Remove this and add a short summary line]:

- Developer(s): JF Lemieux

- Are the code changes bit for bit, different at roundoff level, or more substantial? NA

- Other Relevant Details:

k1 was 20. It is now set to the rigth value (k1=8). 

I also removed a comment in ice_dyn_evp.F90 saying that we save the replacement pressure for principal_stress (not used anymore).

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
@@ -691,8 +691,7 @@
          endif
 
       !-----------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


apcraig:
I'm fine with this change, but just want to make sure @eclare108213 has a quick look in case we need to defer or otherwise.  I assume this will change answers for at least a couple tests and that we don't need to run the qc or otherwise beforehand.

PR #215, opened 2018-10-19
Update scripts, code, tests to fix handful of issues including parse namelist checking

apcraig:
Update scripts, code, tests to implement a few important features

- Developer(s): tcraig

- Other Relevant Details:

Add check that namelist parsing strings are valid and in ice_in.

Workaround atan2 bug on nag on hobart.

Update poll_queue script to catch all jobs, stopped working correctly when executable reuse was introduced.

Turn off some questionable OMP loops.

Fix OMP loop in ice_restart_driver.

Update set_nml.gx1, turn off ocn_data again.

Add qstat env variable for hobart.

Add workaround for ice_distribution clean for unallocated datatype variables.

Add new diagnostic to report_results associated with non-fail, non-pass test info and set default test result to FAIL.


apcraig:
I am running the test suite now and will have results shortly.

apcraig:
The atan2 seems to just be an issue on hobart with nag.  It has been a problem for a while, we just hadn't addressed it.  I am not happy about adding it, but it seems to help some of the test results on hobart.  It should not be needed.

configuration/scripts/ice_in
```diff
@@ -171,11 +171,11 @@
     fe_data_type    = 'default'
     fyear_init      = 1997
     ycycle          = 1
-    atm_data_dir    = '/glade/u/home/tcraig/cice_data/'
-    ocn_data_dir    = '/unknown_ocn_data_dir'
-    bgc_data_dir    = '/uknown_bgc_data_dir'
     atm_data_format = 'bin'
+    atm_data_dir    = '/glade/u/home/tcraig/cice_data/'
```
duvivier: Should this really point to your directory, @apcraig ?
ICE_MACHINE_INPUTDATA/CICE_data/ might be better.

apcraig: I noticed that too.  It's been that way for a while.  I'll fix it sometime.  It's always overridden so I don't think it's a problem.  But I agree better to have something more generic there.

PR #184, opened 2018-09-15
Revised EVP updated

mhrib:
Revised EVP parameters, according to Kimmirtz, Danilov and Losch, 2015

- Developer(s): Till Rasmussen and Mads Hvid Ribergaard, DMI

- Please suggest code Pull Request reviewers in the column at right.
Elisabeth?, Tony?

- Are the code changes bit for bit, different at roundoff level, or more substantial?
With *conservative* settings - AND revised_evp=.false. - it should output Bit-for-bit identical.
With revised_evp=.true. the settings has changed. Therefore obvious NOT bit-to-bit identical

- Is the documentation being updated with this PR? (Y/N) N
If not, does the documentation need to be updated separately at a later time? (Y/N) ??
Note: "Documentation" includes information on the wiki and .rst files in doc/source/, 
which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/.

- Other Relevant Details:

affected files:
./cicecore/cicedynB/dynamics/ice_dyn_evp.F90
./cicecore/cicedynB/dynamics/ice_dyn_shared.F90 (for namelist)
./cicecore/cicedynB/general/ice_init.F90 (for namelist)

Namelist (these are the default values, ONLY relevant for revised_evp=.true.):
&dynamics_nml
brlx = 300.0
arlx1i = 1.0/300.0   

NOTE: The values for "brlx and arlx1i" are overwritten in ice_dyn_shared for revised_evp=.false.

----

Have tested identical md5sum's for 24 iterations with GX3 using "revised_evp=.false.".
GNU compiler (Cray XT50) with the following FFLAGS:

FFLAGS := -g -O0 -fsignaling-nans -fno-reciprocal-math -ftrapping-math -fno-associative-math -fno-unsafe-math-optimizations

PR #218, opened 2018-10-24
Revised EVP updated v2

mhrib:
Update of:
https://github.com/CICE-Consortium/CICE/pull/184/

--o--

Revised EVP parameters, according to Kimmirtz, Danilov and Losch, 2015

- Developer(s): Till Rasmussen and Mads Hvid Ribergaard, DMI
    Elisabeth?, Tony?, Jean-Francois?
- With conservative settings - AND revised_evp=.false. - it should output Bit-for-bit identical.
    With revised_evp=.true. the settings has changed. Therefore obvious NOT bit-to-bit identical
- If not, does the documentation need to be updated separately at a later time? (Y/N) ??

- Other Relevant Details:

affected files:

&dynamics_nml
brlx = 300.0
arlx1i = 1.0/300.0

NOTE: The values for "brlx and arlx1i" are overwritten in ice_dyn_shared for revised_evp=.false.

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
@@ -40,7 +40,7 @@
           p2, p222, p25, p333, p5, c1
       use ice_dyn_shared, only: stepu, dyn_prep1, dyn_prep2, dyn_finish, &
           ndte, yield_curve, ecci, denom1, arlx1i, fcor_blk, uvel_init,  &
-          vvel_init, basal_stress_coeff, basalstress, Ktens
+          vvel_init, basal_stress_coeff, basalstress, Ktens, revp
       use ice_fileunits, only: nu_diag
       use ice_exit, only: abort_ice
       use icepack_intfc, only: icepack_warnings_flush, icepack_warnings_aborted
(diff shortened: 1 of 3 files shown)


JFLemieux73:
Thanks Mads and Till.

I will test this today and will update the documentation.

JFLemieux73:
Mads and Till,

I compared the consortium master code (old revised evp) and the new one revpv2.

1) With revised=.false. the master code and revpv2 give BFB results.

As suggested by Elizabeth, I also compared evp to revp. I ran gx3 (1 year) experiments. They start in January and I compare the mean thickness fields in December. I used ndte=1200 to have more convergence of the solutions.

2) master code: experiment with evp - experiment with revp = differences of hi are between -0.06 and 0.08 m. Ok.

3) revpv2 code: experiment with evp - experiment with revp = differences of hi are between -0.6 and 0.7 m...the differences are 10 times larger. 

I am not sure why I see these bigger differences in revpv2. Could you please confirm these results?

Note that I didn't change the alpha and beta parameters for both codes...I just took the default values.

TillRasmussen:
I have checked the revised evp v3. I do get similar differences between the two solutions after 1 year simulation.
(-0.71610379, 0.72629881) Min max diff ice thickness
(-0.011542901, 0.0096831024) min max ice concentration.
See spatial distribution
aice
I used 1200 subcycles like you did.
There is one thing that I would like to test and that alpha and beta are predefined in the revised_evp = False, A similar value could be chosen for the revised_evp = true.

JFLemieux73:
Thanks Till. Do you also see smaller differences if you compare evp and revp with the current master code (with the old formulation of revp)?

TillRasmussen:
Yes I do see a smaller difference both between evp and revp from the master branch
master (min,max)
(-0.055969715, 0.068563938) ice thickness
(-0.005423367, 0.0094287395) ice concentration
At the same time the difference between evp of the revp3 branch and the master branch is relatively small.
evp revp3-master (min,max)
(-0.035111904, 0.039922237) ice thickness 
(-0.0025034547, 0.0005081296)  ice concentration

eclare108213:
superseded by #219

eclare108213:
We can continue this conversation here, even though the PR is closed and #219 is current.
I agree with JF, that we shouldn't merge the new rEVP changes when the thickness difference is so big, until we understand them.  Especially if the new rEVP isn't converging!  The differences that @TillRasmussen cites above are on the order of what I would expect for EVP.

PR #219, opened 2018-10-25
Revised EVP updated v3

mhrib:
Update of:
#218 (and #184)

As #218, but added rEVP variables "brlx" and "arlx1i" to ice_in file:
configuration/scripts/ice_in

--o--

Revised EVP parameters, according to Kimmirtz, Danilov and Losch, 2015

- Developer(s): Till Rasmussen and Mads Hvid Ribergaard, DMI

    Elisabeth?, Tony?, Jean-Francois?

    With conservative settings - AND revised_evp=.false. - it should output Bit-for-bit identical.
    With revised_evp=.true. the settings has changed. Therefore obvious NOT bit-to-bit identical

- If not, does the documentation need to be updated separately at a later time? (Y/N) ??

- Other Relevant Details:

affected files:
./configuration/scripts/ice_in

&dynamics_nml
brlx = 300.0
arlx1i = 1.0/300.0

NOTE: The values for "brlx and arlx1i" are overwritten in ice_dyn_shared for revised_evp=.false.

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
@@ -40,7 +40,7 @@
           p2, p222, p25, p333, p5, c1
       use ice_dyn_shared, only: stepu, dyn_prep1, dyn_prep2, dyn_finish, &
           ndte, yield_curve, ecci, denom1, arlx1i, fcor_blk, uvel_init,  &
-          vvel_init, basal_stress_coeff, basalstress, Ktens
+          vvel_init, basal_stress_coeff, basalstress, Ktens, revp
       use ice_fileunits, only: nu_diag
       use ice_exit, only: abort_ice
       use icepack_intfc, only: icepack_warnings_flush, icepack_warnings_aborted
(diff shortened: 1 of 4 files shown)


apcraig:
Does this PR replace #184 and #218.  If so, can we close those PRs?  I'm fine leaving them around if we want to reference them or the conversations in them.  Just wondering.

mhrib:
Hi Tony, JF and Till,

Yes, #219<https://github.com/CICE-Consortium/CICE/pull/219> replaces
#218<https://github.com/CICE-Consortium/CICE/pull/218> and
#184<https://github.com/CICE-Consortium/CICE/pull/184>

The only reason to keep #218 + #184 is to keep the conversations ... if they are needed/important for later use?

Admit: I am a bit lazy, so I did the changes similar to what we did with the "alloc" branch, where you afterwards deleted the "not used" PR's. I do not know how to close/delete a PR or if I have permissions to do so?

rgs
Mads



________________________________________

eclare108213:
We can close the others -- they (and their conversations) will still be accessible on the 'closed PR' list and through the referenced links above.

JFLemieux73:
Till and Mads,

Is it just a question of different parameters (brlx and arlx1i) in both versions of revp? 

When we compare the old revp and the evp we get small differences in thickness. 

In you use the paremeters from the old revp in the new (revpv3) do you see similar thickness differences with the evp?

jf

TillRasmussen:
I will try to do that. I have had the same thought. 

Another thing which really puzzles me is the terms that include beta in the stepu function

These are
 cca = (brlx + revp)*umassdti(i,j) + vrel * cosw + Cb
and 
   cc1 = strintx(i,j) + forcex(i,j) + taux &
             + umassdti(i,j)*(brlx*uold + revp*uvel_init(i,j))
         cc2 = strinty(i,j) + forcey(i,j) + tauy &
+ umassdti(i,j)*(brlx*vold + revp*vvel_init(i,j))
They all include an addition of brlx and revp (cc1 and cc2 includes multiplication of velocities).  revp is a binary 0/1 converted to float that changes the formulation between the two evp versions. brlx is a much larger number that in reality means that one should be able to remove these. This is not new to the implementation.

eclare108213:
@TillRasmussen @JFLemieux73 
The derivation of the original implementation of rEVP in CICE is provided here:
https://cice-consortium-cice.readthedocs.io/en/master/science_guide/sg_modelcomps.html#revised-approach
including a couple of notes at the end of that section about how the revp flag is used (and also the difference in initialization strategies, which I think @mhrib commented on).  I haven't looked closely at the changes that you all have made, but our documentation should reflect what's in the code, or be made to do so.  I think @JFLemieux73 has re-derived it in the past few months, and so now would be a good time to make sure both the docs and the code are consistent with the changes that Martin L. recommended.  Thank you for following up!

TillRasmussen:
Thanks,
I think that the problem is solved. I will update tomorrow but the difference between evp and revp is ~5cm  for a one year run with the new code.

apcraig:
Superseded by #226

PR #223, opened 2018-10-31
update icepack

apcraig:
Update Icepack to latest version, 0b4ee4e33 Oct 4, 2018

- Developer(s): tcraig

- Are the code changes bit for bit, different at roundoff level, or more substantial? NOT bit-for-bit (expected)

- Other Relevant Details:

Test results from conrad are here, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks, hash e6b9fee8db55.  All tests pass, but answers change, as expected.

### DIFF
--- a/configuration/scripts/tests/QC/cice.t-test.py
+++ b/configuration/scripts/tests/QC/cice.t-test.py
@@ -83,7 +83,7 @@ def read_data(path_a, path_b, files_a, files_b, ni, nj):
     def fill_data_array(path, files, nj, ni):
         '''Function to fill the data arrays'''
         # Initialize the data array
-        data = np.zeros((len(files), nj, ni))
+        data = np.zeros((len(files), nj, ni),dtype=np.float32)
         # Read in the data
         logger.debug('Reading in data for files in %s', path)
         cnt = 0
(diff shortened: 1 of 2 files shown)


eclare108213:
@apcraig, did you run the QC tests?  Since the answers are changing, it would be nice to document the results from that here, now, for future reference.

apcraig:
I have not run the qc tests.  How does that work when all the configurations have changed answers?  Do I need to run multiple configurations thru the qc?  What configuration should I test?

eclare108213:
Good question.  We should choose a test or set of tests that exercise the part of the code that changed.  It looks like there have been 2 Icepack merges that were not BFB since the last update, 
https://github.com/CICE-Consortium/Icepack/pull/222 (transport time level inconsistency)
https://github.com/CICE-Consortium/Icepack/pull/227 (shortwave intent inout).
For QC, a single gx3 smoke test was run for 222 but not for 227.   I suggest doing the same thing as in 222 (look at the PR) here, and if it does not pass then dive deeper, e.g. run QC for 227 alone.  

We need to think about whether we should do other tests from the base suite for QC -- certainly that would be necessary when the changes are in parts of the code that are not exercised in the basic gx3 smoke test.  Both of these changes are exercised there, so I think it should be enough for now.

apcraig:
I'm working on some qc validation.  Hope to have some results soon.

apcraig:
Matt and I are working on the qc testing on conrad, we hit a problem.  Sorry for the delay.

apcraig:
I ran the QC test on the standard configuration compared to the current master and it passed.

Running QC test on the following directories:
  /p/work1/apcraig/CICE_RUNS/conrad_intel_smoke_gx1_32x1_medium_qc.qc_base/
  /p/work1/apcraig/CICE_RUNS/conrad_intel_smoke_gx1_32x1_medium_qc.qc_test/
Number of files: 1825
2 Stage Test Passed
Quadratic Skill Test Passed for Northern Hemisphere
Quadratic Skill Test Passed for Southern Hemisphere
Quality Control Test PASSED

mattdturner:
Everything looks good to me.  I'll continue working on reducing the memory footprint of `cice.t-test.py`, but I think this PR is ready to be merged.

PR #226, opened 2018-11-02
Revp v4

TillRasmussen:
Updated revised EVP - Update of #219 

- Developer(s): Till Rasmussen Mads Hviid Ribergaard, DMI

Jean-Francois Lemieux

Changes are not bit for bit as coefficients within the evp solver has been changed and thereby the convergence has changed.

Test has been been made with the original evp/revised evp and the original evp/modified EVP according to Loesch. This is a follow up to Revised EVP updated v3 #219 which is now obsolete. It has removed a bug that made the new version of revised evp diverge significantly from the others
Results of max differences of monthly averaged december.
new revp - new evp
(-0.043477774, 0.053825378)
(-0.010963917, 0.0040149093)
master revp - master evp
(-0.055969715, 0.068563938)
(-0.005423367, 0.0094287395)
new evp - master evp
(-0.035111904, 0.039922237)
(-0.0025034547, 0.0005081296)
new revp - master revp
(-0.028660536, 0.05374217)
(-0.0085960627, 0.0045653582)
aice

A convergence test should be made. for one time step.

- A one year simulation has been executed based on the GX3 setup including all standard settings with the exception that new parameters (brlx and alrx) has been added to the namelist

This influence ice_init (name list parameters), ice_dyn_shared and ice_dyn_evp.
No 
Probably

- Other Relevant Details:
affected files:
./configuration/scripts/ice_in

&dynamics_nml
brlx = 300.0
arlx = 300.0 Note this differs from previous implementations where the inverse was provided in the namelist. This adds an additional scalar to the code but it also moves it closer to the theory.


JFLemieux73:
Thanks Till. That's great. What was the bug by the way?

I will review this asap and will update the documentation. Could you please do the 1 day test with 20000 and 25000 iterations (ndte)?

TillRasmussen:
The bug was in the ice_in. arlx1i = 1.0/300.0 was read as 1.0
I have tested convergence according to the method you used JF. This means running the code for 1 day. As a difference to the suggestion by JF I have used the snapshot (hisnap) after 1 day.
ndte is in the interval of 200 and 25800 with spacing of 200. (In Python this would be np.arrange(200,200,26000).
alpha = beta has been set to 300, 600 and 900 (see color coding for plot)
The error has been measured as max(abs(HISNAP_{i+1}-HISNAPE_[i})).
There is a clear convergence (note that y is logarithmic) up to ndte = ~2000. After this the convergence is lost. Probably due to noise or limits to accuracy introduced elsewhere in the code. This result is similar to some of the results where Mads measured noise based on compilers. This number will probably be domain dependent.

conv
The last line on the figure is evp. This seems to have reach its minimum error after 200 iterations.

This does not meassure the error from the evp/revp method alone. In order to do so a binary identical input to the routine should be ensured then the error should be measured based on the velocity of the output.

JFLemieux73:
As discussed yesterday I think we should accept this PR.  But before that, Till could you please clean up the parameters (Se, xi, ...) that are not needed anymore with the new revp version?

TillRasmussen:
I have removed Se, Xi, gamma, xmin and ymin. Unless someone objects I would also like to remove dte2T, tdamp2 (only used for calculation of arlxli in traditional evp) 
ecc ( only used for ecc)

eclare108213:
It's fine to remove dte2T, tdamp2, and ecc from the active calculation and declarations, but please leave them in the code as comments, since they relate directly to variables discussed in the Hunke 2001 paper and possibly in the documentation.  Thanks!

eclare108213:
The compile is failing with this error in ice_init.F90:

         write(nu_diag,1006) ' arlx                      = ', arlx
                           1
         Error: FORMAT label 1006 at (1) not defined

The fix is probably simple:  change 1006 to 1005.

eclare108213:
@TillRasmussen @JFLemieux73 
Please, would one of you change the format label in the write statements to 1005, and check that it works?  (see above)

TillRasmussen:
I am testing the update now.. sorry for the delay but I have been occupied all week, Should be there within an hour or so

eclare108213:
Icepack was updated in the meantime, so now it conflicts in this PR.  Sigh.

TillRasmussen:
I realized that it was updated after I commited the changes.... New update will soon be there

apcraig:
Something has gone wrong on this branch.  It now shows conflicts that are too complicated to fix in the web browser.  That can't be good.  The diffs also show a bunch of changes that should already be on the master.  It might be worth branching off the current master again and then merging this branch to the new branch.  That should create a clean branch again.  You would then want to create a new PR from the new branch.  Sorry for the hassles.

TillRasmussen:
As Tony writes there is too much new sand in this sandbox. I have updated a new version. I will do a few short test and publish it. Probably before the end of the weekend. This will be closed when that happens

TillRasmussen:
Closed. Continued in #242

PR #231, opened 2018-11-06
Update gordon and onyx default queues from debug to standard

mattdturner:
Change default queue on Gordon and Onyx from `debug` to `standard`.

- Developer(s):  Matt Turner

- Does this PR create or have dependencies on Icepack or any other models? No

- Other Relevant Details:
This change was made because a few of the tests in the test suites require longer walltimes than permitted on the debug queue.


mattdturner:
That's a fair point.  I was under the impression (which I just noticed is incorrect) that Conrad had `standard` as the default queue, and so was updating other servers to match.  Since that was incorrect, I agree that we can go ahead and close this.

apcraig:
Sounds good.  Thanks.

PR #235, opened 2018-11-09
Refactor tracer initialization

eclare108213:
Cleaning up tracer initialization and namelists

- Developer(s):  E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?   BFB

161 of 165 tests PASSED
0 of 165 tests FAILED
3 of 165 tests PENDING

- Does this PR create or have dependencies on Icepack or any other models?  Yes. 
https://github.com/CICE-Consortium/Icepack/pull/230 

- Is the documentation being updated with this PR? (Y/N)  No
If not, does the documentation need to be updated separately at a later time? (Y/N)  Yes

- Other Relevant Details:

This code is NOT READY to be merged.  It's a mess (but it's a bit-for bit mess).  It'll be a few days before I get back to this, so if anyone wants to jump in, go for it...   I hope it's now arranged in the proper order to dynamically allocate the tracer arrays using ntrcr.  

Here are the steps I suggest be done next, not necessarily in this order -- these are somewhat negotiable.

1. combine input_data2 and input_zbgc0 into a routine called “count_tracers”
2. keep n_trbgcz, n_aero, n_zaero, n_algae, n_doc etc in namelist, and remove the following n_[trcr] and set them in the code based on namelist logical flags:

        n_trbri = 0
        n_trzs = 0
        n_trbgcz = 0
        if (tr_brine) then n_trbri = 1
        if (tr_bgcs) then n_trzs = 1
        if (tr_bgcz) then n_trbgcz = 1
        if (.not. tr_bgc_C) then n_doc = 0
        etc

I did this for the age, FY, lvl and pnd tracers.

3. in the combined input_data2 and input_zbgc0:
a.  instead of this:

       nt_[trcr] = max_ntrcr
       if (tr_[trcr]_ then
           ntrcr = ntrcr+1
           nt_[trcr] = ntrcr
       endif

  do this:

      nt_[trcr] = 0  ! collect these for all [trcr]
      if (tr_[trcr]_ then
         ntrcr = ntrcr+1
         nt_[trcr] = ntrcr
     endif

     ! repeat above code blocks for all [trcr], adding up ntrcr
     ! then add 1 to ntrcr, to hold the unused tracer indices
     ntrcr = ntrcr + 1
     max_ntrcr = ntrcr 
     if (.not. tr_[trcr]) nt_[trcr] = max_ntrcr  ! collect these for all [trcr]

b. remove the old calculation of max_ntrcr (or use it as a sanity check)

4. use ntrcr throughout the code instead of max_ntrcr 
5.  allocate bgc arrays using n_[trcr] instead of max_[trcr]
6. change namelist flags for consistency (this will affect icepack and not be backward compatible, so maybe we shouldn't):
change skl_bgc to tr_bgcs
change z_tracers to tr_bgcz
7. fix hard-wired array sizes passed into subroutines (use (:) instead)
8. clean up routines input_* and init_zbgc (remove extraneous tracer_queries and tracer_inits; write out variables when they are first read/initialized; remove unused variables)
9. clean up CICE_InitMod.F90 and replicate changes in other drivers

FYI:  for max numbers for bgc tracers, see options/set_env.bgc in icepack


eclare108213:
Some clarifying details:
I divided the original input_data and init_zbgc subroutines into pieces and rearranged them in order to get the calculation of ntrcr all in one place (currently in input_data2 and input_zbgc0, which need to be combined).   Here is the order now in drivers/cice/CICE_InitMod.F90:

      call input_data           ! namelist variables 
      call input_zbgc           ! vertical biogeochemistry namelist
      call input_data2          ! count tracers
      call input_zbgc0          ! count additional bgc tracers
        ... initialize other things, including dynamical allocations of almost everything
      call init_zbgc            ! vertical biogeochemistry initialization 

I also started removing redundant info from the namelist, for the 'easy' tracers (age, etc).  For those, I kept the on/off flags (tr_iage, etc) and set the number of tracers associated with that flag (n_trage etc) in the code instead of namelist.  Note:  it would be better to be consistent with the names of these numbers and use n_iage, n_FY, n_lvl, n_pnd instead of n_trage etc, if possible.

eclare108213:
Yes, the names are confusing and this is why I suggested simplifying them.  It helps to remember where they came from -- the n_trage etc were originally the cpp defs, and they were there so they could be turned off to reduce the memory footprint of the big tracer array.  Many of them were just 1, i.e. on/off, and we also had a namelist flag that could then turn off the process or capability at runtime.  Now with dynamic allocation, the n_trage etc are redundant in namelist, but they are still useful in the code to count up the total number of tracers for the allocation.   So I started using the logical flags tr_[trcr] as the primary input, setting the tracer numbers n_[trcr] in the code (except for BGC, see below).  E.g. tr_brine should be in namelist, n_trbri should be moved out of namelist and preferably renamed to n_brine.  It would be assumed 0 unless tr_brine=T, in which case it's 1.  Some of the tracer flags turn on/off more than one tracer (e.g. lvl and some bgc).

> Also, just to be clear, the n_ are the number of tracers for that tracer group? nt_ is the start index of that tracer group in the tracer array? tr_ is a logical indicating whether that tracer group is active. Correct? It seems like n_ could just be hardwired for most tracers and never changed. tr_ would indicate whether it's active and nt_ would index it (if it's active) in the tracer array. Why do we need to zero out the n_ variables if the tracer is inactive. It that used as a flag later in the code instead of the tr_ flag?

Yes to all of that.  The n_ variables are zeroed out because of the (old) bit of code that adds them up all at once, which we can delete (item 3b above).  I think n_aero and other bgc numbers are used elsewhere in the code; my preference is to initialize all of the n_ variables to 0 and then set them to whatever they need to be based on the logical flags.  The logical flags should wrap ALL related tracer calculations -- the logicals are the primary, defining inputs.

The wrench in all of this is that we have the option of using an arbitrary (up to some max) number of bgc tracers.  I decided not to hardwire the bgc tracers to a particular number, mainly because of the aerosols in CESM and I'm not sure if @njeffery  varies the number that she uses, depending on the scenario.  For those, we'll still have the logical for consistency with the other tracers, and it will override the numbers being nonzero.  However if one of the bgc numbers is zero, then it makes sense to set the logical to false, to prevent loop weirdness.

I'd like to rename the flags and numbers to be more consistent -- need to make a list of them......

apcraig:
Thanks, that's helpful.

I'm actually writing up a list myself right now, just to better understand what's there.  Having the count_tracers subroutine with everything together helps a lot.  I'l share the list, should have it in a few minutes.  May have more questions then.

njeffery:
It is useful to be able to vary the bgc tracer numbers to suit the forcing or the ocean bgc model.

n

apcraig:
I just pushed a bunch of changes.  These should address 1, 2, 3, and 4 on the list.  Some of 8 is also done, abort checks on the bgc input have been added, and 9 variables were removed from CICE, 5 are namelist (max_ntrcr, max_nsw, n_bgc, nltrcr, n_trbgcz, n_trzs, n_trzaero, n_trbri, n_trbgcs).   I am working on 5 and 7 now, I think they are related.

apcraig:
I have verified bit-for-bit on conrad for all 4 compilers.  The only failure was the alt01 which requires an update to the namelist setting (not yet done) to be compatible with the updates to the input checks.

apcraig:
I just pushed a few more changes, including updates to alt01.  @eclare108213, have a look when you have a chance and lets figure out what still needs to be done.

eclare108213:
This looks good, @apcraig, thanks!  I'll pull the changes and test them here at LANL.  These are some notes just from looking through the code changes, which either of us might address:

Once we start changing Icepack, let's fix this to properly use the logical flag (unless nltrcr is a multiplier), from ice_step_mod.F90:
      ! tcraig, nltrcr used to be the number of zbgc tracers, but it's used as a zbgc flag in icepack
      if (z_tracers) then
         nltrcr = 1
      else
         nltrcr = 0
      endif

Do we have a test with all pond parameterizations off and using the ccsm3 shortwave?  Or do we need to keep this in alt01 and put the formdrag test in a different option set?

For abort_flag=16 in ice_init.F90, the error message should be
> ERROR: formdrag=T and tr_pond=F

I'm a little concerned that we might get out-of-bounds errors when aerosols are off.  For now, we are specifying a single array index for tracers that are not in use, but the aero tracers needs many indices because they are always treated in loops like n=nt_aero, nt_aero+4*n_aero.  That's why we initialized nt_aero as max_aero-4*n_aero in the original "bad kludge".   If we have a testing option that has aerosols turned off and debugging on, it should catch the problem, if it exists.

Remove the format statements from the bottom of input_data?

apcraig:
I will run some tests, especially debug with n_aero set to 0.  I'll also try to address some of the comments with code changes.  I will test on conrad.

apcraig:
running debug tests on conrad for all alt and box configs.  getting some aborts.  will try to fix.

format statements are needed in input_data from what I can see.

I'll look at the alt tests and try to move formdrag to a different alt test.

eclare108213:
I just read through ice_init.F90 and ice_init_column.F90, where the biggest changes are, and fixed a few things, nothing answer-changing.  My base_suite built and the runs are still sitting in the queue.

From the original list above, it looks like 6 is not started, 5 and 7 are partially complete, and everything else is done along with some additional enhancements that weren't in the list.   5, 6, 7 could become a 'cleanup' task in a new PR.  7 is captured in issue #238.

apcraig:
I just pushed a new update.  I added several new debug tests and updated the alt01 (now with cesm shortwave and no ponds) and alt04 (now with lvl ponds and formdrag) tests.  Things are working pretty well.  A couple of the new debug tests with the box configuration are failing, but all the alt tests with debug on are passing.  The nt_aero and n_aero settings do not seem to be causing any problems in debug mode.  Remember that when tr_aero is off, nt_aero is pointing at the unused tracer and n_aero=0.  So it's likely the array sections in arguments are valid, and as long as they're not being used, should be OK.  That seems to be true based on testing.  See current results here, hash 770bd8f231.

I am not particularly worried about the box failure.  We should try to fix it, but I would not hold up the release for it either.

I also agree that 6 on the list was never done, 5 and 7 were started but need quite a bit more thought and refactoring.

@eclare108213 before or after the merge, could you review the model output for settings/namelist. I think we should make sure we are writing out the right stuff and that it's happening in the correct place during the init sequence.  I also think if there are any other "abort checks" that could be added, that we should try to do that.

I think it would be good to merge this before the weekend if we can, and we feel it's ready.  We can continue to additional changes on a new branch as needed.  I think this refactor has been very good to do!

eclare108213:
Question for @njeffery:

We are checking the numbers of tracers that are allocated - are these checks correct, or e.g. should we check that n_fep+n_fed <= icepack_max_fe and similarly for doc, dic?

      if (n_algae > icepack_max_algae) then
            write(nu_diag,*) subname//'ERROR: number of algal types exceeds icepack_max_algae'
         endif
         abort_flag = 112
      endif
      if (n_doc > icepack_max_doc) then
            write(nu_diag,*) subname//'ERROR: number of doc types exceeds icepack_max_doc'
         endif
         abort_flag = 113
      endif
      if (n_dic > icepack_max_doc) then
            write(nu_diag,*) subname//'ERROR: number of dic types exceeds icepack_max_dic'
         endif
         abort_flag = 114
      endif
      if (n_don > icepack_max_don) then
            write(nu_diag,*) subname//'ERROR: number of don types exceeds icepack_max_don'
         endif
         abort_flag = 115
      endif
      if (n_fed  > icepack_max_fe ) then
            write(nu_diag,*) subname//'ERROR: number of dissolved fe types exceeds icepack_max_fe '
         endif
         abort_flag = 116
      endif
      if (n_fep  > icepack_max_fe ) then
            write(nu_diag,*) subname//'ERROR: number of particulate fe types exceeds icepack_max_fe '
         endif
         abort_flag = 117
      endif

eclare108213:
@apcraig I suggest to go ahead and merge, if you are happy with this so far, to get a new baseline since alt01 changed (right?  That’s the only fail in my baseline tests yesterday).  Then we can continue cleaning up in a new PR.

Issue #221, opened 2018-10-30
results.csh output


eclare108213:
The output from the results.csh script is confusing when it's executed while a test suite is running -- it returns FAIL for jobs that are still in process.  Can these be changed to PEND?

apcraig:
I had that implementation until recently, but then thought that we need more clear pass/fail info.  If the default is pend, then any job that fails to submit will be left pending, not failing.  I think it's better to have the default be fail and then have a running job go to pend and and a completed job go back to pass or fail.  I think we just have to understand that when using results.csh while testing.

eclare108213:
Okay.  Can we have the script print a message to that effect, if there are still jobs pending that are marked FAIL?

apcraig:
When a test is created, the default result is FAIL.  As soon as the test is built, the status changes to PASS for build and PEND for run.  Then after that, the run and test results will depend on the results of the run.  It is still hard to know whether a PEND run is still pending or whether it was never submitted.  Users have to look at that.  These results need to interact with the reporting as well.  If someone can come up with a clear set of requirements, that address what is needed for both the results.csh script and report_results script, I can implement them.

apcraig:
Good idea.  thanks.

apcraig:
Fixed in #239

Issue #216, opened 2018-10-20
mpirun vs serial ./cice


eclare108213:
When creating a serial case using -p 1x1, 'mpirun' is put into the cice.run script.  This works, but is it necessary?  Are there ramifications, other than that the job scheduler at LANL insists on there being a proper allocation for the job (i.e. I can't quickly run it on the front-end machine)?  I can live with it, I just want to be sure it's intended and has no unintended side effects.

apcraig:
It's a good point.  We should be able to fix this and get rid of the mpirun for serial builds.  I will look into it.  I don't think there are any serious ramifications.  For machines that have mpi, launching a serial application with mpi is fine.  For machines that don't have mpi, they would not implement an mpirun in their launch.  Having said that, it's probably worth exploring whether we can selectively launch with or without mpi on the fly.

apcraig:
Capability added and tested for conrad and gordon in #239.  Can be implemented by default on a machine by machine basis.  This will provide testing of the capability on an ongoing basis, at least on gordon and conrad.  Also good to test that serial jobs can launch with mpi, still doing that on other test platforms.

PR #239, opened 2018-11-17
fix iobinary failure introduced in recent PR, address non-mpi launch …

apcraig:
fix iobinary failure introduced in recent PR, address non-mpi launch or serial cases, add test results descriptions

- Developer(s): tcraig

  Tested iobinary cases on conrad and passing.
  Tested travis suite on all 4 compilers on conrad and all passing and bit-for-bit, validates serial/mpi launch differences.

- Other Relevant Details:

addresses iobinary failure in recent PR for hycom, needed #ifdef ncdf around new netcdf code
tests and validates serial vs mpi launch #216 , implemented for conrad and gordon by default
adds desciptions for test results output #221.

### DIFF
--- a/cice.setup
+++ b/cice.setup
@@ -862,6 +862,12 @@ echo "#fail = \$failures" >> results.log
 echo "#pend = \$pends" >> results.log
 
 echo ""
+echo "Descriptors:"
+echo " PASS - successful completion"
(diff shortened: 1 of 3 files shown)


apcraig:
We should try to merge this by Sat midday for weekend testing.  This fixes the iobinary problem currently on the master.

PR #253, opened 2018-11-22
Updates for RASM compatibility

apcraig:
Update CICE to support RASM integration.

- Developer(s): tcraig

- Does this PR create or have dependencies on Icepack or any other models?  RASM needs updated Icepack as well, does not affect CICE standalone though.

- Other Relevant Details:

Synchronize CICE6 in RASM and in Consortium for pre release testing.  Mostly impacts either cesm driver code or code inside CESMCOUPLED cpp.  A couple minor changes were made to the cice driver code for consistency.  The cice and cesm driver codes are now well aligned.

Tested full suites on conrad with 4 compilers against current trunk and everything is bit-for-bit.  See 925e299 https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks.

### DIFF
--- a/cicecore/cicedynB/general/ice_init.F90
+++ b/cicecore/cicedynB/general/ice_init.F90
@@ -132,6 +132,7 @@
       integer (kind=int_kind) :: rpcesm, rplvl, rptopo 
       real (kind=dbl_kind) :: Cf, puny
       integer :: abort_flag
+      character (len=64) :: tmpstr
 
       character(len=*), parameter :: subname='(input_data)'
 
@@ -472,8 +473,8 @@
          ! each task gets unique ice log filename when if test is true, for debugging
          if (1 == 0) then
             call get_fileUnit(nu_diag)
-            write(str,'(a,i4.4)') "ice.log.task_",my_task
-            open(nu_diag,file=str)
+            write(tmpstr,'(a,i4.4)') "ice.log.task_",my_task
+            open(nu_diag,file=tmpstr)
          endif
       end if
       if (trim(ice_ic) /= 'default' .and. trim(ice_ic) /= 'none') then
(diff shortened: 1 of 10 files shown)


dabail10:
I will run this through the NAG compiler on hobart.

apcraig:
@dabail10, It's possible the istep stuff was changed, are you seeing something in particular because I don't see anything in the diffs?  This implementation seems to work in rasm which is based on cesm1.x.  cesm2 might have some differences, not sure.  We might actually have to eventually implement a cesm1 and cesm2 driver.  maybe even a cesm2_nuopc as well.  But we'll take that when it comes.

dabail10:
In the model CICE_InitMod.F90 there is diffs in the declarations. These do not cause problems. Just unused variable messages. I will run it through NAG.

apcraig:
Good point.  I did a few things comparing old and new versions of the code.  Basically, I took the cice driver and modified it as needed for RASM and dropped it into the cesm driver directory.  I think it's handy for the two files to be as much alike as possible and only have needed differences.  It could be that I got something wrong or that there are some things from the cice driver that are not needed in the cesm driver.  I guess my preference is to keep them the same as much as possible because I think it will make them easier to keep synchronized.

dabail10:
It occurred to me that I can't really run the RASM driver through NAG. I'll eventually have to work on the CESM driver as well. So, this is probably fine as is.

apcraig:
Yes, this includes an update to the icepack code.  I pushed a one line change over the weekend to the icepack master.

eclare108213:
@dabail10 are you satisfied with this PR or do we still need to clean up some variables?

dabail10:
I'm fine with this. I can't run the CESM driver changes through the NAG compiler yet.

PR #255, opened 2018-11-27
eap update_stress_rdg updates

apcraig:
Merge in hheorton eap_efficient and eap_efficient_interpolate branches, review and test.

- Developer(s): hheorton, apcraig

- Are the code changes bit for bit, different at roundoff level, or more substantial? more substantial?

If not, does the documentation need to be updated separately at a later time? (Y/N) once we test the interpolate mode a new namelist will be added.

- Other Relevant Details:

-Merged in hheorton eap_efficient and eap_efficient_interpolate branches.
-Added interpolate_stress_rdg variable to turn on interpolation feature (untested).  Default is for this feature to be turned off.  This needs to be tested and turned into a namelist.
-Add a couple lines that were missing relative to new_efficient.txt file that was provided earlier.  The computation of dtemp1 and dtemp2 seems to have been accidently left out of the eap_efficient branch.
-Added check on atan2 if second argument is zero.  We have had problems with this on some compilers and it is implemented elsewhere in the code.

Tested on conrad with kdyn=2 tests and they are now passing.  Answers have changed.  More extensive testing to be carried out concurrent with further review.

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_eap.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_eap.F90
@@ -21,7 +21,7 @@
       use ice_blocks, only: nx_block, ny_block
       use ice_communicate, only: my_task, master_task
       use ice_domain_size, only: max_blocks, ncat
-      use ice_constants, only: c0, c1, c2, c3, c12, p1, p2, p5, &
+      use ice_constants, only: c0, c1, c2, c3, c4, c12, p1, p2, p5, &
           p001, p027, p055, p111, p166, p222, p25, p333
       use ice_fileunits, only: nu_diag, nu_dump_eap, nu_restart_eap
       use ice_exit, only: abort_ice
@@ -1601,19 +1601,19 @@
       ! local variables
 
       integer (kind=int_kind) :: &
-         kx ,ky, ka
+         kx ,ky, ka, kxw, kyw, kaw
 
       real (kind=dbl_kind) :: &
          stemp11r, stemp12r, stemp22r,   &
          stemp11s, stemp12s, stemp22s, &
-	 a22, Qd11Qd11, Qd11Qd12, Qd12Qd12, &
+         a22, Qd11Qd11, Qd11Qd12, Qd12Qd12, &
          Q11Q11, Q11Q12, Q12Q12, &
          dtemp11, dtemp12, dtemp22, &
          rotstemp11r, rotstemp12r, rotstemp22r,   &
          rotstemp11s, rotstemp12s, rotstemp22s, &
          sig11, sig12, sig22, &
          sgprm11, sgprm12, sgprm22, &
-	 invstressconviso, &
+         invstressconviso, &
          Angle_denom_gamma,  Angle_denom_alpha, &
          Tany_1, Tany_2, &
          gamma, alpha, x, y, dx, dy, da, &
@@ -1623,6 +1623,11 @@
 
       real (kind=dbl_kind), parameter :: &
          kfriction = 0.45_dbl_kind
+
+      ! tcraig, temporary, should be moved to namelist
+      ! turns on interpolation in stress_rdg
+      logical(kind=log_kind), parameter :: &
+         interpolate_stress_rdg = .false.
 
       character(len=*), parameter :: subname = '(update_stress_rdg)'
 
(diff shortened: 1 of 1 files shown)


apcraig:
@hheorton I have several questions for you as part of code review.

- First, please have a review of the implementation and let me know if you see any problems.
- I note angle_denom variables are defined different than new_efficient_txt.  It looks OK to me, just want to check with you that it's what you prefer.
- There are various checks on x and y for range, 0 < y < pi and pi/4 < x < 9pi/4.  Do we have adequate checks and corrections on that?  My preference in these cases is to always loop adding or subtracting 2pi "until" we are in the right range.  In this implementation, we do one pass (or less).  Are we sure in all cases, that is going to be adequate?
- In the interpolated lookup, kx, ky, ka are computed then kx, kx+1, ky, ky+1, ka, ka+1 are passed into functions.  Are we sure that we are not going out of bounds on either the indices or the indices+1 in these cases.

apcraig:
@eclare108213 
I am proposing we defer testing the interpolate feature until after release and leave it undocumented.  After release, we can test it, add a namelist, and document.

The answers are not bit-for-bit as expected.  The technical tests are passing (exact restart, debug, etc).  What validation do we need to do on the implementation?  I assume we should run a qc test?  What configuration?  We need one with kdyn=2.  Should I do a qc with alt03?  Or should I run out of the box just setting kdyn=2?  Or some other configuration?  I am willing to carry this out, just want to make sure I understand what configuration we should be testing.

apcraig:
One final note, do not merge this yet!  More testing is required.  I have assigned myself to the merge just to remind others.  I created this PR so we could review, discuss, and test concurrently.

apcraig:
Just finished testing all eap tests on conrad with 4 compilers.  All tests are passing including tests that were failing before.  All tests are also not bit-for-bit, as expected.  Will run full test suite eventually, just to make sure non-eap tests are still running and bit-for-bit.

eclare108213:
This is great, thanks @hheorton and @apcraig.  Since it's not BFB, we should run the QC tests for kdyn=2.  I'd recommend doing this in the standard gx3 setup, at least.  gx1 would be a better test than gx3 for this, I suspect, although that might take a while (5 year simulation).  Isn't there a standard QC gx3 configuration/script, and you'd just need to change kdyn=2?

apcraig:
OK, I'll do the qc test with gx3, out of the box + kdyn=2.

eclare108213:
I agree with this plan:

> defer testing the interpolate feature until after release and leave it undocumented. After release, we can test it, add a namelist, and document.

hheorton:
Thank you for proceeding with the EAP tests. When you've tested the interpolate feature, it will be great to see the computational cost of its inclusion. The uk met office are currently working on using EAP and it will be great to let them know if they should also use the interpolated look up table. They are however using CICE 5 still, so not a direct comparison.

I will be very interested to see the results of your tests. We saw a 30% reduction in total run time with the new efficient stress_rdg, although out comparisons were not particularly thorough.

Get Outlook for Android<https://aka.ms/ghei36>

apcraig:
A full test suite run on conrad with 4 compilers shows bit-for-bit for all tests except tests with kdyn=2.  All tests now pass as expected.  See hash 319ad682 at https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks.  Based on two short tests in non-debug mode with multiple compilers on conrad with kdyn=2, the new code looks like it's 10%-15% faster with one case closer to 5% and another closer to 20%.  These numbers have a lot of variability in them, so I don't think we can be particularly precise without more performance analysis and multiple trials, but I think 10%-15% is probably real given the consistency of the improvement across different compilers and for the 8 test results we have.

apcraig:
qc test on gx3 passed.  This was for default namelist plus kdyn=2.  I am running a gx1 test now too.

conrad> ./cice.t-test.py $WORKDIR/CICE_RUNS/conrad_intel_smoke_gx3_32x1_medium_qc.qc_base $WORKDIR/CICE_RUNS/conrad_intel_smoke_gx3_32x1_medium_qc.qc_eaphh1
INFO:__main__:  /p/work1/apcraig/CICE_RUNS/conrad_intel_smoke_gx3_32x1_medium_qc.qc_base
INFO:__main__:  /p/work1/apcraig/CICE_RUNS/conrad_intel_smoke_gx3_32x1_medium_qc.qc_eaphh1
INFO:__main__:

Issue #244, opened 2018-11-19
Remove "v5" references in doc


duvivier:
@JFLemieux73 found a few places in the documentation where we reference Version 5 or V5. We need to search through and find anywhere else. Known issues in:

https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_implementation.html#restart-files

https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_troubleshooting.html#restarttrouble

eclare108213:
We are still using the v5 restart files, so some references might be okay.

duvivier:
Yes, but we need to go through carefully and be sure. And in some places, we can probably modify the wording to say "for CICE version 5 and later" or something like that to be inclusive of this release.

eclare108213:
Icepack needs a ‘citing the code’ section as in the CICE documentation (1.5)

2.2 instead of ‘model components’ maybe make sections for momentum, stress and horizontal transport

section 4 remove ‘implementation’ from the section titles

I didn’t read sections 3 very carefully, since it looks like they’ve been kept up to date with changes in the scripts, etc., and I didn't get very far into section 4 tonight.

The font convention is very inconsistent in these sections, especially \bf

comp_ice still appears in a number of places and needs to be changed to whatever is appropriate, probably either cice.settings or ice_in?

Also, we should capitalize Icepack whenever we refer to the software package or its modules, but in the code, directory and filenames with 'icepack' are not capitalized. 

section 3.1 second paragraph - I think the link should be to Icepack, not CICE

section 3.1.1 - change icepack to cice

section 3.1.2, 3rd paragraph:  
remove “, including that used for the column configuration”
change “These are binary unformatted, direct access files produced on an SGI (Big Endian). If you are using an incompatible (Little Endian) architecture, choose rectangular instead of displaced_pole in ice_in, or follow procedures as for conejo (⟨OS⟩.⟨SITE⟩.⟨machine⟩ = Linux.LANL.conejo). There are versions of the gx3 grid files available.” to
“These are binary unformatted, direct access, Big Endian files.”

Replace the Bio-grid and Column configurations sections with:

Vertical grids

[...]

PR #256, opened 2018-11-28
doc mods for CICE release

duvivier:
Finalizing documentation for CICE (and Icepack) releases. This will be done in stages, but I want to start a PR so it's easier to compare the rst changes.

- Developer(s): Alice DuVivier w/ Contribution from @eclare108213 

- Does this PR create or have dependencies on Icepack or any other models? Yes (Icepack documentation)

Here is the readthedocs test of the documentation changes:
https://cice.readthedocs.io/en/latest/index.html


duvivier:
I have updated the second set of mods based on conversations with @eclare108213 going on under issue #244 

See new documentation:
https://cice.readthedocs.io/en/latest/index.html
https://media.readthedocs.org/pdf/cice/latest/cice.pdf

duvivier:
@apcraig I have two particular questions:

In ug_running.rst, Elizabeth said this didn't make sense to her:
>The version should be some like n.m.p.string.
I'm not sure how else we'd want to re-word it. Any thoughts?

In ug_troubleshooting.rst, is the discussion of BARRIERS still correct?
>In some configurations, multiple calls to scatter or gather global variables may overfill MPI’s buffers, causing the code to slow down (particularly when writing large output files such as restarts). To remedy this problem, set BARRIERS yes in cice.settings. This synchronizes MPI messages, keeping the buffers in check.

duvivier:
@eclare108213  I propose we go ahead and merge this as it is. I think we still need to do a little bit more in the user's guide section (e.g. sections 3.2, 3.3 and 3.3.5 in particular) and will likely have a few more things as well. But there are a lot of changes at this point and I think it makes sense to merge them and then we can continue editing from there.

eclare108213:
I agree.  We can remove BARRIERS and CAM_ICE from the documentation for this release and clean up the code later.

eclare108213:
Please change Los Alamos National Security to Triad National Security in the documentation footers for both CICE and Icepack.

PR #266, opened 2018-11-30
Final documentation clean up

duvivier:
Final documentation modifications

- Developer(s): Alice and Elizabeth

- Other Relevant Details:

- Removing an extra reference to basalstress in the namelist table (it was repeated 2x).


eclare108213:
I didn't test my changes - apologies if I broke something!  I started changing fonts for variable names in ug_implementation.rst (replacing ` with ``), then decided I'd break something for sure, doing that.

duvivier:
I'll pull everything locally before we merge and test it with read the docs.

apcraig:
I just pushed some changes after testing the documentation.  The new table was not working and it seems to be now.  The other thing I did is update icepack to the latest version on master (with updated docs).  You can see the CICE documentation here for this branch here.

https://apcraig-cice.readthedocs.io/en/duvdoc/index.html

Just FYI with regard to process, what I did is create a branch (duvdoc) in my fork, pulled duvier:doc to it, pushed those changes to my branch, then setup readthedocs to build that branch.  I iterated on the duvdoc branch until I produced the table.  Then I pushed it back to alice's doc branch directly.  It took a few tries to get the table out. 

I'm happy for someone else to run another "doc test" of this version before we pull.

apcraig:
And the new table is here,

https://apcraig-cice.readthedocs.io/en/duvdoc/science_guide/sg_tracers.html

in my test readthedocs.

eclare108213:
Thanks!  I'm slowly coming up to speed on the documentation...  I'll look at the table now.  Do the entries make sense to you?

eclare108213:
I prefer the column order as it is, because it is more or less in the order that things happen in the code.  Good enough, IMHO!
I still intend to check the how-to-add-a-tracer section.

eclare108213:
I've just added what I hope is my last major edit to the docs before the release.  I'm still seeing typos, but I'm going to stop editing for now -- we will get these out eventually.  

One section could be improved though, for both Icepack and CICE, and I'm not enough of an expert to write it:  "Building documentation".  This appears to be written with the user's master branch in mind, which is fine, but it would be helpful to also include instructions for (1) how to find documentation for another branch in my own fork and (2) how to edit and check the documentation in another user's fork.  I think these go something like this:
(1)  Click on Versions, find your branch in the "Inactive" list, click 'Edit' then check the 'Active' box and save. Then click on 'Builds' and the branch name.
(2) (following suggestion by @apcraig above)   create a branch in my fork, pull the other user's branch to it, push those changes to my fork/branch, then set up readthedocs to build my branch. Edit, add, commit. Then push my branch back to the other user's doc branch directly.   I'm not clear on what that last command would look like.  Should this be in the git workflow page instead?

apcraig:
For the readthedocs help, maybe we should have a document outside the CICE document, sort of like the git workflow doc.  I guess it could also be in the git workflow doc, but it makes more sense to me to have a git workflow doc, a readthedocs help page, and the CICE guide which might point to both either explicitly or implicitly via the consortium resource links.   If we can keep documentation of external process out of the CICE doc, that's not a bad idea.  

We certainly can/should document how to use readthedocs off branches.  And the git pull/push between forks/branches also needs to be documented in the github workflow.  I'm figuring out those git tricks slowly.

I do think having the wiki for documenting a bunch of the process is really helpful as it's dynamic and those features don't necessarily track the cice documentation well.  Especially as we improve our understanding of git and readthedocs and other stuff, it'll be good to keep updating that documentation and not necessarily have it committed in the cice repo directly.

I agree that a lot of this can be done after the release.

eclare108213:
I like the idea of having a readthedocs workflow wiki like our git workflow, especially since it seems like a lot of the material is common to both CICE and Icepack.  If this shift can be done quickly, I think we should do it, otherwise it's fine to go ahead and release with the docs as they are.
@duvivier, please merge this branch when you are ready.

duvivier:
@apcraig @eclare108213 , we are almost there. 

1.  I just pulled the "documentation section" (in the developer's guide) out of the documentation and made a wiki page: https://github.com/CICE-Consortium/About-Us/wiki/Documentation-Workflow-Guidance. I think we can modify it easily from here, and I do think it needs a bit of finessing. But I don't think we need to wait on the release for that. 

2. @phil-blain and I noticed a few of the dynamics namelist values were missing from the table. I added these and he's going to add a bit of text. I think as soon as this is done we are ready to release.

duvivier:
Here are the tested versions of the html and pdf. 
HTML: https://cice.readthedocs.io/en/latest/
PDF: https://media.readthedocs.org/pdf/cice/latest/cice.pdf

I think the last thing is to wait for @phil-blain too add his text. If you want to go ahead and release as is, I think everything else is ready to go. We can make typo fixes, etc. later on. We can even add the dynamics info later if we want to go ahead and release now.

duvivier:
@phil-blain Thanks! :)

duvivier:
@apcraig 
I've updated the Documentation workflow guide (https://github.com/CICE-Consortium/About-Us/wiki/Documentation-Workflow-Guidance) and Resource Index (https://github.com/CICE-Consortium/About-Us/wiki/Resource-Index) on the wiki. Please feel free to edit these further if you see issues. I think some general improvement about the processes @eclare108213 describes above would be helpful, but I am not working on that at the moment.

Issue #275, opened 2019-01-25
Place test and test_suite in specific filesystem location


phil-blain:
When using `cice.setup` to create a test or a test_suite, the scripts creates folders in the current CICE directory ; there does not seem to be an option to put the folder and files for the test/test_suite in a different location, as is possible for cases. It would be great to have that possibility, because right now running quick tests on a development machine clutters up the git repository....
Discussion board link : [https://bb.cgd.ucar.edu/put-test-and-testsuites-specific-directory](https://bb.cgd.ucar.edu/put-test-and-testsuites-specific-directory)

apcraig:
Fixed in PR #281 also in Icepack PR 253

PR #281, opened 2019-02-14
Add ability for user to set test directory

apcraig:
Add ability for user to set test directory, addresses #275 

- Developer(s): apcraig

- Does this PR create or have dependencies on Icepack or any other models?  no

- Other Relevant Details:

Tested on conrad, bit-for-bit, also tested error modes.

Tested that documentation builds, https://apcraig-cice.readthedocs.io/en/se1/


phil-blain:
I tried the new tdir option and it seems to work only for test suites and not for single tests, is that normal ?

apcraig:
Thanks @phil-blain, let me look into this.  A quick code review suggests this was overlooked.  I'll fix it if so.

apcraig:
The tdir bug for single tests is included in PR #300

PR #290, opened 2019-03-08
update cheyenne modules associated with machine upgrade

apcraig:
Update cheyenne, requires with machine upgrade early March, 2019

- Developer(s): tcraig

- Are the code changes bit for bit, different at roundoff level, or more substantial?  bit-for-bit, see 38cad in https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks

- Does this PR create or have dependencies on Icepack or any other models?  no

- Other Relevant Details:

Full test suite is bit-for-bit with last weekend test.  Old modules no longer supported with machine upgrade in early March, 2019.  These changes should be usable in older version of CICE as well.


apcraig:
I will merge the PR late this afternoon even if it hasn't gone thru review.  I have tested these changes, they only affect cheyenne, and cheyenne is currently not working on the master.

Issue #299, opened 2019-03-15
poll_queue script / general csh usage


phil-blain:
I tried using the **`poll_queue.csh`** script on my machine (ubuntu 14.04) and it fails ; it just outputs
```end: No match.```
 However, when I run it under tcsh instead of plain csh (by changing the shebang), it works.

On my system I have the following links:
 /bin/csh -> /etc/alternatives/csh -> /bin/bsd-csh

So it seems the csh version that comes with ubuntu is not capable of running this script. I think this should be documented, and maybe we should change the shebang for this script (and maybe others? from what I read csh has a reputation of being buggy...)

phil-blain:
The line 
```set res="$?"```
 in the **`cice.test`** script is also causing a problem with my csh:
``` Illegal variable name.  ```
It works fine with tcsh.

apcraig:
Portability of these scripts is very important.  If you switch the poll_queue script to tcsh, everything works fine?  Any idea what's causing problems in that script for you?

Also, does $status work better than $? for you in your csh environment?

phil-blain:
Also, on a different note:
The poll_queue script is only copied to the suite directory if the suite is created with the --report option (line 372 in cice.setup). I think this script is useful even if the user is not reporting the test results back to the wiki, so I think it should always be copied.

apcraig:
This is being fixed in #300.  I have replaced $? with $status and have refactored the poll_queue so it does not parse the suite.log file.  I am also copying the poll_queue script into the test suite directory for all suites.

apcraig:
PR #300 complete

Issue #204, opened 2018-10-10
bit-for-bit diagnostics


apcraig:
CICE currently has 3 different flags that affect bit-for-bit diagnostics.  The first "DITTO" is an env variable in cice.settings.  That triggers the cpp -DREPRODUCIBLE but only for a few machine/compiler implementations.  It probably should be specified in all machine/compiler implementations if any.  Finally, there is the bfbflag in namelist.  That turns on yet another calculation for the global diagnostics but only on the mpi side.

I see several issues
- multiple independent approaches to address the same issue
- inconsistency in scripts application to various machines and compilers
- inconsistency in implementation between mpi and serial versions (serial does not have a bfbflag implementation)
- lack of reuse in implementation of sums, mpi vs serial, int/float/double, prod sums, etc. 
- no recent testing as far as I know
- really does need to be cleaned up

I would like to refactor this with the following requirements
- run time flag only
- implementation in both mpi and serial, recognizing this is not only a pe count issue but also a decomp/block issue.
- code reuse where possible
- leverage new techniques such as the Worley ddpdd approach to reduce cost
- potentially support multiple approaches via a consistent, well documented namelist input
- test and validate

This feature is useful, especially for quick bit-for-bit comparison of log files during testing.  If the cost is low enough, it may even be a viable default setting.

It is probably not a lot of work, I have implemented the same thing in other projects and would leverage some of that work.  I would like to propose it be added to the CICE6 release project.

Pinging @eclare108213, @mattdturner, @dabail10 for any comments or feedback.

eclare108213:
This makes sense to me.  DITTO and REPRODUCIBLE came from UKMO a long time before that, specifically because they wanted to be able to compare diagnostic output across configurations (these only affect the diagnostic output).   bfbflag came from CESM 4 years ago (the "blame" functionality in github is very handy).    @apcraig, yes, please clean this up.

duvivier:
Also get rid of the following in cice.settings:
setenv CAM_ICE no # set to yes for CAM runs (single column)
setenv BARRIERS no # prevent MPI buffer overflow during gather/scatter

apcraig:
PR #300 complete

PR #300, opened 2019-03-18
Refactor bit-for-bit diagnostics

apcraig:
Refactor bit-for-bit diagnostics, #204 

- Developer(s):  tcraig

- Are the code changes bit for bit, different at roundoff level, or more substantial? bit-for-bit, #77addedd, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks.
All results are bit-for-bit but due to some new additions and some refactoring of the tests, some tests show missing regression results.

- Does this PR create or have dependencies on Icepack or any other models?  Yes, requires latest version of Icepack included in this PR.

- Other Relevant Details:

Removed all references to cpps DITTO, REPRODUCIBLE, and the old bfbflag code.  Added new algorithms to carry out the global sum diagnostics leveraging various implementations.  There is NO global gather to generate a bit-for-bit sum.  bfbflag is now a character string and the supported options are off, lsum4, lsum8, lsum16, reprosum, and ddpdd.  The algorithms are described briefly in the documentation.  off==lsum8 and this option does the global sum via a local reduction then a global scalar MPI allreduce.  reprosum is the recommended setting for bit-for-bit diagnostics.

This has been implemented primarily in the ice_global_reductions.F90 code.  The same algorithms are implemented in the mpi and serial versions and the mpi and serial global reductions are bit-for-bit with each other with bfbflag=reprosum.  The implementations work for different pe counts, different blocks sizes, mpi, threading on and off, and serial.  Prior implementations only worked on a subset of these conditions.

Some new tests and scripts were added to test the log files.  The logbfb test is identical to the smoke test except for comparisons to other cases, the log file is compared instead of the restart.  reprosum_suite is a suite of tests that test the log file diagnostics with bfbflag='reprosum'.

The comparebfb.csh script was refactored to simplify the implementation and the log comparison was separated to a different script.  In addition, there were some errors introduced in October 2018 that broke some of the comparebfb testing and this has been fixed.

A new feature to allow sleeps between test case build/submission was added to try to manage dependencies between cases better.   This is not a perfect solution.  The problem is that when one test is comparing it's results to another, the second has to complete after the first is done, but this does not always happen depending on how jobs are flowing thru the queue.

Obsolete cpps were removed and a bug that broke the cpps implementation in the build was fixed.

The decomp test was refactored to be more compatible with the updated comparebfb.csh script.  It also now should work with multiple binary files.  As a result, the directory structure of the results is slightly different and regression testing with prior versions results in missing data.  This will self correct as we test in the future.

Fixed a bug in a few of the diagnostic output fields.  This bug created non-reproducible results when running threaded tests.  There was an error in zeroing out of the al*_init fields in threaded regions in ice_step_mod.F90.

Added 8 byte integer kinds, needed for the reprosum and ddpdd algorithms.  Also use real*16 sums and reduction in the "lsum16" option.  PGI does not support this type, so had to add a NO_R16 cpp for pgi builds.

Update travis to support build for c code.

Refactor Macros file to better support serial and parallel builds as well as building c code.  Also needed a small update to the Makefile which had "cc" hardwired as the c compiler command.

Refactor poll_queue script and fix --bgen default bug with single tests, #299, #301.

Update job launching to avoid mpi launches with serial code.  Also, work towards using serial (vs mpi) builds when possible.


duvivier:
Doc looks okay, not sure why Travis CI failed though.

apcraig:
I just relaunched the travis test, looks like it was failing on the input data download.  Will monitor it.

duvivier:
Ok. Let me know if there's something wrong with the FTP again. I haven't touched it lately and I don't think Dave has, so hopefully there's not a system issue.

apcraig:
ftp worked this time, but I'm running into another problem that I have to fix.  I added a c file and travis doesn't yet know how to compile c.  I need to update the travis env.  Hope to fix this today.

apcraig:
I am going to refactor the Macros files a bit to better support both serial and parallel compiles as well as compiling c code.  More changes coming.  Will try to test on most platforms as I go.

eclare108213:
Thanks @apcraig.
When you do that, could you provide the option for serial runs to be launched without using mpirun, please?  thx

apcraig:
@eclare108213, good idea.  We have that for a few machines already, will try to extend that.  I may need you to do some testing on your local machines.  I'll email you when I need some help with that.

apcraig:
@phil-blain, great to hear things seem to be working for you.  I can add missing ICE_MACHINE_QSTAT settings in the various machine files.  If we need to update them later, we can.

apcraig:
@eclare108213 , are you OK with merging this PR?

apcraig:
@eclare108213 , just another ping.  I think this PR is ready, but it's big enough that I prefer to have your OK before merging.  I will start a round of testing once this is merged, before the automated weekend testing.

eclare108213:
I have started another set of tests on badger.  In the last set, 16 of them failed, but they all seem to be due to machine flakiness, e.g.

> badger_intel_decomp_gx3_4x2x25x29x5.bfb/logs/cice.runlog.190320-155405:ORTE has lost communication with a remote daemon.

I suspect this is not a code-generated problem.  Will update when test results come in.

apcraig:
@eclare108213 , these are two good questions.

Regarding the SERIAL_REMOVE_MPI, one of the problems with having separate serial and mpi directories is that things get out of sync.  Stuff gets added to one, but not the other, and so forth.  In addition, for these global sums, it's important (if we want global sums to be reproducible in serial and mpi mode, which they are not with bfbflag='reprosum') that the implementations in serial and mpi mirror each other.  So I decided to basically take the mpi version that I implemented first and add the ability to turn off the MPI parts and then drop it into the serial directory with a directive turned on that does just that.  The two files are basically identical now which makes it easy to maintain moving forward.  It would also be possible to merge the serial and mpi directories if we wanted to and to use the file we have now for mpi and serial mode.  So anyway, there were a few reasons why it is not bad that the mpi and serial versions of ice_global_reductions.F90 are similar.  A bigger question is whether we want to continue to maintain the serial and mpi directories or whether it's better to have one with MPI cpp'ed out.  I noticed a lot of differences in the global_reductions code in the two directories before I refactored them.  I wonder if there are other differences (like in handling the tripole halo, etc) that might actually be missing or wrong in mpi or (more likely) serial.  Generally, people aren't good at maintaining two versions of things, especially when they are only using one.  I think this is an important question.

On the question about ice_shr_reprosum86.c.  Yes, this can be distributed.  It doesn't do anything unless the code is triggered by cpps set by the machine (or possibly the user).  I inquired about removing it, but decided we should keep it.  What is does is deal with integer size mismatches between fortran and c that exist in some older x86 hardware implementations and this is important to deal with if you are going to use the reprosum algorithms to get bit-for-bit correct results.  A side benefit is that including a piece of c code highlighted some problems in our build that got fixed.  I can remove it and then some bfbflag options might not work correctly on some architectures.  I believe the community is unlikely to encounter those architectures.  Having said that, if we don't like carry around the small c code, then that might be a reason to get rid of it.  Otherwise, there is some benefit (both to the computation and general build capabilities) to keep it in the repo.  But I'm open on this.

eclare108213:
I gravitate toward "simpler" when possible, and so my choices for both of these questions would be to simplify and document.  The mpi and serial directories were hand-me-downs from POP, and I'm happy to get rid of them if the codes in them are sufficiently similar.  I think a large reason for doing it that way in the first place was to reduce cpp usage, so if you have to add a cpp anyhow for reproducibility, and it can be used to merge the mpi and serial codes, then it makes sense to merge them.  That sounds like a big project though, suitable for its own PR.  In the meantime, does it make sense to just have one copy of ice_global_reductions.F90?  It would need to be moved up in the directory structure.

Re the c code, I'd prefer not to carry it around, especially if our users are unlikely to be running on the particular architectures that wouldn't be BFB.  I'd rather put a note in the documentation (there's a 'known issues' section somewhere), and perhaps include a pointer to where a user could get this code if he or she desperately needed it.  I did not happen to see where it interfaced into CICE, but perhaps that part could be commented out with instructions for what to do if it were needed.  I'm also open to keeping it, if there are genuine, ongoing benefits to having it, or if my suggestion to document and/or leave comments in the code doesn't make sense.

apcraig:
Good comments.  

I don't think we want to move to a single global_reductions file until we've thought that thru a bit.  In particular, we'd have to add cpps to the build and we'd want to come up with a clean set of cpps to do that.  We wouldn't want to keep the cpp name that is there.  That name is particularly ugly for good reasons.  I would prefer to keep the current files separate (even if they are nearly the same) until we decide to merge mpi and serial, at which point it should be a separate PR.  I think it would be a mistake to have a mixed strategy of some cpps and some split directories.  Right now we have split directories, the cpps that is in global_reductions is local, so it doesn't act like a model cpp.  It is true that the main tradeoff for merging serial and mpi is adding cpps.  But maybe it's just one, like NO_MPI that is triggered by the serial setting.  That setting would trigger a cpp and not a separate file path as it doesn't now.  I'm not arguing for merging or not merging right now, I think we need to think about it and get more input.  In fact, one of the problems with serial and mpi is they are NOT the same and that worries me.  Merging them would be some work with pros and cons.  But I'm thinking there are more pros than cons at this point.

I'm OK pulling out the c code.  What I would propose is that we create a new issue for it, and I do a bit more research on this implementation.  This is basically Pat Worley's code and I had some emails with him a couple weeks ago about just this issue and we went back and forth about whether it should stay.  I think he felt it didn't hurt anything and it was hard to argue against that.  Let me create an issue at high priority to sort this out and then we can remove it if we decide that's OK.

How does that sound?

apcraig:
See issues #288 and #302.

PR #311, opened 2019-05-03
Fix bug in comparebfb.csh when restart data is missing in both test and base directories

phil-blain:
If no restart files are present in neither `base_dir` nor `test_dir`, the script  `comparebfb.csh` would simply exit with the error 

> foreach: No match.

with error code 1 (csh default I guess) and so the script `cice.test` would report 
> Regression baseline and test dataset are different

which is incorrect. I just added `set nonomatch` before the foreach so that if no restart files are present the script does not error but correctly reports :

> base_data: $baseline_dir/$baseline_name/$test_name/restart/iced**
test_data: $baseline_dir/$baseline_name/$test_name/r/restart/iced**
      missing data
Missing data

- Are the code changes bit for bit, different at roundoff level, or more substantial? BFB

No tests
No

- Other Relevant Details:

I stumbled upon that bug because I was running the model for just 3 time steps with a custom test suite and had forgotten to set the `dump_last` namelist variable to true, which caused no restart files to be written since I was running for less than a day.

### DIFF
--- a/configuration/scripts/tests/comparebfb.csh
+++ b/configuration/scripts/tests/comparebfb.csh
@@ -65,6 +65,8 @@ if ($filearg == 1) then
 
 else
   set end_date = `ls -t1 $test_dir | head -1 | sed 's|^.*\([0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]-[0-9][0-9][0-9][0-9][0-9]\).*|\1|'`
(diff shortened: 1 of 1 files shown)


phil-blain:
Sorry Tony for not responding earlier,  I was on vacation last week. Before this PR, the foreach loop was simply skipped with the csh error "foreach: No match.". With my modifications the foreach loop is entered, the paths are correctly formed and "missing data" is reported as expected. Thanks for merging.

PR #312, opened 2019-05-03
Add option to write restarts each time step, add verification for dumpfreq namelist value in ice_init.F90

phil-blain:
Add option to write restarts every time step and add a check in ice_init.F90 to check the dumpfreq value.

- Please suggest code Pull Request reviewers in the column at right. @eclare108213 

BFB
No tests.
No

- Other Relevant Details:

I stumbled upon that when I tried to set dumpfreq to '1' to have restarts at each time step, without checking in the doc if that value was allowed... The model ran fine but no restarts were written at all, so I figured I'd add that option and that check. 

If you feel like that should be an `ERROR` instead of a `WARNING` I can change it.


apcraig:
Maybe we should add an issue that provides an option to dump restarts each timestep.  It looks like "1" is a valid option for histfreq.  We probably should add that option to dumpfreq.  We can either merge this PR and then deal with it later or update this PR to address the feature.  Thanks for identifying this feature.

phil-blain:
I agree. I can update the PR once we decide if we want to have '1' as an option.

apcraig:
I see no reason not to add it as long as it's straight-forward.  I say we should proceed.  Thanks @phil-blain, If you need any help, let me know.

eclare108213:
It wasn't implemented before because I didn't see a real need for it, but if it's useful for you, then go ahead and implement it.  Thanks!

phil-blain:
Done. I also changed one line of the documentation regarding `histfreq`, since it did not accurately reflect the code in ice_calendar.F90.

Issue #171, opened 2018-08-09
configure rectangular "box" case for symmetry tests


eclare108213:

eclare108213:
See PR #316

eclare108213:
I have been unable to make this test work, and so the PR #316 has been closed.  I'll try not to delete that branch, in case someone else wants to work on this problem.

PR #316, opened 2019-05-28
basic symmetry test

eclare108213:
This PR is NOT READY TO BE MERGED.
Basic changes to implement a symmetry test for CICE.
Unfortunately, the test does not work.  More details below.

- Developer(s):  E. Hunke

- Are the code changes bit for bit, different at roundoff level, or more substantial?   should be BFB

not yet tested 

- Does this PR create or have dependencies on Icepack or any other models?   N

- Is the documentation being updated with this PR? (Y/N)   N
If not, does the documentation need to be updated separately at a later time? (Y/N)   Y

- Other Relevant Details:

Added an option file ( -s boxsymmi ) and namelist options for creating a configuration that can be used for testing i-j symmetry, e.g. 
> cice.setup -m badger -p 1x1 -g gbox80 -s boxsymmi,debug --test smoke --testid symm01

This test is currently clunky:  create two cases, and change the namelist option in one of the cases from
> atm_data_type='boxsymmi'

to
> atm_data_type='boxsymmj'.

Then run both of them and compare the diagnostic output by hand.  

The current test fails because of roundoff differences that creep in during the EVP subcycling, due to order-of-operation differences for divune, etc.  I have sets of print statements in ice_dyn_evp.F90, temporarily -- for any given ksub, compare the uvel block from the boxsymmi run with the vvel block from the boxsymmj run. 

There was a working symmetry test in the past (v4, maybe v5?), and I am not sure what changed that causes it to fail now:  different compilers/options, changes to EVP that went in with rEVP, the new EVP kernel, etc.  The behavior does seem related to problems pointed out by @mhrib et al in their testing.   Perhaps this test could be used to ferret out those issues.  It's also possible that the previous test was accomplished in a single run, instead of comparing 2 separate runs, by setting it up symmetrically along the line x=y.  That wouldn't be too difficult to try, using the new option file (but be careful of the B grid).


apcraig:
I have a few comments.
- I would be happy to help setup a "symmetry" test or provide some guidance.  I believe we could create a test that would setup a case to run the i and j options back to back and then compare their results.  This would be fairly easy to do and would automate the testing.  If we decide an x=y symmetry would also be interesting, we could do that instead or in addition.
- Have you tried to reduce the compiler optimization to see if the results are bit-for-bit.  We could try this on say conrad or gordon with 4 compilers to see if it is machine or compiler dependent.  We could try with different levels of optimization with different compilers.
- Before merging, I assume we might need a little documentation, may clean up the set_nml file if needed, and remove or add some ifs around the print statements.  Just thought I'd formally note that.

Let me know if I can help with some testing and/or "symmetry" test development.

eclare108213:
Thank you @apcraig.  I envisioned an automated test and hoped that you or @mattdturner could help write the scripts for that.  But until we can get the test to work, there's not much point in making it beautiful.  Ditto for the documentation, unless we want to merge the code options without an actual test, pending further work. The print statements should be deleted, eventually, but for now they are helpful for tracking differences in the two runs.

This was tested using -s debug, which I believe sets -O0.  

I thought that the differences were appearing due to adding/subtracting terms that should be 0 but might have some roundoff value.  To test that hypothesis I set up another case in which all of the terms are nonzero, and the differences still appear.   If you or someone else can take a look, maybe try this with some other compilers to see if it works at all, anywhere, I'd be grateful.

rallard77:
Elizabeth,

Which compiler(s) did you try? I ran the tests on gordon with the intel 
compiler and am seeing similar issues.

Rick

eclare108213:
Intel also.

apcraig:
I am going to run some tests as well and see what I can find.  Hope to have additional info soon.

apcraig:
Just trying to understand what is expected.  This should be an 80x80 box, NOT on a rotating sphere, (coriolis=0).  The boundary conditions are the same all around closed by open type?  The forcing has small ocean currents, same in both symi and symj.  symi and symj have same constant wind field but differ in the orientation of the wind forcing, one is i direction, other is j direction.  Is there no thermodynamics?  The results should be that u(symi) = v(symj) and v(symi)=u(symj)?  Is that the ultimate test?  Should u and v also be constant across the domain or anything like that?

apcraig:
I may be way off here, but I'm seeing the symmetry break on the first subcycle in stepu here,

         ! finally, the velocity components
         cc1 = strintx(i,j) + forcex(i,j) + taux &
             + umassdti(i,j)*(brlx*uold + revp*uvel_init(i,j))
         cc2 = strinty(i,j) + forcey(i,j) + tauy &
             + umassdti(i,j)*(brlx*vold + revp*vvel_init(i,j))

         uvel(i,j) = (cca*cc1 + ccb*cc2) / ab2 ! m/s
         vvel(i,j) = (cca*cc2 - ccb*cc1) / ab2

cc1 and cc2 are not symmetric in the two runs and uvel/vvel solutions separate there.  Before I start to look at each term in cc1 and cc2, is there something obvious there?  Again, I'm not sure I fully understand what I'm looking for, but just thought I'd throw that out in case anyone has an idea.  Maybe I'm going off track here.

eclare108213:
Regarding the configuration - all of the statements above are correct.  I tested with both open boundaries and closed.  I made the ocean currents small so that the initial u,v for ice wouldn't be zero.  We probably should run with thermo on, but that should be strictly column physics and at this stage the issue is in the dynamics.  u and v will not be constant in any direction, but we want u(symi)=v(symj) and vice versa.

I believe the symmetry break occurs in the stress routine, before reaching stepu (so it will likely appear in strintx/y in cc1/2).  I was seeing it first in the divune etc fields.

apcraig:
Ignore my prior comments.  I think I'm starting to understand the expected symmetry.  It's across the i=j line and means u and v are also swapped.  So v_j(4,9)=u_i(9,4) and u_j(4,9)=v_i(9,4) and vv.  I'm still trying to wrap my head around the diagnostics, but am looking.

apcraig:
I think I found at least one of the issues.  I have a 3 timestep test now running properly.  I will do a longer run.  The problem is in stepu,

         ! divergence of the internal stress tensor
         strintx(i,j) = uarear(i,j)* &
             (str(i,j,1) + str(i+1,j,2) + str(i,j+1,3) + str(i+1,j+1,4))
         strinty(i,j) = uarear(i,j)* &
             (str(i,j,5) + str(i,j+1,6) + str(i+1,j,7) + str(i+1,j+1,8))

which is the sum of offsetting terms and leads to differences at bigger than roundoff.  Now, that seems to mostly be OK, but occasionally it isn't.  The fact that strintx and strinty do not generate the same roundoff differences in the symmetric runs is also odd because the sum is done the same in both (the terms are in the same order).  I guess, because there are on different lines of code, they could be using different parts of the node memory or something else.  It seems like if IEEE math were enforced, we would not get a difference.  Again, I can't fully explain it.  I tried several things to "fix" the issue, but the only one that seemed to work (so far) was doing the sum at quad precision.

      real (kind=r16_kind) :: &
         val16, val16a, val16b, val16c, val16d  ! temporaries for strintx strinty calc

         val16a = real(str(i,j,1),r16_kind)
         val16b = real(str(i,j,2),r16_kind)
         val16c = real(str(i,j,3),r16_kind)
         val16d = real(str(i,j,4),r16_kind)
         val16 = val16a + val16b + val16c + val16d
         strintx(i,j) = uarear(i,j) * real(val16,dbl_kind)
         val16a = real(str(i,j,5),r16_kind)
         val16b = real(str(i,j,6),r16_kind)
         val16c = real(str(i,j,7),r16_kind)
         val16d = real(str(i,j,8),r16_kind)
         val16 = val16a + val16b + val16c + val16d
         strinty(i,j) = uarear(i,j) * real(val16,dbl_kind)

I will try a few other things and do a longer run.  @eclare108213, if you have a chance to verify this fixes the problems as well, that would be great.   Once we confirm this is the only problem, we can talk about what we want to implement in code.  I would suggest we turn on the higher precision computation only for the symmetry test, but again, first lets confirm this fixes the issue and there aren't others.  It could be that there are more than 1 issue like this.  It could also be that different machines and different compilers behave differently.  I have only tested this with intel.  I will try a few other compilers too.

eclare108213:
I can’t look at this today, but it’s worth checking those terms in the edeco branch of the svn repo (on E3SM GitHub) or maybe the v5 trunk, to see if the order changed.


Sent with BlackBerry Work
(www.blackberry.com)

apcraig:
I'm not sure I have it fixed.  I need to look at bit more.

apcraig:
Also, the CICE5 code has identical coding for strintx and strinty.

apcraig:
Just FYI, that change to the strintx/y seems to have only slowed down the roundoff growth.  I don't believe it is the root cause and is not likely to be part of the solution.

eclare108213:
I changed the test so that the symmetry comparison could be done using output from one run.  Still need to clean up, if we decide to stay with this approach.  However the code is still not symmetric.  I was testing with ktherm=-1, kridge=-1, ktransport=-1 (i.e., thermo, ridging and transport turned off) -- apparently this problem is coming from the dynamics (evp in this case).  In the grid cells I'm looking at (see print statements in the code), the problem appears in the second timestep in the 'csig' fields, e.g. compare csigpne with csigpsw for the (x=y)-reflected grid cell.  Everything going into those appear to be the same.   All of these terms have 3 or 9 in their denominators.  I tried rearranging them to pull out the divisions by 9, without luck.  **Is there a compiler flag that truncates values at a certain number of decimals, which we could turn on just for symmetry testing?**

mhrib:
@eclare108213 : "Compiler flag".
If your testing case includes very small numbers, which should be zero, "Float-to-zero : -ftz for intel" can be useful. Affects sign() etc. Can be useful to compare results with/without -ftz flag

Else, for general testing I have used the following flags on our Cray-XC50:
INTEL: "-g -O0 -traceback -fp-model precise -fp-stack-check -fpe0"
GNU: "-g -O0 -fsignaling-nans -fno-reciprocal-math -ftrapping-math -fno-associative-math -fno-unsafe-math-optimizations"
CRAY with OpenMP: "-O1 -Ofp0 -K trap=fp"
CRAY without OpenMP: "-O0 -Ofp0 -K trap=fp"

Maybe these flags can inspire you...

eclare108213:
Roundoff differences come in, apparently from differing order of operations in the EVP stress routine.  After trying the intel compiler options below without further progress, I am closing this PR.  I will leave the associated issue #171 open, in case someone wants to figure out how to make this work, and I will try not to delete this branch.  Compiler options tried (some are redundant):
-g -O0 -traceback -fp-model precise -fp-stack-check -fpe0
with
-fp-model-consistent
-fimf-arch-consistency -no-fma
-fimf-use-svml
-assume protect parens
-qno-opt-dynamic-align

phil-blain:
@eclare108213, no worries about not deleting your branch. Github stores every PR in a ref inside the upstream repo. For example one could recreate your branch locally by doing:
```shell
git fetch upstream pull/316/head:new-local-branch
```

Issue #283, opened 2019-02-21
cice.build "reuse executable" feature


apcraig:
There is currently a hidden feature in cice.build that support binary reuse. It exists solely to speed up the test suite by reusing the binary as much as possible and avoiding redundant builds.  There is some discussion about whether we want to make this feature public.  If we do, then we need to extend the documentation further and probably provide an example. We might even want to modify the interface so it matches the keyword interfaces that is the current standard in cice.setup (ie. cice.build --exe /scratch/$user/ciceexe/cice.exe.vabcd).  The current implementation just takes an optional single argument without a keyword.

I'm not entirely convinced we want/need to publicize this feature, but am open to discussion.  @phil-blain proposed we advertise this feature in PR #282 which was subsequently closed and shifted to this issue.  I think this is a very powerful in the test suite and will be used even for user defined custom suites.  But there are a number of subtle issues in the use of this feature that worries me if we publicize it.   The user has to be aware that certain options (DEBUG, serial vs MPI, threading flags, and actual compiler) create inconsistencies in the executable from case to case and that reusing a binary for a case with different settings may work, but not test the setup associated with the case.  With the test suite, since all of this is automated, the scripts are fully in control.

Anyway, I think this is worth discussing.  Is there value (or risk) in making this feature public.

duvivier:
I'm not sure we want to make this feature public. My feeling is it's a hidden tool for us in terms of the testing for PRs etc. to the repo. It's not a feature to use for science to make things faster for that. I think it should be something *ONLY* used in the test suite and nothing else. To that end, keeping it hidden except to developers makes sense. The question then is how to document it for developers since all our normal documentation is freely available. I'm not sure how to do that...

apcraig:
The feature is automatically used in all test suites without any control by the user.  So we don't need to document it in the classic sense, like "how to turn it on or off".  We could offer a new argument that would turn it off, although I'm not sure that's so valuable.  I guess the documentation that might be useful would support maintenance and extension of the scripts implementation.  We don't want the feature to get lost or broken in any upgrades to the scripts/testing implementation.  Is that the kind of documentation that is needed?

eclare108213:
[...]

PR #322, opened 2019-06-11
Build system improvements

phil-blain:
Improvements to the build system, add cice.make script

BFB (only scripts are touched)

\-
No. I can issue a similar PR for Icepack if you agree with these changes.

- Other Relevant Details:
  - This PR adds a "cice.make" script that is added to the case directory (mostly a slimmed down "cice.build"). This script can be used to call make with the different goals that are defined in the Makefile, which are useful for debugging (db_files, db_flags, clean). I also updated the output of these flags.
  - I also changed the Makefile so that only a single "-I." flag is used for the call to the compiler, instead of having a "-Ipath/to/source/directory" flag for each source directory, which is useless in our case since all compiled modules are put in the ${ICE_OBJDIR} directory, where the compilation takes place (see commit message of 01f38739581e14). I think this is a historical artifact dating from when the code used "include" instead of "use" directives.
  - Finally, I changed the Makefile so that the program 'makdep' is listed as a prerequisite of the rules that build the dependencies, so we don't need to call "make makdep" separately in cice.build to compile it.

### DIFF
--- a/cice.setup
+++ b/cice.setup
@@ -533,7 +533,7 @@ EOF
     endif
 
     # from basic script dir to case
-    foreach file (cice.build cice.settings Makefile ice_in makdep.c setup_run_dirs.csh)
+    foreach file (cice.build cice.make cice.settings Makefile ice_in makdep.c setup_run_dirs.csh)
       if !(-e ${ICE_SCRIPTS}/$file) then
         echo "${0}: ERROR, ${ICE_SCRIPTS}/$file not found"
         exit -1
--- a/configuration/scripts/Makefile
+++ b/configuration/scripts/Makefile
@@ -60,6 +60,7 @@ OBJS  := $(addsuffix .o, $(sort $(basename $(notdir $(SRCS)))))
 DEPS  := $(addsuffix .d, $(sort $(basename $(notdir $(SRCS)))))
 INCS  := $(patsubst %,-I%, $(VPATH) )
 OBJS_DEPGEN := $(addprefix $(ICE_CASEDIR)/,$(addsuffix .c, $(notdir $(DEPGEN))))
+MODDIR:= -I.
 RM    := rm
 
 .SUFFIXES:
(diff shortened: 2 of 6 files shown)


apcraig:
I think a bunch of these changes are good.  I think the biggest question for me is the creation of the cice.make script.  I think the main downsides is that it adds another script to the case directory (where I think less is generally better) and it seems to duplicate part of the cice.build script which creates extra maintenance issues.  So, this leads to a few questions.  Would it be possible for cice.build to be modified so it would support "cice.build target".  In other words, could the cice.make features be built right into the cice.build script.   Next, if we decide to keep the cice.make around, would it make more sense for the cice.build to call it, thereby reducing redundant code.  Finally, if we feel giving users direct access to clean and other make targets is helpful, should we consider taking a bigger step which would be to have cice.build refactored, even getting rid of it all together and ending up with "make clean" or "make cice" as the way to build the model?  I think my preference at this point is to see if cice.build can take an optional argument and get things working that way.  Is that reasonable and do any of my comments make sense?

phil-blain:
Yes, what you say is reasonable and does make sense, thanks for your comments. I did not want to change things to much and that is why I created a new script. Here are my thoughts:
> Would it be possible for cice.build to be modified so it would support "cice.build target". In other words, could the cice.make features be built right into the cice.build script.

I think it would be possible, but would complicate a little bit the logic of cice.build:
1. the success/failure code after the call to make would only make sense if we compile everything, not for other make targets, so we would need additional logic
2. right now the only argument accepted by cice.build is the path to an already built cice executable; we would need to add more checks to verify if the argument is a make goal or a path to an executable
3. I guess we would only write the cice.bldlog file if we indeed compile the model, not call other Make goals, so additional logic would be needed here also

I think all of these are feasible.
> Next, if we decide to keep the cice.make around, would it make more sense for the cice.build to call it, thereby reducing redundant code.

I think that is a good idea if we go that way.

> should we consider taking a bigger step which would be to have cice.build refactored, even getting rid of it all together and ending up with "make clean" or "make cice" as the way to build the model?

From  what I understand about how Make works with VPATH and all that I don't think it would work easily since the Makefile is in ICE_CASEDIR but the model is built in ICE_OBJDIR... and we would still need to call make with the parameters EXEC, VPFILE, MACFILE, as done in cice.build... I think having a wrapper script that is easy to call, as we have now, is a better way.

What do you think? I will update this PR depending on what we decide.

apcraig:
Thanks @phil-blain, several good points.  I think my preference would be to try to modify cice.build if we can make that work well and clean.  I propose the following interfaces

  cice.build [--exe binary] [target]

Lets leave --exe as a hidden feature.  It's really used just for testing at this point.  target is optional and the script should do what it does now if left unset.  Lets check that --exe and target are not set together.

Otherwise, if target is set, what we should do is just set the env as needed then execute the gmake command with the target.  There should be no interaction with README.case or other fluff.  

The targets are dependent what's in the Makefile, and that may not be easy to access.  Is it possible to add a target in the makefile that indicates the possible targets?  Like

  cice.build targets

would return

  cice.build support the following targets: cice, db_files, db_flags, clean, targets

I think we should do something with the "all" target in the Makefile.  Can we add a "cice" target that is the same thing and make that the first target?  Is "all" something that we really need/want in the Makefile to conform?  I'm not a make expert.

Finally, I think if someone does 

  cice.build cice

it should build the model fine without going thru the extra fluff.  While 

  cice.build

goes thru the extra fluff.  Does that make sense?  Thoughts?

phil-blain:
Yes, I think that this design would work well. 
- I can add a target to list the different targets.
- The 'all' target is a make convention, if you type 'make' in a directory with a Makefile in it, make would execute the recipe for the first target in the Makefile, which by convention is 'all' and which builds everything. That is what we have now : cice.build internally calls make with different variables but without any goal, and thus make executes the recipe for the default target, which is 'all', which has EXEC as a prereq and thus builds the whole model. I think we should leave this as is.
- it is a good idea to have a 'cice' target to just build the model without the logging fluff.
I'll start working on that.

apcraig:
@phil-blain , sounds like your plate is full at the moment.   Would you be OK if I took a crack at some of these changes if I get a chance?  If so, I could branch off your branch and PR back to your branch keeping this PR alive.  Does that sound reasonable?  If you prefer to take this on yourself, that's OK too.  Either way is fine.

phil-blain:
Yes I do have a lot of stuff at the moment. If you have a chance to work on this, go ahead.

apcraig:
Just FYI.  I have been working on this the last couple days and should have a PR soon.  I will include the #283 issue in the documentation.

apcraig:
#354 supercedes this PR.  #354 is the current master, with this PR merged in, and further changes.  I will close this for now.  I can be reopened if needed.

PR #325, opened 2019-06-19
Add maximum depth for grounding scheme

phil-blain:
## PR checklist
- [x] Short (1 sentence) summary of your PR: 
    Add a maximum depth at which the basal stress parameterization is active
- [x] Developer(s): 
    J-F Lemieux
- [x] Suggest PR reviewers from list in the column to the right. @eclare108213 @apcraig 
- [ ] Please copy the PR test results link or provide a summary of testing completed below.
    I have not had the time to completely port to our clusters yet (I'm working on that) but I expect the base_suite to not be BFB since these changes will change the ice velocity at some grid points *if* the basal stress parameterization is active (probably it will only change at very few grid points though). I've tried to run the quality control run (5 years) on my local workstation but that was too much to ask, it made my computer shut down...
- How much do the PR code changes differ from the unmodified code? 
    - [ ] bit for bit
    - [ ] different at roundoff level
    - [x] more substantial -> **probably**, not verified
- Does this PR create or have dependencies on Icepack or any other models?
    - [ ] Yes
    - [x] No
- Does this PR add any new test cases?
    - [ ] Yes
    - [x] No
- Is the documentation being updated? ("Documentation" includes information on the wiki or in the .rst files from doc/source/, which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/.)
    - [x] Yes
    - [ ] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [ ] No 
- [x] Please provide any additional information or relevant details below:

Would it be possible for someone to run the base_suite, as well as the quality control suite ? JF wanted this to make it into the release.. I understand if you have other stuff, and I'll keep working on porting (I'm having trouble getting help about using our custom tools...) If I suceed I will update the PR with the test results.

What would be needed to test this:
1. run the base_suite (it will probably not be BFB for the tests where the basalstress is active, i.e. any tests with options alt01, alt03 or boxrestore)
2. run the quality control test (baseline + test), modifying the namelist for both as follows:
    ```
   use_bathymetry=.true.
   basalstress=.true.
    ```
   so that we check if the new code changes will affect the statistical response if the basal stress is active.

Issue #162, opened 2018-07-30
time manager concerns


apcraig:
There have been some long standing time manager questions.  Particularly, the relationship between istep, istep0, istep1, idate, idate0, and dt when intializing, restarting, and changing calendars.  As well as how they impact history and restart file, their time axis, and what should be written to and used on restart files.  And finally how a coupled model differs from a stand-alone run and what should be set in the model coupling layer and initialization.  This probably impacts the coupled models more than others.  But I think it would be nice to have some clearly defined protocol for dealing with this in CICE6 once and for all.

Mostly, all the variables need to be consistent and they need to work with respect to the netcdf time axis convention.  They also need to work with different calendars, on initial and restart runs, initializing a new run from an old restart on a different date and/or with a different calendar.  Also, what happens when the timestep is changed in the middle of the run.

I will try to pull together some documentation that, when ready, can be added to the user guide.

apcraig:
See https://docs.google.com/document/d/15HmyNDbU0x13BlKWZnYjrTpeetzGh6H9z4grVmqBdEc

eclare108213:
see also #139, #23

apcraig:
#289 noted a problem when writing a history or restart file on timestep 1 in some cases.  #289 results in a known bug being issued in the documentation.  But this should be revisited.  The underlying issue is

> The problem actually is that all the logic to turn on the history and restart flag is within an if test,
> 
> if (istep > 1) then
> 
> at about line 280 of ice_calendar.F90. If I change that to (istep > 0), it writes the first history file fine.

apcraig:
#343 notes some issues with leap years.  We will want to make sure we can address leap years in general.  In particular, this comes into play with the internal time manager as well as forcing files (which may be missing leap days or may have them when we aren't using leap year calendars).

apcraig:
I have updated the google doc (see above), would love to have some feedback.  I think I'm going to start working on this in the next few weeks.  I think it's becoming more important all the time.  I plan to build a unit tester, and testing within the model will need to be quite comprehensive.  It will not be a small task, but I think I can make reasonable progress.

TillRasmussen:
Increment of time in initialization seems to be wrong. If you run 1 time step then time and istep will be increased twice. Once in initalization and once in the end of a time step.

A potential solution could be to remove the increment of time and  istep in the beginning of each time step instead of the end

apcraig:
I think we can close this when #566 is merged.  See also #567.

[...]

PR #331, opened 2019-07-09
Initialize revp stresses to previous time step

abouchat:

## PR checklist
    Change stress initialization to previous time step for revp, to be up-to-date with litterature.
- [X] Developer(s): 
    Amélie Bouchat
- [X] Suggest PR reviewers from list in the column to the right. @eclare108213  
    base_suite: 
       210 of 214 tests PASSED
       4 of 214 tests FAILED
       0 of 214 tests PENDING
    The tests that failed are 'alt02' and 'boxrestore' since revp is activated. 
    We are working on running the quality control suite, but still having the same issues reported by @phil-blain in PR #325 .

    - [ ] bit for bit
    - [X] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [] No, does the documentation need to be updated at a later time?
        - [] Yes 
        - [ ] No 
The stresses for revp were currently initialized to zero at the beginning of the subcycling, however in the latest litterature they are initialized to the previous time step (see references below). Changing initialization to the previous time step makes more sense for the evolution of the stresses with subcycling:
ellipse_compare_stress_init

I could not perform the quality control suite since we are having issues with running the test on our environment. @phil-blain and I will keep working on this.

Kimmritz et al., 2017, https://doi.org/10.1016/j.ocemod.2017.05.006
Koldunov et al., 2019, https://doi.org/10.1029/2018MS001485

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_shared.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_shared.F90
@@ -503,7 +503,7 @@
          taubx    (i,j) = c0
          tauby    (i,j) = c0
 
-         if (revp==1) then               ! revised evp
+         if (icetmask(i,j)==0) then 
             stressp_1 (i,j) = c0
             stressp_2 (i,j) = c0
             stressp_3 (i,j) = c0
@@ -516,20 +516,7 @@
             stress12_2(i,j) = c0
             stress12_3(i,j) = c0
             stress12_4(i,j) = c0
-         else if (icetmask(i,j)==0) then ! classic evp
-            stressp_1 (i,j) = c0
-            stressp_2 (i,j) = c0
-            stressp_3 (i,j) = c0
-            stressp_4 (i,j) = c0
-            stressm_1 (i,j) = c0
-            stressm_2 (i,j) = c0
-            stressm_3 (i,j) = c0
-            stressm_4 (i,j) = c0
-            stress12_1(i,j) = c0
-            stress12_2(i,j) = c0
-            stress12_3(i,j) = c0
-            stress12_4(i,j) = c0
-         endif                  ! revp
+         endif                  
       enddo                     ! i
       enddo                     ! j
 
--- a/configuration/scripts/cice.batch.csh
+++ b/configuration/scripts/cice.batch.csh
@@ -165,6 +165,14 @@ cat >> ${jobfile} << EOFB
 #SBATCH --qos=standby
 EOFB
 
+else if (${ICE_MACHINE} =~ brooks*) then
+cat >> ${jobfile} << EOFB
+#PBS -N ${ICE_CASENAME}
+#PBS -j oe
+#PBS -l select=${nnodes}:ncpus=${corespernode}:mpiprocs=${taskpernodelimit}:ompthreads=${nthrds}
+#PBS -l walltime=${batchtime}
+EOFB
+
 else if (${ICE_MACHINE} =~ theia*) then
 cat >> ${jobfile} << EOFB
 #SBATCH -J ${ICE_CASENAME}
--- a/configuration/scripts/cice.launch.csh
+++ b/configuration/scripts/cice.launch.csh
@@ -154,6 +154,18 @@ EOFR
 endif
 
 #=======
+else if (${ICE_MACHINE} =~ brooks*) then
+if (${ICE_COMMDIR} =~ serial*) then
+cat >> ${jobfile} << EOFR
+./cice >&! \$ICE_RUNLOG_FILE
+EOFR
+else
+cat >> ${jobfile} << EOFR
+aprun -n ${ntasks} -N ${taskpernodelimit} -d ${nthrds} ./cice >&! \$ICE_RUNLOG_FILE
+EOFR
+endif
+
+#=======
 else if (${ICE_MACHINE} =~ theia*) then
 cat >> ${jobfile} << EOFR
 #mpirun -np ${ntasks} ./cice >&! \$ICE_RUNLOG_FILE
--- a/configuration/scripts/machines/Macros.brooks_intel
+++ b/configuration/scripts/machines/Macros.brooks_intel
@@ -1,0 +1,78 @@
+#==============================================================================
+# Makefile macros for ECCC brooks
+#==============================================================================
+# For use with intel compiler
+#==============================================================================
+
+CPP        := fpp
+CPPDEFS    := -DFORTRANUNDERSCORE ${ICE_CPPDEFS}
+CFLAGS     := -c -O2 -fp-model precise
+#-xHost
+
+FIXEDFLAGS := -132
+FREEFLAGS  := -FR
+FFLAGS     := -fp-model source -convert big_endian -assume byterecl -ftz -traceback  -diag-disable 5140
+#-xHost
+FFLAGS_NOOPT:= -O0
+
+ifeq ($(ICE_BLDDEBUG), true)
+  FFLAGS     += -O0 -g -check -fpe0 -ftrapuv -fp-model except -check noarg_temp_created
+# -heap-arrays 1024 
+else
+  FFLAGS     += -O2
+endif
+
+SCC   := cc
+SFC   := ftn
+MPICC := cc
+MPIFC := ftn
+
+ifeq ($(ICE_COMMDIR), mpi)
+  FC := $(MPIFC)
+  CC := $(MPICC)
+else
+  FC := $(SFC)
+  CC := $(SCC)
+endif
+LD:= $(FC)
+
+#NETCDF_PATH := /fs/ssm/hpco/tmp/eccc/201402/04/intel-2016.1.150/ubuntu-14.04-amd64-64/
+
+PIO_CONFIG_OPTS:= --enable-filesystem-hints=gpfs 
+
+#PNETCDF_PATH := $(PNETCDF)
+#PNETCDF_PATH := /glade/u/apps/ch/opt/pio/2.2/mpt/2.15f/intel/17.0.1/lib
+
+INCLDIR := $(INCLDIR)
+
+LIB_NETCDF := $(NETCDF_PATH)/lib
+LIB_PNETCDF := $(PNETCDF_PATH)/lib
+LIB_MPI := $(IMPILIBDIR)
+
+#SLIBS   := -L$(LIB_NETCDF) -lnetcdff -lnetcdf -L$(LIB_PNETCDF) -lpnetcdf -lgptl
+#SLIBS   := -L$(LIB_NETCDF) -lnetcdff -lnetcdf
+
+ifeq ($(ICE_IOTYPE), netcdf)
+   INCLDIR += $(shell nf-config --fflags)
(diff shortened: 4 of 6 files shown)


abouchat:
I don't think revp is on with the qc option. I will change it manually.

phil-blain:
@eclare108213 No, the qc test do not include the revp option. In general, the QC test could be more automated; at the moment if any test fail in the `base_suite`  and we want to run the QC test to check the impact of the code change, one has to manually check which options are 'on' for the failing test(s) and run a 'base' QC test and a 'test' QC test with these options added (possibly running several QC tests since some set of options are incompatible). So I think that in general the QC could be more automated @mattdturner @apcraig

mattdturner:
The problem with having a failed test in `base_suite` automatically spawn the QC test is that simulations done for `base_suite` do not output the data files necessary to run the QC test.  Also, the `base_suite` simulations are generally for shorter time periods than are required for the QC test.  No matter what, a new `base` QC test case and a `test` QC test case would need to be generated.

Perhaps a short-term solution could be to modify the `results.csh` script to, if failures are encountered, auto-generate a new script that would create the QC test cases.  I hesitate to have it automatically create the cases and submit the jobs since the simulations are rather lengthy.  However, I'd be happy to get input from others.

phil-blain:
I think that having an automatically generated script to generate the QC test cases would be a great addition, or maybe just add outpout to `results.csh` taht indicate which set options to use when creating the QC test cases.  You are right that automating the case creation and job submission is not necessarily the best idea.

apcraig:
Just pinging, not clear what the status of this PR is.  We should definitely add an revp test to the suite.  Has a qc test been done and is it still needed?

I think qc tests cannot really be completely automated.  When answers change for specific namelist that require qc testing, it makes sense to me that one would want to think carefully about the best set of namelist options in order to isolate the change and demonstrate "same climate".  I'm not sure automation fits there.

phil-blain:
We were waiting for the QC script problem to be fixed. We will run that this week and update the PR with the results.

phil-blain:
Amélie still has the same error as I had with the updated python script ;https://github.com/CICE-Consortium/CICE/issues/333#issuecomment-518804787

abouchat:
Here are some updates about the QC tests:

1) I am not able to run the QC test with the alt02 option. It gives me the same error as mentioned by @phil-blain in #333 . 

2) However, I was able to run the QC test with the boxrestore option, where the revised EVP is also activated. It fails: 
```
INFO:__main__:  ../data/eccc-ppp2/runs/brooks_intel_smoke_gx1_44x1_boxrestore_medium_qc.qc_base_original_boxrestore_no_leap_years
INFO:__main__:  ../data/eccc-ppp2/runs/brooks_intel_smoke_gx1_44x1_boxrestore_medium_qc.qc_test_boxrestore_no_leap_year
INFO:__main__:Quadratic Skill Test Failed for Northern Hemisphere
INFO:__main__:Quadratic Skill Test Failed for Southern Hemisphere
INFO:__main__:
ERROR:__main__:Quality Control Test FAILED
```
Given that the final stress states after few iterations are quite different when they are initialized with the previous time level rather than at zero, it makes sense that the solution is different enough to fail the QC test. 

3) When running with the boxrestore option, there are problems with the option 'use_leap_years' that is enabled in the namelist. I therefore had to turn it off manually for this case to run. I will add this as a new issue (probably related to #162 ?)

apcraig:
An update to master, #346, should fix the qc testing.  Can someone pull the master and try again (sorry about all the redos).  thanks!

phil-blain:
Acutally Matt did it, for the alt02 option the QC test also fails : 
```python
INFO:__main__:  /p/work1/turner/cice_qc_debug/brooks_intel_smoke_gx1_44x1_alt02_medium_qc.qc_base_original_alt02/
INFO:__main__:  /p/work1/turner/cice_qc_debug/brooks_intel_smoke_gx1_44x1_alt02_medium_qc.qc_test_alt02/
INFO:__main__:Quadratic Skill Test Failed for Northern Hemisphere
INFO:__main__:Quadratic Skill Test Failed for Southern Hemisphere
INFO:__main__:
ERROR:__main__:Quality Control Test FAILED

DEBUG:__main__:Northern Hemisphere skill score = 0.979277
DEBUG:__main__:Southern Hemisphere skill score = 0.912965
``

**Comment by apcraig** [comment]
OK, just trying to understand.  Is this correct?  We were able to run the revp qc tests with the box grid all along.  We were unable to run the revp tests with gx1 until the latest fixes to the qc scripts.  The above skill scores (.97, .91) are for gx1 with alt02 turned on, comparing master with this branch.  These values are expected and so we can now merge.

I assume the alt02 is a reasonable test case?  We could also do an out-of-the-box run just turning on revp, is that valuable too?  I guess I wonder whether some of the other settings in alt02 (ncat=1, kitd=0, etc) is giving us the best qc test configuration for testing revp qc.  I'm just asking, not saying what has been done isn't adequate.  As a general rule, I don't think we need to match the qc testing with the test suite results.  The test suite exercises various parts of the code.  I think the qc test should be a "best" test that exercises the new non-bit-for-bit changes.

**Comment by eclare108213** [comment]
I don't think the QC tests are designed to be run with the box configuration.  They rely on (somewhat) realistic simulation of the sea ice thickness, and so I believe they should only be run on the global grids.  Moreover, I believe they should only be run for gx1 configurations, to get proper statistics, although I might not be remembering that correctly.  @proteanplanet will most definitely have more to say...

**Comment by apcraig** [comment]
I agree we should not use the box grid for the qc tests.  But I think that is what was done until the gx1 test was working?  I guess that's why I'm asking for clarification, I'm a little confused about what was done, whether the qc test we want was completed, and whether this is ready to merge.

**Comment by eclare108213** [comment]
Ah, I see.  The gx1 QC tests need to be run for this PR.

**Comment by apcraig** [comment]
So, looking again at the results above, they note the case is 

brooks_intel_smoke_gx1_44x1_alt02_medium_qc

so this was a gx1 test with alt02.  Is that a reasonable test for the revp qc?  And what was compared?  Is this master with alt02 vs change-stress-init-revp with alt02?  master does not have the brooks mods yet, how was master run on brooks?  was some of it done manually?

**Comment by abouchat** [comment]
Yes, we locally added machine files for brooks. I also added these files in this PR for future use at ECCC.

The QC tests were indeed done with gx1. We compare 
i) master with alt02 VS change-stress-init-revp with alt02 
ii) master with boxrestore vs change-stress-init-revp with boxrestore

These two QC tests were done because they were the two options that failed the base suite, since they are the only two with revised EVP activated. 

I can run another test with a different set of options if something else would be better to test the revised EVP here.

**Comment by phil-blain** [comment]
I confirm the alt02 test was run on the gx1 grid (we followed the instructions at https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_testing.html#end-to-end-testing-procedure) to set up the cases, and added "alt02" and "boxrestore" separately (we ran 2 QC tests : 1 with alt02 (base+test) and one with boxrestore (base+test) **both on the gx1 grid**)

We compared the up-to-date master with the changes from this branch. (master with alt02 vs change-stress-init-revp with alt02, that's right). (and the same thing with the boxrestore option)
For the master runs the brooks mods were cherry-picked on top of the latest master (just the changes to Macros.brooks, env.brooks, cice.batch and cice.run in the commit https://github.com/CICE-Consortium/CICE/pull/331/commits/d0246cc05284d92186fcc9aa81d5610689c88651). 

Tony,  you say we were not able to run the QC test with revp all along on gx1. I think maybe this is the first time that it was tried, so we don't know. The fact that for the outputs from machine brooks the python script would fail because the values arrived exactly in the middle of 2 values of the student test look-up table does not mean that it would be the same for other machines/compilers (as you note in #247, we do not do testing across machines/compilers.)

Regarding the QC test with alt02:
> Is that a reasonable test for the revp qc?

@eclare108213 @abouchat what do you think ? do you think we should isolate the change and run a QC case "brooks_intel_smoke_gx1_44x1_medium_qc" (default options) and manually turn on revp ?

**Comment by eclare108213** [comment]
I'm starting to understand what is happening now.  The boxrestore option is intended only for use with one of the box grids, and in particular the restoring option is designed (if you can call it that) for regional grids, not global grids.  Since restoring hasn't been tested on global grids, it's best not to test using that!  This might trip up other users, so maybe we can think about how to prevent these kinds of mistakes.  Likewise for alt02, there might be something in there that doesn't make sense, although this brings up a larger question about cross-testing the many different options available in CICE.  At some point in the past, we decided that was too much to try to do, but we could gradually add more tests as needed.  My preference, for PRs like this, is to just test the default configuration with the settings needed to test the modification, which requires some manual intervention.  I don't think the QC can ever be completely automated.

**Comment by abouchat** [comment]
Ok, I see. I am restarting a QC test that will compare master VS change-stress-init-revp, both with the default configuration + revp manually activated, and on gx1 grid.

**Comment by abouchat** [comment]
Here are the new results for the QC test, comparing master with revp VS change-stress-init-revp with revp, on gx1 grid and all other options set to default:

```
(cice-qc) amb001@eccc-ppp2:>./cice.t-test.py -v ../data/eccc-ppp2/runs/brooks_intel_smoke_gx1_44x1_medium_qc.qc_base_revp ../data/eccc-ppp2/runs/brooks_intel_smoke_gx1_44x1_medium_qc.qc_test_revp
INFO:__main__:  ../data/eccc-ppp2/runs/brooks_intel_smoke_gx1_44x1_medium_qc.qc_base_revp
INFO:__main__:  ../data/eccc-ppp2/runs/brooks_intel_smoke_gx1_44x1_medium_qc.qc_test_revp
INFO:__main__:Quadratic Skill Test Failed for Northern Hemisphere
INFO:__main__:Quadratic Skill Test Failed for Southern Hemisphere
INFO:__main__:
ERROR:__main__:Quality Control Test FAILED
```

and skill scores are:
```
DEBUG:__main__:Northern Hemisphere skill score = 0.969654
DEBUG:__main__:Southern Hemisphere skill score = 0.918635
```

eclare108213:
Thank you @abouchat.  I'm ready to merge this one -- it's an important change.  One final request:  could you make maps of ice thickness from the control and the modified runs, from the same point at/near the end of your runs, and post them in this PR, please?   I'd like to make sure the new run isn't doing something totally unexpected, and the maps also will document the essential difference for future reference.

abouchat:
I have the maps for ice thickness at the end of the run, for the Base case, Test case, and the difference between the two. The pattern is the same overall but the ice thickness is smaller when initializing with the previous time step. I think that what is happening is that, when running with a small number of subcycles and with the stresses initialized at zero, the ice is weaker (since the normalized value of sigma_I is smaller than -1, for all points, as we see from the plots of the stress states I posted with the PR checklist) and therefore it gets thicker by ridging. On the other hand, if the ice is stronger when initializing with the previous time step solution, then it resists more convergence, and does not thicken as much. Does that make sense?

h_base
h_test
h_diff

eclare108213:
That does make sense.  Thank you Amélie !   I hope we can continue working with you in the future.

Issue #347, opened 2019-08-09
standardize QC testing workflow, requirements, documentation


eclare108213:
Define and document a standard workflow for QC testing, including instructions for what to do when the configuration being modified is not part of the default setup.   There should be a control run without the modification with which to compare a second run using the modified code; both runs may need some manual changes.  For modifications that do cause major changes in model output, before and after plots of ice thickness should be made and posted to the relevant PR.

#331 might be used as a test case for defining this workflow.

Also related:  #166 #333

phil-blain:
Also related: #345 to make the plots :) !

mattdturner:
I thought, to some extent, that this was already done.  [Code Compliance Testing Procedure](https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_testing.html#code-compliance-testing-procedure) has some of the software requirements for the QC test, and [End to End Testing Procedure](https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_testing.html#end-to-end-testing-procedure) is an "end-to-end" testing procedure.  I can see that this might need to be expanded upon, though.

eclare108213:
Yes, thanks.  Let's review what's already there and clarify as needed.

eclare108213:
The plots I was thinking of are maps, and #345 is for time series.  We don't have scripts to make maps (yet).

mattdturner:
The current QC script will plot a map of 2-stage test failures, if the 2-stage test does indeed fail (only if all necessary Python packages are available).  Here is an example:

![image](https://user-images.githubusercontent.com/8278916/63449757-5a93e300-c406-11e9-8a75-cd63f409a89a.png)

Note that this QC test was done with randomly perturbed data in order to force a failure.

Should I update the QC script to plot a map of the `base` and `test` ice thicknesses for cases where either the 2-stage test or skill test fail?  Or would it be better to just have a separate script to handle creating maps of data?

mattdturner:
I just realized that the map has the x-axis flipped... I'll have to create a PR to fix that

apcraig:
I believe this issue has been addressed and will close it.  The documentation has been update thru a number of PRs and I think things are working OK and documentation is consistent.  We recently added a few sentences to the qc documentation for further clarification.

> QC testing should be carried out using configurations (ie. namelist settings) that exercise the active code modifications.  Multiple configurations may need to be tested in some cases.  Developers can contact the Consortium for guidance or if there are questions.

PR #355, opened 2019-08-22
Update plotting routines in QC script

mattdturner:
## PR checklist
    Fix an issue where the 2-stage failure map had a flipped x-axis map.  Also add new plotting capability.
- [x] Developer(s): 
    @mattdturner 
    Tested the QC script plotting capability on 4 datasets on 1 server.
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [x] Yes
        - [ ] No 

### DIFF
--- a/configuration/scripts/tests/QC/cice.t-test.py
+++ b/configuration/scripts/tests/QC/cice.t-test.py
@@ -375,32 +375,46 @@ def plot_data(data, lat, lon, units, case):
     warnings.filterwarnings("ignore", category=UserWarning)
 
     # Create the figure and axis
-    fig = plt.figure(figsize=(12, 8))
-    ax = fig.add_axes([0.05, 0.08, 0.9, 0.9])
+    fig, axes = plt.subplots(nrows=1, ncols=2,figsize=(14, 8))
 
+    # Plot the northern hemisphere data as a scatter plot
     # Create the basemap, and draw boundaries
-    m = Basemap(projection='moll', lon_0=0., resolution='l')
-    m.drawmapboundary(fill_color='white')
+    plt.sca(axes[0])
+    m = Basemap(projection='npstere', boundinglat=35,lon_0=270, resolution='l')
     m.drawcoastlines()
+    m.fillcontinents()
     m.drawcountries()
 
-    # Plot the data as a scatter plot
     x, y = m(lon, lat)
     sc = m.scatter(x, y, c=data, cmap='jet', lw=0, s=4)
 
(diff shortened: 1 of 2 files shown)


mattdturner:
The documentation has not been updated yet.  I wanted some feedback on the capability that is being added before updating documentation.  This PR adds a new `--plot` option that will plot the mean ice thickness in a spatial map, similar to the figures below.  It creates plots for the baseline dataset, test dataset, as well as the difference.

image

image

image

This addition of plotting might potentially address a portion of #347

apcraig:
Is there a reason not to have the --plot option on all the time?

The plots look a bit odd with the whitespace between data, but I assume that's an artifact of the plotting package, correct?

These plots are in addition to a plot of the pass/fail status per gridcell?

eclare108213:
This is super helpful, thanks!

mattdturner:
Yeah, the whitespace is a result of the fact that I am using the `scatter` function to create the plots.  I could set up the script to create filled contour plots:

image

or pcolor plots:

image

Or, I could also setup the script to allow all 3 and have the user select which one with an optional argument (while setting a default if no argument is passed).  If this were solely a plotting script, I would likely do that.  However, I'm not sure if we want plotting-specific command line options for a script whose main functionality is to do the QC test.  I wonder if that might get confusing...

I originally set it up to have plotting as an option because I thought that plots would only be desired if tests failed.  I could set it to always plot the ice thickness data, and just include it in a `try` block so that the script continues if the plotting fails (i.e., if the user doesn't have the necessary Python packages).

And yes, this is in addition to the plot of pass/fail grid cells for the 2-stage test

apcraig:
This all is great.  I am fine with any of the plotting options as the default.  I guess I like pcolor the most (but just one opinion) and I would probably turn on the --plot all the time with a package check.  But I'm also happy to merge the PR now.

mattdturner:
I'll have to update the documentation prior to merging the code.  I'll go ahead and do that today and let you know when its ready.

mattdturner:
I added a brief paragraph to the documentation that mentions the plotting capability.  I also modified the script to always attempt to plot the data (I removed the `--plot` option), and added a new `--plot_type` argument that accepts "scatter", "contour", or "pcolor" (with "pcolor" set as the default).

dabail10:
This script is dying on cheyenne with X window related errors. Is there a python version issue here or do I need to do something different with my X window environment?

Dave

Issue #342, opened 2019-08-07
NCAR izumi


dabail10:
Need to add support for NCAR izumi. The machine hobart will eventually go away and so we should move the test suite to izumi.

apcraig:
Done for both icepack and cice

PR #356, opened 2019-08-27
Port to Izumi

apcraig:
## PR checklist
    port to izumi intel, nag, gnu, pgi
- [x] Developer(s): 
    apcraig, dabail10
    Ran new "nothread_suite" on izumi.  Results are here, #daa1689e38 at  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks.  Not perfect but probably acceptable as a starting point.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

izumi does not work well with threading, so a new test suite (nothread_suite) was created that does not leverage threading.  There are some tests with OpenMP on but that is invoked only with 1 thread per task.  Testing on izumi should be done with the nothread_suite.

@mhrid, the nag compiler complained about some code in the OpenMP sections of ice_dyn_evp_1d where "8" was used a kind type.  It said 8 was not valid/known or something like it.  I changed 8 to dbl_kind.  Could you review that for me and make sure it is reasonable.  There are also some places where 4 is used as an int kind in the same area of code.  The compiler didn't complain about that, but it did leave me wondering whether this is correct.  My understanding is the real*8 != kind(8) != selected_real_kind(8) and I assume the same is true with integer and "4".  Could we just confirm that the of real(*,8) and integer(*,4) in that section of code.  I'm happy to update them as needed.

I updated the implementation of the ICE_MACHINE_MAX constraints and added one for ICE_MACHINE_MAXTHREADS and set it to 1 for izumi to help users avoid running on izumi with more than 1 thread.

I removed some hardwired decomp info in the boxrestore nml.  This was creating some problems with some of the tests and don't think it needs to be set there.  The scripts/tests will generate/impose reasonable decomp settings.

I chmod'ed a few files in machines to make them consistent.  This should have no effect but is just cleanup.


apcraig:
Hi @mhrib, I realized after I made some initial changes to the DBL_KIND capitalization that I shouldn't.  I saw some of the documentation in the code.  The latest change from yesterday reverted those changes again.  If you look at the current code changes, those are no longer there.

eclare108213:
Addresses #342

apcraig:
@mhrib and I have had some discussions offline.  He has blessed the changes in the kinds.

The decomposition info in the boxrestore option actually prevented the decomposition to work with many pe counts because slenderX1 is somewhat restrictive.  By removing those, we add some flexibility plus the boxrestore option doesn't overwrite whatever pe/decomp settings we're trying to test if we explicitly set them with the -pes option.

I am going to merge.

PR #357, opened 2019-08-29
Add flexibility to cice.setup for test suites

phil-blain:
## PR checklist
    Add --no-submit and --no-build options to cice.setup
- [x] Developer(s): 
    Philippe Blain
- [x] Suggest PR reviewers from list in the column to the right. @apcraig 
    No changes to code, so I did not perform the full base_suite. I tested to check that the new options work as they should
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [x] Yes
https://external-builds.readthedocs.io/html/cice-consortium-cice/357/user_guide/ug_testing.html#test-suites

        - [ ] Yes
        - [ ] No 

I added two options to cice.setup (for test suites) that I considered useful for my workflow.
1. --no-submit  : this creates the test case folders and compiles the executables but does not submit the jobs. This is useful when doing both a baseline + test (regression test) : I can use` ./cice.setup` to setup, build and submit the 'baseline' base_suite and then `./cice.setup --no-submit` to setup and build the 'test' base_suite (I don't want to submit it immediately since the 'baseline' base_suite might not be finished running, so some tests in the 'test' base_suite might run before the corresponding ones in the 'baseline' suite, which would make them fail). The 'test' suite can then be manually submitted with the existing suite.submit script.
2. --no-build : this only creates the test case folders. The tests can then be built with the new script `suite.build` and then submitted with `suite.submit`. 

Do you think that the options would be better named --build-only (instead of --no-submit) and --setup-only (instead of --no-build) ?

### DIFF
--- a/cice.setup
+++ b/cice.setup
@@ -34,6 +34,7 @@ set stime = `date -u "+%H%M%S"`
 set docase = 0
 set dotest = 0
 set dosuite = 0
+set submit = 1
 
 if ($#argv < 1) then
   set helpheader = 1
@@ -104,6 +105,7 @@ DESCRIPTION
     --testid   : test ID, user-defined id for testing (REQUIRED with --test or --suite)
     --diff     : generate comparison against another case
     --report   : automatically post results when tests are complete
+    --no-submit: for test suites, create the tests and compile the executables but do not submit the jobs
 
 EXAMPLES
     cice.setup --version
@@ -213,6 +215,9 @@ while (1)
 # arguments without settings
   if ("$option" == "--report") then
     set report = 1
+    shift argv
+  else if ("$option" == "--no-submit") then
+    set submit = 0
     shift argv
 
 # arguments with settings
@@ -379,7 +384,7 @@ else
   cp -f ${ICE_SCRIPTS}/tests/report_results.csh ${tsdir}
   cp -f ${ICE_SCRIPTS}/tests/poll_queue.csh ${tsdir}
 
-  foreach file (${tsdir}/suite.run ${tsdir}/suite.submit)
+  foreach file (${tsdir}/suite.build ${tsdir}/suite.run ${tsdir}/suite.submit)
   cat >! $file << EOF0
 #!/bin/csh -f
 
@@ -406,6 +411,7 @@ echo "#vers = ${vers}" >> results.log
 echo "#------- " >> results.log
 EOF0
 
+  chmod +x ${tsdir}/suite.build
   chmod +x ${tsdir}/suite.run
   chmod +x ${tsdir}/suite.submit
   chmod +x ${tsdir}/results.csh
@@ -860,12 +866,13 @@ EOF2
 
     if ( ${dosuite} == 1 ) then
       cd ${ICE_SANDBOX}
-      # Write build and run commands to suite.run and suite.submit
+      # Write build commands to suite.build, suite.run, suite.submit
+      # and run commands to suite.run and suite.submit
 
 cat >> ${tsdir}/results.csh << EOF
 cat ${testname_base}/test_output >> results.log
 EOF
-  foreach file (${tsdir}/suite.run ${tsdir}/suite.submit)
+  foreach file (${tsdir}/suite.build ${tsdir}/suite.run ${tsdir}/suite.submit)
     cat >> $file  << EOF
 echo "-------test--------------"
 echo "${testname_base}"
@@ -882,7 +889,7 @@ EOF
   cat >> ${tsdir}/suite.run << EOF
 ./cice.test
 EOF
-  foreach file (${tsdir}/suite.run ${tsdir}/suite.submit)
+  foreach file (${tsdir}/suite.build ${tsdir}/suite.run ${tsdir}/suite.submit)
     cat >> $file << EOF
 cd ..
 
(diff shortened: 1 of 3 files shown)


apcraig:
I very much like this new feature.  I think I also prefer --setup-only and --build-only as the options.

But I would like to propose a slightly different strategy.  With the current implementation, we are generating suite.build, suite.run. and suite.submit scripts.  These scripts mostly overlap.   What if we created a single script that could handle the different options.  What I envision is a single suite.submit script that would either have a few command line options or variables set at the top of the script that would be implemented something like as follows

```
set dobuild=true
set dosubmit=true
set dorun=true

if (dosubmit == 'true' && dorun == 'true') exit -2

echo "-------test--------------"
echo "conrad_intel_restart_gx3_4x2_debug_diag1_run5day.xs01"
cd conrad_intel_restart_gx3_4x2_debug_diag1_run5day.xs01
source ./cice.settings
if (dobuild == 'true') then
  set ciceexe = "../ciceexe.${ICE_COMPILER}.${ICE_COMMDIR}.${ICE_BLDDEBUG}.${ICE_THREADED}.${ICE_IOTYPE}"
  ./cice.build ${ciceexe}
  if !(-e ${ciceexe}) cp -p ${ICE_RUNDIR}/cice ${ciceexe}
endif
if (dorun == "true") then
  ./cice.test
endif
if (dosubmit == 'true') then
  ./cice.submit | tee -a ../suite.jobs
endif
cd ..
```

There are a few benefits.  First, we avoid having to deal with the buildincremental issues.  Rather than having each case have to turn on incremental builds for the suite to behave in a reasonable way, the suite.submit script would cleanly handle this and the incremental (or not) build would be purely up to the user whether they want to use it or not for other reasons.  Second, we would merge 3 scripts to 1 without the redundancy between scripts.  Third, it may provide easier future extensions to have everything in one script.

The cice.setup would still take --no-build --no-submit options which would then be converted to defaults in the script and then the suite.submit script would be run with the flags set appropriately.

I have suggested and prefer that we use variables in the script to set the flags rather than command line arguments to suite.submit just because the implementation is easier.  I also think these features are not going to be used all that much.  But I am open to further discussion.  Any thoughts on this proposal?

phil-blain:
I opted for the easiest implementation. Regarding the fact that we have 3 scripts: since they are not version controlled but simply created by cice.setup, I don't think it creates much redundancy... 
Also, I think that having 3 scripts is easier from a user's point of view since each script is named after its purpose, and mimic the scripts in a case directory. 
I think it's easier and makes more sense from a UI/UX perspective to run suite.build to compile rather than editing suite.setup to turn on the corresponding variable and then run suite.submit to compile, then later edit it again to submit...

apcraig:
@phil-blain, that makes some sense.  But one issue I have is how the buildincremental needs to be turned on for all this to work OK.  I think that's a complicating factor.  Without using that flag, these new scripts don't provide much use.

As an alternative, If we had a script that just built, just submitted, or did both, that would remove the buildincremental factor.  But then there is also the ability to run interactively or run in batch which you also seem to want to have.  Again, I think very few users are going to use any of this which is why I also think it makes sense for it to be somewhat hidden in one big script.  If it's not hidden in one script, then I think we need to probably have separate .build (just build), .submit (just submit), .run (just run), .build_run (build and run), and .build_submit (build and submit).  I think this is a little over the top.  The point of having the build_run and build_submit is so you can build and run/submit interleaved which is how you want to do it in practice if you are doing it together.  You don't want to build everything then submit everything as that slows the runs down quite a bit.  If it's all in one script, we can support any kinds of modes we want easier.  

I can see advantages and disadvantages in both approaches.  I also understand that these scripts are generated so the redundancy is less of an issue.  But that redundancy has to be maintained in the generation of the scripts, so there is some cost there.  I still prefer the idea that everything is in one script in large part because I think most people will just run the test suites fully automated and never use the other scripts.  But again, that might argue that most folks will never see the scripts anyway.

phil-blain:
I see your point. I just don't link having to hand-edit the script to change it's behavior. If we don't want to do argument-handling in this script, would just using environment variables work for you ? for example depending on the options chosen when calling cice.setup we would have
**cice.setup :**
```shell
setenv SUITE_BUILD=${dobuild}
setenv SUITE_RUN=${dorun}
setenv SUITE_SUBMIT=${dosubmit}
./suite.submit
```

and then suite.submit would just reference these variables and when we want to use the script in other ways we set these variables on the command line:
`SUITE_BUILD=true ./suite.submit`
or something link that.
The defaults would be SUITE_BUILD=true, SUITE_RUN=false, SUITE_SUBMIT=true.

Or go with arguments handling...

apcraig:
I think env variables is a great idea.  So, just to suggest a couple other changes.
- How about if we rename the script to suite_script.csh or something like it?
- The script defaults will always be the same.  If a user invokes --no-build or --no-submit then those flags would be converted to an internal variable in cice.setup and then a setenv command is executed before the suite_script is run.  My suggestion for implementation would be to have the defaults in cice.setup be the same as the script and just always invoke it with the env variables.
- The script needs to check for existance of the env variables before using them like
```
  set dobuild = "true"
  if ($?SUITE_BUILD) set dobuild = "${SUITE_BUILD}"
```
- We should echo dobuild, dorun, and dosubmit at the top of suite_script.csh
- I think it's fine to keep the implementation simple such that "true" is the only matched value.  Anything else (TRUE, false, tru, xyz, etc) will not match "true" in the script.  If anyone else thinks we need some checks (like that true and false are the only valid values), I would be open to something like that too, but am not convinced it's critical.  I prefer to avoid a bunch of code to support some superset of valid values at this point.  If we find that usability is an issue in the community, we can extend it in the future.

apcraig:
See #395, closing now.

PR #377, opened 2019-11-08
Added DMI/nuopc driver

TillRasmussen:

## PR checklist
Added drivers for DMI CICE/NUOPC. cice cap not included.  In addition a bug in the calculation of  anglet was discovered and fixed
- [ ] Developer(s): 
    Till Rasmussen, DMI
  Elizabeth, Tony
  
    ENTER INFORMATION HERE 
    - [ ] bit for bit
    - [x ] more substantial 
The bugfix solves an issue when the angle is averaged to angleT.  It is the result of the discontinuity of averaging angles. In this case they jump from ~-pi to ~pi (instead of ~pi). Previous code resulted in 0. New code converts angles to cos (angle) and sin(angle) as these are continouous, AngleT is most likely only used in coupling interfaces.
    - [ ] Yes
    - [ x] No
    - [ ] Yes
    - [ x] No
    - [ ] Yes
        - [ ] Yes
        - [x ] No 


apcraig:
@TillRasmussen, let me know when you are ready for this to be reviewed.  thanks!

apcraig:
I ran a full test suite on izumi.  The tests ran and passed, but most were not bit-for-bit with the current master.  You can see results here with hash #cbc49686812c9 at https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks.  The box tests seem to be bit-for-bit but other grids mostly not.  I think that would be expected in this case.  

I'll probably run the full suite on another machine and then I can try to do a qc test as well.  Or should I wait for some further updates?  I ran the code as defined on @TillRasmussen master as of this afternoon.  At a minimum,  we know we need an update to the icepack version, but that should not impact CICE runs.  I think we can proceed knowing that there will be no answer changes in icepack even if it's revised and assuming any other changes that come in will be in the dmi driver at this point.  I'll try to move forward with the testing tomorrow.

TillRasmussen:
@apcraig I did not add the cice_cap as I would like to remove the dependency of ESMF from the compilation of cice. It is located here https://github.com/TillRasmussen/DMI_HYCOM_CICE along with the rest of the esmf code that I use. Do you know if there is a way of having the same file in two  repositories. The main reason for doing it like this is that I will be able to compile my cice libary without linking to esmf. If necessary I can also add the cice cap

apcraig:
@TillRasmussen that makes a lot of sense.  There is no requirement that the coupling cap be stored in the CICE repository for any model.  As you point out, there may be good reasons to keep the cap separate.  There are probably a few ways to keep the cap in both repos, but that ultimately involves some combination of manually syncing the files when they change and modifying the build so one or the other may be skipped.  I don't think there is a way to point to the same file in two repos.  You could create a repo just for that nuopc cap file and then make that repo a submodule of both your DMI_HYCOM_CICE and the CICE repos.  If that submodule was dropped in a separate directory, it would be easier to skip it during the build.  But again, that's not a very clean solution.  A final compromise might be to copy the file manually into the CICE Consortium area but rename it so it doesn't get compiled.  Maybe you have a file called cice_nuopc_cap.F90 in your DMI_HYCOM_CICE repo.  Maybe you could copy that to driver/nuopc/dmi/cice_nuopc_cap.F90.for_information_only.  Then it would be visible and probably won't build.  It would be nice for others to see your cap, but what you're doing, keeping it in a separate repo is fine too.

TillRasmussen:
I think that I will copy the file into the cice repo. I would like to solve the other issues first related to the bug

apcraig:
I ran a qc test on gordon and it passed.

> /p/home/apcraig/cice-consortium/cice.tr.nuopc>./cice.t-test.py $WORKDIR/CICE_RUNS/gordon_intel_smoke_gx1_44x1_medium_qc.191115qcb $WORKDIR/CICE_RUNS/gordon_intel_smoke_gx1_44x1_medium_qc.191115qct
> INFO:__main__:Running QC test on the following directories:
> INFO:__main__:  /p/work1/apcraig/CICE_RUNS/gordon_intel_smoke_gx1_44x1_medium_qc.191115qcb
> INFO:__main__:  /p/work1/apcraig/CICE_RUNS/gordon_intel_smoke_gx1_44x1_medium_qc.191115qct
> INFO:__main__:Number of files: 1825
> INFO:__main__:2 Stage Test Passed
> INFO:__main__:Quadratic Skill Test Passed for Northern Hemisphere
> INFO:__main__:Southern Hemisphere data is bit-for-bit
> WARNING:__main__:Error loading necessary Python modules in plot_data function
> WARNING:__main__:Error loading necessary Python modules in plot_data function
> WARNING:__main__:Error loading necessary Python modules in plot_data function
> INFO:__main__:
> INFO:__main__:Quality Control Test PASSED
> 

Based on full test suites on izumi and gordon as well as the QC test, I think we can accept this change.  I will approve this PR, but will wait for @eclare108213's review as well as confirmation that no other changes are desired from @TillRasmussen before merging.  It seems like @TillRasmussen may still want to make a few more changes?  This could be included in this PR or at a later time.  If this PR is updated with non-trivial changes, we may have to retest before merging.  I'm happy to help with that if that's the case.

eclare108213:
@apcraig thanks!  @TillRasmussen let us know when you've finished changes to this PR, and we'll do a final review.  Thx

TillRasmussen:
@apcraig  and @eclare108213 I don't know if it easier to:
a/ add cice_cap.F90 as cice_cap.F90_for_info in a new pull request afterwards?
b/ Can I wait unti after the acceptance of the pull request to update my icepack?

if both are yes then I dont have more changes

apcraig:
@TillRasmussen, I'm fine waiting on both a) and b).  a) has no impact on the Consortium.  As noted above, it would be nice to share your cap, but not necessary.  On b), I'm not sure what is being done with icepack, but testing with this change passes with the current icepack, so I think that's fine.  If there are changes needed in icepack later, we can review them at that point.

@eclare108213, if you are comfortable, feel free to merge the PR.  Or I can do it later as well.

phil-blain:
I think Icepack would be reverted to an earlier commit if this is merged now. 
@TillRasmussen please try the instructions here https://github.com/CICE-Consortium/CICE/pull/372#issuecomment-544638320 to put icepack in the right state and update this branch.

apcraig:
Thanks @phil-blain, I had forgotten about the icepack situation, so wasn't sure what was being asked.  We definitely want that fixed before we merge.

apcraig:
@TillRasmussen , as @phil-blain points out, the process for updating icepack is not difficult and should just take a couple minutes.  I will provide a similar approach below that is maybe a little simpler to understand,

>   checkout/cd to your cice sandbox (ie. git clone https://github.com/tillrasmussen/cice; cd cice; git checkout master; git submodule update --init)
>   cd icepack
>   git checkout 5af4649
>   cd ..
>   git add icepack
>   git commit -m "reset icepack version"
>   git push origin master
> 

I could also create a PR to your repo (or possibly push directly to your repo if I have permission) if that would be easier for you.   I guess we could also just merge this and then I could quickly fix it via a separate PR.

TillRasmussen:
@phil-blain, @apcraig. Thank you for your suggestions. Icepack should now be updated.

apcraig:
Looks good.  I think we can merge.  Will give @eclare108213 a chance for final review.  Just to summarize again for @eclare108213 , full test suites on izumi and gordon with 4 compilers each pass, but most answers change compared to the current master.  A qc test on gordon passed.  The latest change just resets the icepack version to the current default.  Testing was done with a slightly older version of Icepack but I don't think there will be any impact.  Those differences were not associated with source code changes but scripts.  If there are concerns, I could retest, but I don't think it's necessary.

Issue #363, opened 2019-09-18
Modernize MPI interface


phil-blain:
I just noticed that the CICE code uses 2 (very) old ways to get MPI functionality, that is
**a preprocessor 'include' macro**:
```fortran
#include <mpif.h>
```
(in `cicecore/cicedynB/infrastructure/comm/mpi/ice_reprosum.F90` and `cicecore/cicedynB/infrastructure/comm/serial/ice_reprosum.F90` )

or **a Fortran 'include' statement** :
```fortran
include 'mpif.h'
```
(everywhere else)

The `mpif.h` header is [strongly discouraged](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node411.htm#Node411) by the MPI standard and will be deprecated in the future; the more modern way is to use **a Fortran 'use' statement**:
```fortran
use mpi 
! or 
use mpi_f08 ! (the only way that is compatible with the Fortran standard)
```
Reference : https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node408.htm

I think making this change would make CICE more future-proof.

phil-blain:
As an aside, I'm starting to work on coupling CICE6 with NEMO for our systems and the `#include <mpif.h>` statement in **ice_reprosum.F90** makes the NEMO build system fail (it is a [unnecessarily] complex, black-box build system). 
As a first step I propose changing just `cicecore/cicedynB/infrastructure/comm/mpi/ice_reprosum.F90` and `cicecore/cicedynB/infrastructure/comm/serial/ice_reprosum.F90` from a preprocessor macro to a Fortran include statement. @apcraig would you be open to that ? I can make a PR.

apcraig:
That would be good.  Have you verified that will fix your problem in NEMO?

phil-blain:
Yes I did, it does. However I have another problem, should I open a new issue ?

apcraig:
I just merged #365.  If there are other issues, feel free to add them to this one or open a new issue, whatever you prefer.

phil-blain:
I'd vote for reopening since the code modernization issue was not addressed.. maybe we can talk about it at the next teleconference ?

phil-blain:
In #389 we switched to the `mpi` module. We should investigate using the `mpi_f08` module as this is the way forward. 

Basically in `mpi_f08` handles are not integers anymore but of custom type, so I don't think the code changes would be that big.

apcraig:
[...]

PR #389, opened 2019-12-09
MPI: use mpi module instead of Fortran include statement

phil-blain:
## PR checklist
    Switch to `mpi` Fortran module instead of Fortran `include` statement
- [x] Developer(s): 
    P Blain.
- [x] Suggest PR reviewers from list in the column to the right. @eclare108213 @apcraig 
    **I did checked that it compiles correctly, was not sure if full-blown tests were necessary.**
    - [ ] bit for bit -> should be
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 

This is a first step towards #363. I tried to use the `mpi_f08` module but this required more extensive changes to the code, so I chose to do it incrementally by first switching to the `mpi` module.

I had done this a while ago bu had not submitted it yet.


phil-blain:
The reason why it compiled correctly before is because the old way for CICE to have MPI functionality (`include 'mpif.h'`) does not allow the compiler to check that subroutine interfaces are correctly used. 

So I think all along we were not correctly using these subroutines. Since they are now defined in a module, the compiler automatically generates an interface and can thus check that they are called correctly. Citing the [latest MPI specification](https://www.mpi-forum.org/docs/mpi-3.1/mpi31-report/node408.htm) (emphasis added): 

> MPI defines three methods of Fortran support:
> 1. USE mpi_f08: This method is described in Section Fortran Support Through the mpi_f08 Module . It requires compile-time argument checking with unique MPI handle types and provides techniques to fully solve the optimization problems with nonblocking calls. This is the only Fortran support method that is consistent with the Fortran standard (Fortran 2008 + TS 29113 and later). This method is highly recommended for all MPI applications. 
> 2. USE mpi: This method is described in Section Fortran Support Through the mpi Module and **requires compile-time argument checking**. Handles are defined as INTEGER. This Fortran support method is inconsistent with the Fortran standard, and its use is therefore not recommended. It exists only for backwards compatibility. 
> 3. INCLUDE 'mpif.h': This method is described in Section Fortran Support Through the mpif.h Include File . The use of the include file mpif.h is strongly discouraged starting with MPI-3.0, **because this method neither guarantees compile-time argument checking** nor provides sufficient techniques to solve the optimization problems with nonblocking calls, and is therefore inconsistent with the Fortran standard. It exists only for backwards compatibility with legacy MPI applications.

The reason it *ran* correctly before... that's murkier. It's kind of above my technical knowledge, but I can speculate: 
- [MPI_Abort](https://www.open-mpi.org/doc/v3.0/man3/MPI_Abort.3.php) : we were not sending the 3rd argument, so our `ierr` was interpreted as the `ERRORCODE` argument. I guess whatever set the `IERROR` argument inside the MPI implementation did not complain that it was not given this argument... maybe it was writing it at a random location in memory.
- [MPI_SEND](https://www.open-mpi.org/doc/v3.0/man3/MPI_Send.3.php): we were not sending the last argument, so our  `status` flag was used inside MPI_SEND functions as `IERROR`. The additional argument `ierr` was probably ignored at compilation time.
- [MPI_GROUP_RANGE_INCL](https://www.open-mpi.org/doc/v4.0/man3/MPI_Group_range_incl.3.php), where I had to [correct the dimension of the `range` argument](https://github.com/CICE-Consortium/CICE/pull/389/files#diff-0a36efae32a525d2864ef4a1d22fce61L190-L219): I guess the memory layout for `integer, dimension(3)` and `integer, dimension (3,1)` is the same or close enough so it worked...

phil-blain:
Also, the `MPI_Send` calls worked fine unmodified using the Intel 2016 compiler, but Intel 2019  complained. So both the MPI implementation and the compiler vendor and version can influence how the MPI specification is enforced.

apcraig:
@phil-blain, thanks for the clarifications/speculation.  That makes sense to me.  It's surprising to me that there was never a problem before.  These have probably been implemented incorrectly for a decade or more!

PR #393, opened 2020-01-10
Add conda environment to run on personal computers

phil-blain:
## PR checklist
    Add a conda environment spec file describing the needed dependencies to compile and run CICE on personal computers. Also update the documentation with instructions to use this environment.
- [x] Developer(s): 
    Philippe Blain
@eclare108213 @CICE-Consortium/devteam (anyone interested can try it!)
   No tests.
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

This PR adds a "environment.yml" file that describes the required software needed to compile and run CICE under a conda environment. I chose to use the conda package and environment manager because:
- It is cross platform, so it works the same way under Linux and macOS (and Windows, although the CICE scripts won't work so we don't support it)
- It is user-space, so users don't need administrative rights to install it 
- It is has binary-only packages, so that there is no need to compile NetCDF or OpenMPI.
- It is a good way to create isolated, reproducible environments.

@eclare108213 please test the procedure and tell me if there is any hiccups. I will update the instructions accordingly.

### DIFF
--- a/configuration/scripts/Makefile
+++ b/configuration/scripts/Makefile
@@ -119,10 +119,6 @@ db_flags:
 # the needed macros
 #-------------------------------------------------------------------------------
 
-ifndef $(CFLAGS_HOST)
-  CFLAGS_HOST :=
-endif
-
 $(DEPGEN): $(OBJS_DEPGEN)
 	@ echo "Building makdep"
 	$(SCC) -o $@ $(CFLAGS_HOST) $<
--- a/configuration/scripts/cice.batch.csh
+++ b/configuration/scripts/cice.batch.csh
@@ -242,6 +242,12 @@ cat >> ${jobfile} << EOFB
 # nothing to do
 EOFB
 
+else if (${ICE_MACHINE} =~ conda*) then
+cat >> ${jobfile} << EOFB
+# nothing to do
+EOFB
+
+
 else
   echo "${0} ERROR: ${ICE_MACHINE} unknown"
   exit -1
--- a/configuration/scripts/cice.launch.csh
+++ b/configuration/scripts/cice.launch.csh
@@ -213,6 +213,19 @@ endif
 #EOFR
 
 #=======
+else if (${ICE_MACHINE} =~ conda*) then
+if (${ICE_COMMDIR} =~ serial*) then
+cat >> ${jobfile} << EOFR
+./cice >&! \$ICE_RUNLOG_FILE
+EOFR
+else
+cat >> ${jobfile} << EOFR
+mpirun -np ${ntasks} ./cice >&! \$ICE_RUNLOG_FILE
+EOFR
+endif
+
+
+#=======
 else
   echo "${0} ERROR ${ICE_MACHINE} unknown"
   exit -1
--- a/configuration/scripts/machines/Macros.conda_linux
+++ b/configuration/scripts/machines/Macros.conda_linux
@@ -1,0 +1,51 @@
+#==============================================================================
+# Makefile macros for conda environment, GNU/Linux systems
+#==============================================================================
+
+# Preprocessor macros
+CPPDEFS    := -DFORTRANUNDERSCORE ${ICE_CPPDEFS}
+
+# Flags for the C compiler
+CFLAGS     := -c -O2
+
+# Flags for the Fortran compiler
+FREEFLAGS  := -ffree-form
+FFLAGS     := -fconvert=big-endian -fbacktrace -ffree-line-length-none
+
+# Additional flags for the Fortran compiler when compiling in debug mode
+ifeq ($(ICE_BLDDEBUG), true)
+  FFLAGS   += -O0 -g -fcheck=bounds -finit-real=nan -fimplicit-none -ffpe-trap=invalid,zero,overflow
+else
+  FFLAGS   += -O2
+endif
+
+# C and Fortran compilers and MPI wrappers
+SCC   := gcc
+SFC   := gfortran
+MPICC := mpicc
+MPIFC := mpifort
+
+ifeq ($(ICE_COMMDIR), mpi)
+  FC := $(MPIFC)
+  CC := $(MPICC)
+else
+  FC := $(SFC)
+  CC := $(SCC)
+endif
+
+# Linker
+LD:= $(FC)
+
+# Location of the compiled Fortran modules (NetCDF)
+MODDIR  += -I$(CONDA_PREFIX)/include
+
+# Libraries to be passed to the linker
+SLIBS   := -lnetcdf -lnetcdff
+
+# Necessary flag to compile with OpenMP support
+ifeq ($(ICE_THREADED), true)
+   LDFLAGS += -fopenmp
+   CFLAGS  += -fopenmp
+   FFLAGS  += -fopenmp
+endif
+
--- a/configuration/scripts/machines/Macros.conda_macos
+++ b/configuration/scripts/machines/Macros.conda_macos
@@ -39,6 +39,9 @@ LD:= $(FC)
 # Location of the compiled Fortran modules (NetCDF)
 MODDIR  += -I$(CONDA_PREFIX)/include
 
+# Location of the system C header files (required on recent macOS to compile makdep)
+CFLAGS_HOST = -isysroot$(shell xcrun --show-sdk-path)
+
 # Libraries to be passed to the linker
 SLIBS   := -lnetcdf -lnetcdff
 
--- a/configuration/scripts/machines/Macros.conda_macos
+++ b/configuration/scripts/machines/Macros.conda_macos
@@ -1,0 +1,51 @@
+#==============================================================================
+# Makefile macros for conda environment, macOS systems
+#==============================================================================
+
+# Preprocessor macros
+CPPDEFS    := -DFORTRANUNDERSCORE ${ICE_CPPDEFS}
+
+# Flags for the C compiler
+CFLAGS     := -c -O2
+
+# Flags for the Fortran compiler
+FREEFLAGS  := -ffree-form
+FFLAGS     := -fconvert=big-endian -fbacktrace -ffree-line-length-none
+
+# Additional flags for the Fortran compiler when compiling in debug mode
+ifeq ($(ICE_BLDDEBUG), true)
+  FFLAGS   += -O0 -g -fcheck=bounds -finit-real=nan -fimplicit-none -ffpe-trap=invalid,zero,overflow
+else
+  FFLAGS   += -O2
+endif
+
+# C and Fortran compilers and MPI wrappers
+SCC   := clang
+SFC   := gfortran
+MPICC := mpicc
+MPIFC := mpifort
+
+ifeq ($(ICE_COMMDIR), mpi)
+  FC := $(MPIFC)
+  CC := $(MPICC)
+else
+  FC := $(SFC)
+  CC := $(SCC)
+endif
+
+# Linker
+LD:= $(FC)
+
+# Location of the compiled Fortran modules (NetCDF)
+MODDIR  += -I$(CONDA_PREFIX)/include
+
+# Libraries to be passed to the linker
+SLIBS   := -lnetcdf -lnetcdff
+
+# Necessary flag to compile with OpenMP support
+ifeq ($(ICE_THREADED), true)
+   LDFLAGS += -fopenmp
+   CFLAGS  += -fopenmp
+   FFLAGS  += -fopenmp
+endif
+
--- a/configuration/scripts/machines/env.conda_linux
+++ b/configuration/scripts/machines/env.conda_linux
@@ -8,7 +8,7 @@ endif
 if ("$inp" != "-nomodules") then
 
 # Init conda
-source `conda info --base`/etc/profile.d/conda.csh
+source `$CONDA_EXE info --base`/etc/profile.d/conda.csh
 # Activate "cice" conda environment
 conda activate cice
 
--- a/configuration/scripts/machines/env.conda_linux
+++ b/configuration/scripts/machines/env.conda_linux
@@ -8,9 +8,21 @@ endif
 if ("$inp" != "-nomodules") then
 
 # Init conda
+if ! $?CONDA_EXE then
+  echo ""
+  echo "${0}: conda executable not found, see the CICE documentation for how to initialize your login shell to use conda"
+  echo ""
+  exit 1
+endif
 source `$CONDA_EXE info --base`/etc/profile.d/conda.csh
 # Activate "cice" conda environment
 conda activate cice
+if $status then
+  echo ""
+  echo "${0}: 'cice' conda environment not found, see the CICE documentation for how to create the conda cice env"
+  echo ""
+  exit 1
+endif
 
 endif
 
--- a/configuration/scripts/machines/env.conda_linux
+++ b/configuration/scripts/machines/env.conda_linux
@@ -1,0 +1,29 @@
+#!/bin/csh -f
+
+set inp = "undefined"
+if ($#argv == 1) then
+  set inp = $1
+endif
+
+if ("$inp" != "-nomodules") then
+
+# Init conda
+source `conda info --base`/etc/profile.d/conda.csh
+# Activate "cice" conda environment
+conda activate cice
+
+endif
+
+setenv ICE_MACHINE_ENVNAME conda
+setenv ICE_MACHINE_COMPILER conda
+setenv ICE_MACHINE_MAKE make
+setenv ICE_MACHINE_WKDIR  $HOME/cice-dirs/runs
+setenv ICE_MACHINE_INPUTDATA $HOME/cice-dirs/input
+setenv ICE_MACHINE_BASELINE $HOME/cice-dirs/baseline
+setenv ICE_MACHINE_SUBMIT " "
+setenv ICE_MACHINE_TPNODE 4
+setenv ICE_MACHINE_ACCT P0000000
+setenv ICE_MACHINE_QUEUE "debug"
+setenv ICE_MACHINE_BLDTHRDS 4
+setenv ICE_MACHINE_QSTAT " "
+
--- a/configuration/scripts/machines/env.conda_macos
+++ b/configuration/scripts/machines/env.conda_macos
@@ -8,7 +8,7 @@ endif
 if ("$inp" != "-nomodules") then
 
 # Init conda
-source `conda info --base`/etc/profile.d/conda.csh
+source `$CONDA_EXE info --base`/etc/profile.d/conda.csh
 # Activate "cice" conda environment
 conda activate cice
 
--- a/configuration/scripts/machines/env.conda_macos
+++ b/configuration/scripts/machines/env.conda_macos
@@ -1,0 +1,29 @@
+#!/bin/csh -f
+
+set inp = "undefined"
+if ($#argv == 1) then
+  set inp = $1
+endif
+
+if ("$inp" != "-nomodules") then
+
+# Init conda
+source `conda info --base`/etc/profile.d/conda.csh
+# Activate "cice" conda environment
+conda activate cice
+
+endif
+
+setenv ICE_MACHINE_ENVNAME conda
+setenv ICE_MACHINE_COMPILER macos
+setenv ICE_MACHINE_MAKE make
+setenv ICE_MACHINE_WKDIR  $HOME/cice-dirs/runs
+setenv ICE_MACHINE_INPUTDATA $HOME/cice-dirs/input
+setenv ICE_MACHINE_BASELINE $HOME/cice-dirs/baseline
+setenv ICE_MACHINE_SUBMIT " "
+setenv ICE_MACHINE_TPNODE 4
+setenv ICE_MACHINE_ACCT P0000000
+setenv ICE_MACHINE_QUEUE "debug"
+setenv ICE_MACHINE_BLDTHRDS 4
+setenv ICE_MACHINE_QSTAT " "
+
--- a/configuration/scripts/machines/env.conda_macos
+++ b/configuration/scripts/machines/env.conda_macos
@@ -8,9 +8,21 @@ endif
 if ("$inp" != "-nomodules") then
 
 # Init conda
+if ! $?CONDA_EXE then
+  echo ""
+  echo "${0}: conda executable not found, see the CICE documentation for how to initialize your login shell to use conda"
+  echo ""
+  exit 1
+endif
 source `$CONDA_EXE info --base`/etc/profile.d/conda.csh
 # Activate "cice" conda environment
 conda activate cice
+if $status then
+  echo ""
+  echo "${0}: 'cice' conda environment not found, see the CICE documentation for how to create the conda cice env"
+  echo ""
+  exit 1
+endif
 
 endif
 
--- a/configuration/scripts/machines/environment.yml
+++ b/configuration/scripts/machines/environment.yml
@@ -1,0 +1,18 @@
+name: cice
+channels:
+  - conda-forge
+  - nodefaults
+dependencies:
+# Build dependencies
+  - compilers
+  - netcdf-fortran
+  - openmpi
+  - make
+# Python dependencies for plotting scripts
+  - numpy
+  - matplotlib-base
+  - basemap
+  - netcdf4
+# Python dependencies for building the HTML documentation
+  - sphinx
+  - sphinxcontrib-bibtex
--- a/doc/source/user_guide/ug_running.rst
+++ b/doc/source/user_guide/ug_running.rst
@@ -601,7 +601,7 @@ On macOS:
 .. code-block:: bash
 
   # Download the Miniconda installer to ~/Downloads/miniconda.sh
-  wget https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O ~/Downloads/miniconda.sh
+  curl -L https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -o ~/Downloads/miniconda.sh
   # Install Miniconda
   bash ~/Downloads/miniconda.sh
   
--- a/doc/source/user_guide/ug_running.rst
+++ b/doc/source/user_guide/ug_running.rst
@@ -465,7 +465,7 @@ and write permissions such that a set of users can update the inputdata area as
 new datasets are available.
 
 CICE input datasets are stored on an anonymous ftp server.  More information about
-how to download the input data can be found at https://github.com/CICE-Consortium/CICE/wiki.
+how to download the input data can be found at https://github.com/CICE-Consortium/CICE/wiki/CICE-Input-Data.
 Test forcing datasets are available for various grids at the ftp site.  
 These data files are designed only for testing the code, not for use in production runs 
 or as observational data. Please do not publish results based on these data sets.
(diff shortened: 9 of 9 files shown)


phil-blain:
Formatted documentation can be read here : 
https://external-builds.readthedocs.io/html/cice-consortium-cice/393/user_guide/ug_running.html#running-cice-on-laptop-personal-computers

phil-blain:
Here are a little more details about how it works under the hood:

> First, the documentation says the model is always run interactively (./cice.run) but the launch scripts suggest otherwise. The launch script runs with mpirun in mpi model and interactively in serial mode.

I mean interactively as not batch (there is no job scheduler involved). Even in MPI mode the model will run interactively, i.e. the user types `./cice.run` (or `./cice.submit`), and he/she does not get a shell prompt back before the model is done runnnig (all MPI processes are done). 

> Second, how does the conda environment.yml file really work?

The environment file lists the required packages and a conda channel (`conda-forge`) to download them from. Conda-forge is, to cite [their homepage](https://conda-forge.org/):
> A community-led collection of recipes, build infrastructure and distributions for the conda package manager.

It is important to download all packages from the same channel to keep ABI compatibility between the packages.

>  It doesn't seem to specify any particular compiler, netcdf, or openmpi version or even vendor (gnu, intel, etc). 

How conda-forge works is that the whole stack is built using GCC on linux and the LLVM toolchain on macOS (with one exception: Gfortran is the Fortran compiler). There is no other choice with conda.

The `compilers` conda package will install the *same* compilers as are used by the Conda-forge CI systems to build the packages. Again, this is to ensure ABI compatibility. So on Linux we get gcc/gfortran and on Mac clang/gfortran. 

Regarding version of dependencies: since no versions are specified in the environment.yml file, the dependency solver of the conda package manager will do the job to choose mutually compatible versions of *all* listed packages. At the moment it installs (at least on my Linux machine):

- gcc/gfortran 7.3.0
- NetCDF (`libnetcdf` package) 4.7.3
- OpenMPI 4.0.2
and several other required packages.

It installs the same versions on mac, expect Clang instead of GCC as the C compiler.

Conda will refuse to create an environment with mutually incompatible packages. What I had in mind is that if ever there is a dependency conflict we can adjust the environment file accordingly by pinning some packages to specific versions.

apcraig:
@phil-blain Thanks for the clarifications.  That makes sense.  I guess I'd prefer to be able to explicitly specify versions so we have complete control.  I worry that a nice suite of compatible versions as determined by conda could, at some point, have a compiler bug that affects us and that it could be hard to avoid it.  The other issue is that if baselines are generated at different times by different compiler versions, answers could change.  Again, we can control that by controlling when compiler versions are updated.

Just a quick followup.  When is the local conda environment created/updated.  Obviously, first install is important.  Then does each case, each build, each run potentially update the conda environment with new versions?  Or is the conda environment static until the user does an update?  Maybe some of this information should be included in the documentation?

phil-blain:
The environment is created when the user runs `conda env create -t environment.yml` as per the instructions. It is not never updated if the user does not take explicit steps (`conda update`). 

So you are right that we want to tell users to *not* touch the environment after it is created. I'll add it to the doc.

phil-blain:
@apcraig Another thought: I think it's good that the model gets test with different versions of its dependencies. If ever we have a report at some point in time that the environmment installs versions of dependencies that CICE is not compatible with, then we can go back and edit the environment file to pin our dependencies to a last known working version. 

What do you think ?

apcraig:
@phil-blain I agree that's a good plan.  I still need to test and will try to do so in the next few days.  Thanks!

phil-blain:
I just edited the PR description to add another positive point of conda:
- It is user-space, so users don't need administrative rights to install it

apcraig:
I tried to build CICE on my mac laptop and there are a few problems.  I'm not sure everything needs to be revised, but this is where I ran into problems.

My laptop doesn't have wget by default, so I ended up just pointing to https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh in my browser and downloading that way.

Once I installed the conda download, "which conda" returned nothing.  I had to formally "source $HOME/miniconda3/etc/profile.d/conda.csh" in order to get conda started.  Thereafter, I have to do that source in order to use the conda environment whenever I fire up a terminal window.  I have created an alias for that.  It certainly would be possible to automatically source conda whenever someone opens a terminal window.  We probably need to make a recommendation.

When I tried to build the case, I ran into an error,

> Building makdep
> clang -o makdep  /Users/tcraig/Desktop/cice-consortium/cice.conda/case01/makdep.c
> /Users/tcraig/Desktop/cice-consortium/cice.conda/case01/makdep.c:19:10: fatal error: 'stdio.h' file not found
> #include <stdio.h>         /* printf, puts */
>          ^~~~~~~~~
> 1 error generated.
> 

That's where I stopped for now.  Any thoughts on how to resolve the stdio.h error?   I tried a few thing like manually doing "conda activate cice" but haven't figured it out yet.

phil-blain:
Hi Tony ! thanks for giving it a try.

> My laptop doesn't have wget by default, so I ended up just pointing to https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh in my browser and downloading that way.

I guess you mean https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh since you are on mac ?

> Once I installed the conda download, "which conda" returned nothing. I had to formally "source $HOME/miniconda3/etc/profile.d/conda.csh" in order to get conda started. 

Basically the [installation instructions for macOS](https://conda.io/projects/conda/en/latest/user-guide/install/macos.html) say: 

> 1. Download the installer:
> 
> 2. Install:
> 
> - Miniconda---In your terminal window, run:
>   ~~~bash
>   bash Miniconda3-latest-MacOSX-x86_64.sh
>   ~~~
> 4. Follow the prompts on the installer screens.
> 
> 5. To make the changes take effect, close and then re-open your terminal window.

If you follow the installation prompts, the installer will asks you if you want to initialize your shell to use conda. This should add the conda initialization sequence to your  `.bash_profile` and make the `conda` command available when you start up your shell. That is why step 5 mentions that you need to close and reopen your terminal window (as I also mention in the CICE doc). Did you do that ?

You mention that you need to "`source $HOME/miniconda3/etc/profile.d/conda.csh` to get conda started". Csh is not the default shell in macOS, did you change your login shell manually ? Aside: Csh is not supported by conda (but tcsh is), but I think that's ok because on mac /bin/csh is really tcsh.


> When I tried to build the case, I ran into an error,
> 
> ```
> Building makdep
> clang -o makdep /Users/tcraig/Desktop/cice-consortium/cice.conda/case01/makdep.c
> /Users/tcraig/Desktop/cice-consortium/cice.conda/case01/makdep.c:19:10: fatal error: 'stdio.h' file not found
> #include <stdio.h> /* printf, puts */
> ^~~~~~~~~
> 1 error generated.
> ```
> 

Ok, tha's a bigger problem. What version of macOS are you running ? Apple is making it harder and harder for open-source developer in each release...  I'm  doing some research and I think I'm missing some steps if I want the procedure to work on macOS >= 10.13... I can reproduce your error on both 10.13 and 10.14:: 
I'm using this script https://travis-ci.com/phil-blain/CICE-conda/jobs/275049909/config 
result on 10.13 (same error you get):
https://travis-ci.com/phil-blain/CICE-conda/jobs/275049909
same result on 10.14:
https://travis-ci.com/phil-blain/CICE-conda/jobs/275049910

phil-blain:
Also:
In the build log we can see that the compilation of makdep fails, but then the build still goes on (and of course eventually fails because make has no dependency info). That's a bug in our Makefile that should be easy to fix (compilation should abort if makdep can't be compiled).

apcraig:
Just a quick followup.  My default shell on my mac is csh/tcsh.  I am running 10.14.2 on my laptop.  I am also running 10.11.6 on my desktop mac which I will try to do an install on as well at some point just to check the procedure.  In general, I haven't download a bunch of "software tools" like Xcode or whatever it's called now.  I think we should assume users have not done that.

I did logout of my terminal and log back in.  I also tried working with bash instead of csh and I still think it didn't know about conda.  You also have this command in your documentation "conda config --set auto_activate_base false".  Does that turn off the auto activate?  I'm not sure we should recommend auto activation anyway.  And whether people use it or not, we should let them know they may need to manually source the conda profile file if it doesn't exist.  I'm think conda should work with both bash and csh/tcsh.

I agree that the bigger problem is the compiler problem for now.  The other stuff is just settling on process and documentation.  Thanks for working on this.  If we can get something working, that would be awesome.  One really nice thing is that users don't even need to "port" if this works as proposed.  Just setup the env with conda and then the CICE port is already there.  That's cool.

phil-blain:
> My default shell on my mac is csh/tcsh

Ok. I did not check that case as I guess most users will not ever change their login shell, so they get Bash (or Zsh [starting with macOS 10.15](https://support.apple.com/en-us/HT208050)). 
I checked, and the miniconda installer initializes only the Bash shell on mac/linux, so you would have to do 
~~~shell
source $HOME/miniconda3/etc/profile.d/conda.csh
conda init tcsh
~~~
for your `~/.tcshrc` startup file to be written with the conda startup sequence. 

Then you would have access to the `conda` command at shell startup. **Caveat:** If your login shell command is configured to  `csh` and not `tcsh`, then even if `csh` is really `tcsh`, maybe when it is when invoked as `csh` it does not read `~/.tcshrc`, so I'm not sure. But I think it will work in both cases.

---

> I am also running 10.11.6 on my desktop mac which I will try to do an install on as well at some point just to check the procedure. 

That would be super. We want the procedure to work on as many OS versions as possible.

I created a small project here: https://github.com/phil-blain/CICE-conda that gets tested using TravisCI (on 10.10 - 10.14) and GitHub actions (on 10.15). With the commits I just pushed all versions work.

>  In general, I haven't download a bunch of "software tools" like Xcode or whatever it's called now. I think we should assume users have not done that.

It's true that we should not assume that our users have Xcode or the command line tools installed, but I *think* it will be required to have at least the command line tools on macOS >= 10.14. I could not test it yet as I don't have access to a Mac running either 10.14 or 10.15 that *do not* have Xcode installed (all macOS CI images either on Travis or GitHub of course have Xcode installed...)

> I agree that the bigger problem is the compiler problem for now. 

With the two commits I just pushed it should work on 10.14 and 10.15 also (with the caveat that maybe Xcode/the command line tools needs to be installed). Once this is confirmed I will add a mention in the doc as to how to install the command line tools ( `xcode-select --install`).

--- 

>  I also tried working with bash instead of csh and I still think it didn't know about conda. 

If you installed miniconda and answered 'yes'' to 
> Do you wish the installer to initialize Miniconda3
by running conda init? [yes|no]

Then I think that the miniconda installer should write it's initialization sequence to `~/.bash_profile` (that's what happens in my macOS Travis builds (see https://travis-ci.com/phil-blain/CICE-conda/jobs/275092695#L210 for example). 
However, if you then invoke `bash` from your login csh shell then `~/.bash_profile` is not read as this bash is not a login shell so it makes sense that the `conda` command is not available. If you invoke `bash --login` it should read `~/.bash_profile` and the `conda` command should be available.

---

> You also have this command in your documentation "conda config --set auto_activate_base false". Does that turn off the auto activate? I'm not sure we should recommend auto activation anyway.

This command turns off activation of the base environment on shell startup. `conda` command are still available, but the `root` environment created by the installer is not autoactivated on shell startup. Since the base environment contains a version of python possibly different from the system one, I think it's good and less disruptive for unaware users that we don't unecessarily change their default shell environment, so that's why I added this command.

>  And whether people use it or not, we should let them know they may need to manually source the conda profile file if it doesn't exist. I'm think conda should work with both bash and csh/tcsh.

`conda init --help` says:
>  Currently compatible shells are {bash, fish, powershell, tcsh, xonsh, zsh}

If people follow the installation instructions (and they did not change their default shell!) then it works. On macOS 10.15 they might have to `conda init zsh` since the miniconda installer initializes Bash by default. I think we might need to add that to the doc, and might as well make a note that if you use another login shell then you have to `conda init {shell}`.

apcraig:
I got it to build and run, but there were a few problems.  

First, this line in env.conda_macos cannot work, "source `conda info --base`/etc/profile.d/conda.csh".  You are trying to use conda, "conda info", at the same time you are trying to install conda via the source call.  conda info returns an error because conda is not known.

Second, I believe when you do "conda activate cice" in the env file as part of the cice.build, that also doesn't seem to work, although I'm not sure why.   If I source the conda profile file manually first and then run cice.build, that fails.  If I source the conda profile file manually and then source env.conda_macos then run cice.build, it works.  If I manually source the conda profile file and manually run "conda activate cice" then run the cice.build script, that works.  It seems doing "conda activate cice" as part of the cice.build script does not load the conda cice stuff permanently.

Also, I played around with the "conda init tcsh" and that didn't seem to do anything.  I think it's not clear whether the .tcshrc is sourced under csh and/or tcsh.  Of course, lots of our scripts also have #!/bin/csh -f so that may also be creating a problem.

This is what I recommend.  I suggest we remove all of the "conda" stuff from the scripts.  I think that's just in the env file.  Then I propose the documentation say that to build/run with conda, the user needs to manually execute

  source $HOME/miniconda3/etc/profile.d/conda.csh  (or conda.sh)
  conda activate cice

and that those could easily be aliased via something like

  alias loadcondacice 'source $HOME/miniconda3/etc/profile.d/conda.csh; conda activate cice'

What that does is remove all the complexity of the scripts/setup with regard to csh/tcsh/bash, whether the user has does something in their default environment to turn on conda or otherwise setup some defaults that may or may not be correct, and whether it all works together under various hardware/software versions.   This may not be the most sophisticated implementation but I think it's the simplest and easiest to understand.  Expert users can do the automation part on their own if they want and in the way that is most comfortable for them, but for new users, what I propose is probably cleaner and more robust. 

If we wanted, we could probably add something to the env script to check that the conda cice is active.  This probably works

  set chk = `conda info --envs | grep cice | wc -l`
  if ($chk < 1) then
     echo ${0}: conda env not setup, see documentation for how to activate the conda cice env 
  endif

but there are many ways it could probably be done.  The above should fail if conda is not found or if the cice env is not loaded.  It's probably worth having a check like that in any case.

Anyway, the main thing is that I have been able to build and run a cice case on my Mac Laptop with 10.14.2, but I had to manually do the conda part and that is how I suggest we implement and document usage going forward.  Excellent!  Once the scripts are updated, I will check again on my laptop and then try on my desktop.

phil-blain:
I understand what you are saying, although I really think it's cleaner to not require to activate manually. 

The method I used 
~~~
source `conda info --base`/etc/profile.d/conda.csh"
~~~
does work if the shell initialization procedure puts the conda executable in the $PATH, which is the case for Bash and zsh but not for tcsh. 

I just pushed a commit that uses a more robust method that should work for all possible login shells as long as it's correctly configured to use conda (as we show in the doc). 

I think it's reasonable to assume that the user correctly read and executed the installation instructions. I also added some documentation as to how to correctly initialize conda in different login shells.

I also added error checking to both env file, first checking if the conda executable exists, then if the "cice" conda environment can be correctly activated.

What do you think of the new instructions ?

phil-blain:
Maybe some other members of the team could also test the updated procedure ?

apcraig:
I think the documentation is better and we're almost there.  The main thing is that it seems to work which is awesome!  

I will say that overall, I do not like that we are telling people that they have to change their default environment.  I have learned the hard way that this always has unintended consequences.  If this were an HPC environment, I would say it's unacceptable.  We have the "env" file so we can setup the local environment; modules, env variables, etc, automatically for the model at hand.  My preferences for handling the environment in general is to implement by
  - set it all up in the env file
  - have the user manually initialize what's needed ahead of time either thru a set of commands or a precanned script (maybe that they write)
  - never require/expect that something is set in the default environment
  - as a last step, set something in the default environment

So again, overall, I don't think setting stuff in the default environment is the best approach.

Also, I am unclear about what is going on in the env file.  You have this line,

`  source `$CONDA_EXE info --base`/etc/profile.d/conda.csh
`

Isn't that what the default initialization is doing?  Why does it have to be sourced again?  Also, if the env file does the source, why can't the "default initialization" just be to 

`  setenv CONDA_EXE $HOME/miniconda3/bin/conda
`

if the env file is sourcing the conda env anyway, how does having this

```
if ( -f "/Users/tcraig/miniconda3/etc/profile.d/conda.csh" ) then
    source "/Users/tcraig/miniconda3/etc/profile.d/conda.csh"
else
    setenv PATH="/Users/tcraig/miniconda3/bin:$PATH"
endif

```

in your default scripts add any value?  It seems there is some redundancy that might mean the conda default stuff is not needed.  In fact, I tested that with the current implementation, I can delete the conda init stuff in my .tcshrc file, simply set the CONDA_EXE env variable manually (it could also be in my login), and everything works.  Personally, I think this is a much better way to go.  Just set one env variable and not have to set anything about conda by default.  In fact, I'd change the env file so it is just

  if ! $?CONDA_EXE then
     setenv CONDA_EXE $HOME/miniconda3/bin/conda
  endif

Then the documentation does not have to say anything about installing conda in the default environment and it can just say, "if you did not install miniconda3 under your $HOME directory, you need to set an env variable, CONDA_EXE, in order to build and run CICE on your laptop.  ...".

But assuming we are going to keep things as they are, I have a few other comments. 
  - How about moving the Running CICE on Laptop/Personal computers to a separate section under porting.  Also, I might just call it "Porting to Laptops or Personal Computers"
  - I would add a note somewhere indicating that by using conda, you can leverage the already setup conda macos or conda linux ports and they should work on your machine without having to go thru the formal machine porting process.  I think it's important to indicate the port is largely already done if you setup the conda env as documented. 
  - I would refactor the "Installing Miniconda" section a bit.  The logic is a little confusing in that you recommend initializing conda in your shell when you install miniconda, then you talk about how that doesn't work for 10.15+ and then you talk about what to do if you are using another shell.  I would reorganize as follows.  Install miniconda (for mac and linux) without mentioning initializing the default shell or the conda config step.  I would then have a separate section that says "initialize the default shell for bash, zsh, tcsh, etc".  You can mention what the default shell is as a function of OS version if you like, but you could also say if you aren't sure, do it for them all :).  I would also add a separate section that says, "if you don't want to initialize conda by default in your login, you can do the following each time prior to building and running CICE on your laptop, ....".  That should cover cases where the default initialization is not done.
  - I would make it a little clearer what steps are "one-time".  I might even separate the sections into "one-time steps" and "to build and run".  It doesn't have to be separate by section, but I think emphasis on one-time and what's not one-time is helpful.  
  - I would just add a few details about how "the conda environment is automatically activated when compiling or running the model using the ./cice.build and ./cice.run scripts".  I think it's helpful to know, at least, where these commands are and maybe what they are.

phil-blain:
Hi Tony, I'll try to answer the points you bring:

> I think the documentation is better and we're almost there. The main thing is that it seems to work which is awesome!

super!
> I will say that overall, I do not like that we are telling people that they have to change their default environment. I have learned the hard way that this always has unintended consequences. If this were an HPC environment, I would say it's unacceptable. We have the "env" file so we can setup the local environment; modules, env variables, etc, automatically for the model at hand. My preferences for handling the environment in general is to implement by
> 
> * set it all up in the env file
> * have the user manually initialize what's needed ahead of time either thru a set of commands or a precanned script (maybe that they write)
> * never require/expect that something is set in the default environment
> * as a last step, set something in the default environment
> 
> So again, overall, I don't think setting stuff in the default environment is the best approach.
> 
I do agree with most of that, especially in an HPC environment. However, I think that most people that will use this procedure on their laptop are not expert shell users, and will happily follow the installation instructions we provide. So what I propose is to support both cases:
- We advise users that they should initialize their login shell to use conda, so that `conda` is a recognized command when they open their terminal (that's what the miniconda installation procedure recommends)
- We also mention what steps are needed for more advanced users that do not want to modify their shell startup scripts.

---

> Also, I am unclear about what is going on in the env file. You have this line,
> ~~~
>  source`$CONDA_EXE info --base`/etc/profile.d/conda.csh
> ~~~
> Isn't that what the default initialization is doing? Why does it have to be sourced again?

The conda shell initialization procedure (`conda init <shell>`, by itself or as part of the miniconda installation by answering "yes") makes sure that users can run "conda" in an interactive shell `<shell>`. 
The way `conda init` does that is by writing some initialization sequence to the shell startup file (`.bashrc`, `.tcshrc`, `.zshrc`, `.bash_profile` on macOS, etc.). This initialization sequence is shell-dependent. 
For Bash and Zsh, one of the thing it does is put the path to the `conda` Python executable in the $PATH. In tcsh, it just adds a `conda` alias that sources `path/to/miniconda/etc/profile.d/conda.csh`. Another thing it does  (for all shells) is set the `$CONDA_EXE` environment variable to the path to the `conda` executable.

The file `path/to/miniconda/etc/profile.d/conda.csh` needs to be sourced inside the CICE env.conda_{macos,linux} file for `conda activate cice` to work, because the shell initialization procedure is shell-dependent. What's more, this is the case even if this parent shell is tcsh since `cice.build` and `cice.run` both have a `#! /bin/csh -f` shebang, and from what I understand the `-f` flag makes the shell (the one that runs these scripts) *not* inherit any alias set by the parent shell. It *does* inherit environment variable though, so that is why using `$CONDA_EXE` works.

We could do
~~~diff
diff --git a/configuration/scripts/machines/env.conda_linux b/configuration/scripts/machines/env.conda_linux
index 9186a160..c54b9044 100755
--- a/configuration/scripts/machines/env.conda_linux
+++ b/configuration/scripts/machines/env.conda_linux
@@ -8,7 +8,7 @@ endif
 if ("$inp" != "-nomodules") then
 
 # Init conda

-source `$CONDA_EXE info --base`/etc/profile.d/conda.csh
+source $HOME/miniconda3/etc/profile.d/conda.csh
 # Activate "cice" conda environment
 conda activate cice
~~~
instead but then if a user wants to install Miniconda somewhere else than in their $HOME, our env file would not work. So it is more versatile this way, we do not impose where Miniconda should be installed.

---

> But assuming we are going to keep things as they are, I have a few other comments.
> 
> * How about moving the Running CICE on Laptop/Personal computers to a separate section under porting.  Also, I might just call it "Porting to Laptops or Personal Computers"
> * I would add a note somewhere indicating that by using conda, you can leverage the already setup conda macos or conda linux ports and they should work on your machine without having to go thru the formal machine porting process.  I think it's important to indicate the port is largely already done if you setup the conda env as documented.

Yes I think you're right. it would make more sense as section 3.2.3.5. 

> * I would refactor the "Installing Miniconda" section a bit.  The logic is a little confusing in that you recommend initializing conda in your shell when you install miniconda, then you talk about how that doesn't work for 10.15+ and then you talk about what to do if you are using another shell.  I would reorganize as follows.  Install miniconda (for mac and linux) without mentioning initializing the default shell or the conda config step.  I would then have a separate section that says "initialize the default shell for bash, zsh, tcsh, etc".  You can mention what the default shell is as a function of OS version if you like, but you could also say if you aren't sure, do it for them all :).  I would also add a separate section that says, "if you don't want to initialize conda by default in your login, you can do the following each time prior to building and running CICE on your laptop, ....".  That should cover cases where the default initialization is not done.

yes. I think that's a good way forward. I'll update the documentation.

> * I would make it a little clearer what steps are "one-time".  I might even separate the sections into "one-time steps" and "to build and run".  It doesn't have to be separate by section, but I think emphasis on one-time and what's not one-time is helpful.

good idea. I'll review that.
> * I would just add a few details about how "the conda environment is automatically activated when compiling or running the model using the ./cice.build and ./cice.run scripts".  I think it's helpful to know, at least, where these commands are and maybe what they are.

OK I'll can  a few words about that.

phil-blain:
I implemented the changes proposed above in the documentation.

I just noticed I added a new section "Porting to Laptop or Personal Computers" **after** "Porting" instead of **under** it. However I think it's maybe  more useful that way as it limits the number of subsubsections.

I could also change it.

eclare108213:
@phil-blain this is AWESOME!  I followed the directions in the documentation and managed to get it to work, stumbling over a few things.  My only complaint about this is that the instructions force the input data to be in ~/cice-dirs.  I'd rather not have cice-dirs in my home directory, and so I tried to put it where I want it, but that didn't work, maybe because I'd set up conda from my home directory (?) or maybe because the locations are hardwired somewhere in the scripts.  Now that it's there, am I stuck with it?

I'm not sure that these need to be documented, but for the record, here are the things that gave me problems (macOS 10.14.6):
1.  > curl: (7) Failed to connect to repo.anaconda.com port 443: Connection refused

So I downloaded it manually.

2.  > EnvironmentFileNotFound: '/Users/eclare/configuration/scripts/machines/environment.yml' file not found

This was because I hadn't cloned your branch yet...   :)
This is only going to be possible for CICE versions after this PR is merged.

3. 
> (base) pn1807697:cice_conda-env eclare$ conda env create -f configuration/scripts/machines/environment.yml
> Collecting package metadata (repodata.json): failed

I suspended my VPN session and this then worked.  This was likely the problem with 1 above.

4.  the input data wasn't where it was expected to be, so I moved it there

The only thing I suggest is that the conda environment and/or cice-dirs be installed where the user wants them.  For other machines, there are standard names for the directories (CICE_RUNS, CICE_BASELINE or something like that, I think), and the user can choose where to put them by modifying the machine files.  Since there are a number of "do one time" instructions for the conda environment, I wonder if the user should be given some of these choices early on?  It would be good to be consistent, anyhow.

THANK YOU.  This is fantastic.

apcraig:
I'll just add a couple thoughts based on @eclare108213 comments.  The ability to set your own input data, baseline, and run directory is an important one.  But what it means is that each user would have to create their own machine port.  What @phil-blain has done is create a single "conda" port which requires leveraging the paths set there.  We do not want users to customize the conda port and then push back to master as that would break the port for others.

But there are a couple options.  First, we could suggest users can actually do a machine port, taking the conda port as a starting point and then creating a new set of machine files.  The second option would be to let people know they can create links from the hardwired conda paths to wherever they want.  So, if you want to put the input data somewhere else, just use the conda port and manually create a permanent link from $HOME/cice_dirs/cice_input (or whatever it's called) to $WORK/cice_input (or wherever the data actually exists on your computer).  I think the link is a reasonable workaround in this case and it should be "one-time".  But that's not to say a machine port isn't also an option folks can do as well or instead.  I think we probably just need to add a little documentation about both scenarios.

phil-blain:
I agree that we need to strike a balance between "works out-of-the-box" and "I  don't want this `cice-dirs` directory in my home."

Personally, I would not want them there either :P But as @apcraig says then the user would have to modify the env file.

@apcraig I agree that at least mentioning that creating symlinks is possible is a good idea.

We can also mention that if they want their paths elsewhere they can port to their own machine by customizing the env file based on the conda one. That's also a good idea.

Another thing we could do is modify the env file so that all paths are defined using a `CICE_HOME` (or similar) environment variable *if this variable is set.*
Then just as we tell users that they can manually source the conda init script if they don't want to modify their environment, we could tell them to export this variable with the correct path to where they want to put the CICE directories.

@apcraig @eclare108213 what do you think of that approach ?

apcraig:
The CICE_HOME env variable is an interesting one.  I worry a little it could drive that same implementation into other machine files or introduce some confusion when users switch between their mac and other machines.  So my initial instinct is to not implement it.

But it does raise some interesting questions about whether we could move away from machine ports and move to a more env driven configuration system where the machine dependent differences are addressed by defining some local env variables and largely reusing the same env, Macros, launch and batch scripts across machines.  I'm not convinced it's viable or better.  But an interesting idea to ponder moving forward.

eclare108213:
I wondered the same thing, @apcraig.  Would this procedure also work on something bigger than a desktop, like a linux cluster with a few hundred processors?  It's a headache to have to rely on admins to keep the software on some bigger, local machines up to date (especially netCDF).

phil-blain:
@eclare108213 it would work in theory but for optimal performance the MPI implementation must be the one from the machine vendor, or OpenMPI/MPICH must be compiled on the machine itself... 

and with the conda environment there is not job scheduler so all runs are interactive.

Altough it might be possible to use a similar approach for the env files for that cluser to install say just NetCDF through conda... but then there's the complication that each user must have access to the conda environment in their interactive and batch sessions... so it's not easy

eclare108213:
I see.  This is very cool for laptops, anyhow!  A number of years ago, someone sent me (or pointed me to) a video of CICE output scrolling on their hotel TV screen, being run from a raspberry pi...  I thought that was hilarious.  Pre iPhone days.  I guess someone will make an app sooner or later.

phil-blain:
@apcraig @eclare108213 I added a note about symlinks and also about creating their own port.

I think this is ready for final review now.

phil-blain:
the RTD build is failing with some cryptic Python error... probably a change of config on their end. I'll try to re-trigger it later.

apcraig:
RTD was failing intermittently last week too.  Just force push or retrigger when you think it's time.

apcraig:
We need to trigger RTD again, still not working.  Last week it took me a couple tries on one of the PRs too.  Odd because other PRs are working.  Saw this last week too.  I'm pretty sure it's not the PR but RTD.  It would be good to figure out how to retrigger RTD without a push.  I know how to do it for travis but not RTD.

phil-blain:
@apcraig  @eclare108213 RTD worked. Formatted documentation is here:
https://external-builds.readthedocs.io/html/cice-consortium-cice/393/user_guide/ug_running.html#porting-to-laptop-or-personal-computers

For the record a simple way to trigger a new build is just to amend the last commit in the branch and force push:
~~~bash
git checkout <branch>
git commit --amend --no-edit
git push --force
~~~

this doesn't change anything in the last commit apart from the commiter date.

apcraig:
Documentation looks good.  Thanks @phil-blain, this is an awesome capability!  Once travis is done, I think we can merge.

apcraig:
One other thing, it would be nice if we could retrigger RTD without having to push.  We can do that with Travis (you just go into travis and click the retrigger build link).  Have looked around RTD, but have not found a retrigger capability.  The requirement to push means only users with write access to the development branch can retrigger which is not ideal.  If anyone has any ideas how to deal with this, lets discuss (maybe we can move to email).

phil-blain:
@apcraig I googled the python error and found this:
https://stackoverflow.com/questions/59846065/read-the-docs-build-fails-with-cannot-import-name-packagefinder-from-pip-in

They suggest wiping the build environment in RTD:
https://docs.readthedocs.io/en/stable/guides/wipe-environment.html

Someone with admin rights need to do that. It might not stick though until this [RTD PR](https://github.com/readthedocs/readthedocs.org/pull/6571) is merged, according to [this comment](https://github.com/readthedocs/readthedocs.org/issues/6554#issuecomment-577741186)

apcraig:
OK, I wiped the master and latest versions.  I don't know if it'll help with the current PRs and there is no version associated with them that I can wipe.  Hopefully this will stop being a problem as we generate new PRs, we'll see.

PR #395, opened 2020-01-15
update suite.submit script and add new controls and cice.setup option…

apcraig:
## PR checklist
    Update suite.submit script and add new features.
- [X] Developer(s): 
    apcraig
    Tested on izumi, defaults settings work fine and are backward compatible.
    Tested the suite.submit manually as well.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Added cice.setup options

-   --setup-only
-   --build-only
-   --run-interactive

Added ability to set env variables for use with suite.submit directory

-   SUITE_BUILD
-   SUITE_RUN
-   SUITE_SUBMIT

Removed suite.run script, this is now part of suite.submit script.

Also updated the documentation.  This is an update of #357 based on the latest code and discussions in that PR.

### DIFF
--- a/cice.setup
+++ b/cice.setup
@@ -34,6 +34,9 @@ set stime = `date -u "+%H%M%S"`
 set docase = 0
 set dotest = 0
 set dosuite = 0
+set suitebuild  = true
+set suiterun    = false
+set suitesubmit = true
 
 if ($#argv < 1) then
   set helpheader = 1
@@ -79,6 +82,7 @@ SYNOPSIS
     --suite SUITE[,SUITE2] -m MACH --testid ID 
         [-e ENV1,ENV2][--acct ACCT][--bdir DIR][--bgen DIR]
         [--bcmp DIR][--tdir PATH][--report]
+        [--setup-only][--build-only][--run-interactive]
 
 DESCRIPTION
     --help, -h : help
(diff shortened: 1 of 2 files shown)


apcraig:
I am closing #357.

apcraig:
I'm open about how to name the options.  I think build-only implicitly means you have to create the cases, but understand these things can be tricky.  Also, maybe we need 4 options.  For clarity, how about

  - setup-only
  - setup-and-build-only
  - setup-and-build-and-run-interactive
  - setup-and-build-and-submit

That seems a little too wordy but is clear.  Other proposals?

eclare108213:
How about 
 * setup
 * setup-build
 * setup-build-run
 * setup-build-submit

apcraig:
How about 

 - setup-only
 - setup-build
 - setup-build-run
 - setup-build-submit

apcraig:
I updated the options and committed.  Can still make another change if needed.

phil-blain:
I'm still looking at the tutorial activities and will have more time to test this next week...

apcraig:
Just a quick update that we are waiting for @phil-blain to do a test/review.  There is no rush or urgency, but that's why this PR is on hold for the moment.

apcraig:
I just made some commits to update the documentation as suggested by @phil-blain.  I think this improves things and have confirmed the formatting in the RTD output.  

With regard to the csh vs bash syntax, my preference is to avoid providing multiple examples for multiple shells if it's something pretty basic, so instead, I have added a quick comment about that.  I think for the conda stuff, given the install is different in different shells, then it should be documented.  For this, where we're just setting an env variable, I think we should be able to rely on the community to understand that difference.

I think we can probably merge before the weekend testing unless there are other comments or feedback.

phil-blain:
looks good to me!

Issue #344, opened 2019-08-07
setting diag_type = 'file' in namelist makes cice.run end with 'CICE run did not complete'


phil-blain:
Not a big issue, but still it would be nice if cice.run could grep the diag file specified in the namelist (diag_file) if diag_type is set to 'file' instead of the default value 'stdout'. Then real runtime errors would be spotted faster when diag_type = 'file' is used.

apcraig:
I just had a look at this and don't think there's an easy or general fix.  We could possibly detect the diag_type='file' setting at runtime and then figure out the output filename, but there are a whole cascade of other issues like

  - does it get saved to the case logfile
  - does it get saved to the test suite baseline directory
  - do the baseline comparisons work properly.  can the scripts figure out all the filenames for any case ever run.

My recommendation is that we not try to fix this at this time.  What we could do, if others felt it was important, would be to detect the diag_type='file' setting and then just turn off the successfully completed check and add a note that users are on their own.  Is that something worth doing?

phil-blain:
I think that's sound. Also, just having this issue open can serve as rudimentary documentation if anyone stumbles onto this and ends up here.

phil-blain:
@apcraig we can close this issue as it was fixed in #396 right ?

apcraig:
It was fixed in part in #396.  The successful completion is found, but the rest of the testing will not work.  I agree it should be closed.  We don't normally use diag_type=file.  I think we're OK for now.

PR #396, opened 2020-01-17
Macros fixes and diag_type fix

apcraig:
## PR checklist
    Fix Macros files, threading was not active on a few platforms including cheyenne.
    Update run status output based on diag_type setting
- [X] Developer(s): 
    apcraig
    Full test suite on cheyenne, will post results when done
    - [ ] bit for bit
    - [X] more substantial 
        For threaded tests, answers change on some platforms, namely cheyenne
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Surprised we didn't find this error sooner!
Changes answers for some threaded cases.  Machines affected are

  - cheyenne
  - cesium
  - fram
  - millikan
  - phase2
  - phase3

Other Macros changes don't affect answers.
Partly address CICE #344


apcraig:
I will post test results when complete.  I have already run the suite a few times on cheyenne, just doing a final test.  I expect everything to pass but answers to change for threaded cases.  Most machine unaffected.

I will also merge this on Friday afternoon.  This is a significant bug that we need to fix asap, especially with the user meeting coming up and needing to work on cheyenne.  I want this in the weekend testing.

apcraig:
Test results are in
[https://github.com/CICE-Consortium/Test-Results/wiki/5bbe517b77.cheyenne.intel.20-01-17.012453](https://github.com/CICE-Consortium/Test-Results/wiki/5bbe517b77.cheyenne.intel.20-01-17.012453)
Turning on threading sure does destroy performance on cheyenne.  We may need to rethink the cheyenne test suite and just run the nothread_suite.  Pretty surprising.  That's something we can think about after this PR.

apcraig:
I just updated the documentation.  I think it's as suggested.

configuration/scripts/cice.run.setup.csh
```diff
@@ -73,10 +75,18 @@ echo " "
...
+if ( \${diagtype} > 0) then
+  grep ' CICE COMPLETED SUCCESSFULLY' \${ICE_RUNLOG_FILE}
+  if ( \$status != 0 ) then
+    echo "CICE run did not complete - see \${ICE_LOGDIR}/\${ICE_RUNLOG_FILE}"
+    echo "\`date\` \${0}: \${ICE_CASENAME} run did NOT complete \${ICE_RUNLOG_FILE}"  >> \${ICE_CASEDIR}/README.case
+    exit -1
+  endif
+else
+  echo "\`date\` \${0}: \${ICE_CASENAME} run completed "
+  echo "CICE NOT run with diag_type='stdout', run status unknown"
```
phil-blain:
```suggestion
  echo "CICE NOT run with diag_type='stdout', run status unknown - see \${ICE_LOGDIR}/\${ICE_RUNLOG_FILE}"
```

I think in that case it would be nice to also output the path to the log file for easy (\<middle click path\>  + `vi <middle click>`) access

apcraig:
Thanks @phil-blain, several good ideas there.  I'll try to clean up some of that.

One thing is that if the diag_type is not set to stdout, it really needs to return a negative exit code.  We could grab the filename from ice_in and check whether the run successfully completed, but all the rest of the stuff that is part of the test script cannot work unless diag_type is stdout.  That could be fixed too, but that would be a bunch more work.  Plus I don't see the need.  All testing can be done with "stdout".  I would also argue that using diag_type='file' is not a great idea in general.  As you point out, it will continually overwrite (or append) the prior output file.  The current implementation, leveraging a date stamp in the ice log file set by the scripts ensures each output file is a unique name which I think is a nice feature.

Finally, I think I figured out the performance issue with threading on cheyenne.  We need omplace on job launch.  I'm testing that too.  I should have some updated commits soon.

phil-blain:
Yes I agree with all these considerations. That's why I don't feel too strongly about the diag_file one; I  only use it for cases and not test/suites. 
I don't see the need either to adapt the testing architecture. 

So I'm glad with the changes you're proposing in this PR, plus outputting the path to the log file in the diag_file case.

apcraig:
Just pushed an update.  cheyenne timing is much better now.  [https://github.com/CICE-Consortium/Test-Results/wiki/af3d93e40c.cheyenne.intel.20-01-17.183453](https://github.com/CICE-Consortium/Test-Results/wiki/af3d93e40c.cheyenne.intel.20-01-17.183453).  All tests pass and most are faster.  Many regression comparisons fail (as expected).

Also cleaned up the scripts a bit and update the run checking as guided above.

Would like to merge the PR today if possible.

apcraig:
@eclare108213, usage doesn't change.  I have tested the updated scripts with diag_type = stdout and diag_type = file and it seems to work as designed.

apcraig:
Not sure why readthedocs is failing right now.  Have tried to run it twice.  It was fine before and my recent changes did not affect documentation.  I think there is a problem with readthedocs, will have to see whether it's temporary or something that we'll need to fix.  Here is the version from #2078b884 which was after my doc change, [https://readthedocs.org/projects/cice-consortium-cice/builds/10275392/](https://readthedocs.org/projects/cice-consortium-cice/builds/10275392/)

apcraig:
Read the docs is working again.  I am going to merge this PR now.

Issue #405, opened 2020-02-17
cice.setup broken for cases and tests since PR #395


phil-blain:
Since PR #395 was merged it appears that `cice.setup` can't create cases or individual tests, in both cases I get a syntax error in the `cice.setup` script:

~~~shelll
$ ./cice.setup --test smoke --mach cesium  -s diag1,debug --testid myid
 
./cice.setup:
else: endif not found.
~~~

I did not have time to find where is the error yet, just letting you know @apcraig.

We should add a call to `cice.setup --case` and `cice.setup --test` in the `.travis.yml` such that breakages like this do not go unnoticed in the future.

phil-blain:
I guess this would be high or extreme priority...

apcraig:
Thanks @phil-blain, I'll have a look and get a fix in this evening.

apcraig:
I just committed the fix #406 and added some tests to travis that would trap this in the future.  Sorry about the error, I guess this'll happen occasionally.  Glad we caught it fairly quickly.  I will close this.

PR #406, opened 2020-02-18
Fix errors in cice.setup script and add travis tests to catch this sooner

apcraig:
    fix error in cice.setup script introduced in #395 and documented in #405
- [X] Developer(s): 
    apcraig
  Tested case and test on izumi.  Also added test to travisCI.  Ran a quick suite on izumi,
   https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#b2ca7858f35d20b2ba00a1e73d3ddfe2299c9f6f
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

I am first testing if the new travis tests will catch this then I am going to implement the fix.

I verified that the new travis tests would have failed in this case.  

I have also fixed the error.

### DIFF
--- a/cice.setup
+++ b/cice.setup
@@ -389,13 +389,11 @@ else
       cat ${tarray} >> $tsfile
     else if (-e ${tarray}.ts) then
       cat ${tarray}.ts >> $tsfile
-    else 
-      if (-e ${ICE_SCRIPTS}/tests/${tarray}.ts) then
-        cat ${ICE_SCRIPTS}/tests/${tarray}.ts >> $tsfile
-      else
-        echo "${0}: ERROR, cannot find testsuite file ${tarray}, also checked ${ICE_SCRIPTS}/tests/${tarray}.ts"
-        exit -1
-      endif
+    else if (-e ${ICE_SCRIPTS}/tests/${tarray}.ts) then
+      cat ${ICE_SCRIPTS}/tests/${tarray}.ts >> $tsfile
+    else
+      echo "${0}: ERROR, cannot find testsuite file ${tarray}, also checked ${ICE_SCRIPTS}/tests/${tarray}.ts"
+      exit -1
     endif
   end
 
@@ -422,6 +420,7 @@ set dorun    = false
 set dosubmit = true
 if (\$?SUITE_BUILD) then
   set dobuild  = "\${SUITE_BUILD}"
+endif
 if (\$?SUITE_RUN) then
   set dorun    = "\${SUITE_RUN}"
 endif
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,10 @@ install:
   #-e 'mirror /archive/Model-Data/CICE/ ~/ICEPACK_INPUTDATA; quit'"
 
 script:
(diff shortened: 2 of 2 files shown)


apcraig:
This is ready to merge.  I fixed the error and also added some tests to travis to check that cice.setup --case and cice.setup --test complete successfully (but I don't build or run those case, it's just a scripts check).  I also verified that travis would have failed with the new tests.

Since this is critical, I will merge this now.  If there are still problems, we can execute another PR later.

cice.setup
```diff
@@ -389,13 +389,11 @@ else
...
-        cat ${ICE_SCRIPTS}/tests/${tarray}.ts >> $tsfile
-      else
-        echo "${0}: ERROR, cannot find testsuite file ${tarray}, also checked ${ICE_SCRIPTS}/tests/${tarray}.ts"
-        exit -1
-      endif
+    else if (-e ${ICE_SCRIPTS}/tests/${tarray}.ts) then
+      cat ${ICE_SCRIPTS}/tests/${tarray}.ts >> $tsfile
+    else
+      echo "${0}: ERROR, cannot find testsuite file ${tarray}, also checked ${ICE_SCRIPTS}/tests/${tarray}.ts"
+      exit -1
```
phil-blain: this just simplifies the logic, right ?

cice.setup
```diff
@@ -422,6 +420,7 @@ set dorun    = false
 set dosubmit = true
 if (\$?SUITE_BUILD) then
   set dobuild  = "\${SUITE_BUILD}"
+endif
```
phil-blain: so from what I understand and tested this forgotten `endif` was the cause of the 

    else: endif not found.

error. I am pretty surprised, since this is inside the heredoc, so it is just some text written to a file, still the shell running `cice.setup` dies with a syntax error...

cice.setup
apcraig: correct.  noticed this as I was debugging.

cice.setup
apcraig: exactly.  this is pretty shocking.  The endif is missing in the job.submit script yet that script never complained.  This is exactly what I had been testing all along.  Then when we run a non-suite with cice.setup, it failed due to the missing endif in the part of the script that was generating the job.submit suite script.  I really cannot believe how this bug was manifested in the end.

eclare108213: This kind of weird behavior usually means there's some other bug, doesn't it?  Is it possible to step through the script to find out why it hit that block of code?  Or is this a "compiler" bug?

phil-blain: I think this weird behavior is due to the peculiarities of the C shell. As far as I know it is not possible to step a shell script...

So yes I think this is could be classified as a bug in the C shell.

daveh150:
FWIW, I had to manually fix this in the basalstress tests I just ran (I did not commit this for the pull request). Otherwise cice.setup crashed for me. How it worked before is a good question. 

 

David

 

David A. Hebert

Naval Research Laboratory 
(228)688-5846; DSN 828-5846 

PR #410, opened 2020-02-27
Iceopen

daveh150:

## PR checklist
- [x ] Short (1 sentence) summary of your PR: 
    Update ice_open to use 8 byte integers for large global grids.
- [x ] Developer(s): 
    David Hebert, Naval Research Lab, Stennis Space Center, MS
- [x ] Suggest PR reviewers from list in the column to the right.
    none shown
- [x ] Please copy the PR test results link or provide a summary of testing completed below.
[results.log](https://github.com/CICE-Consortium/CICE/files/4264060/results.log)

    - [x ] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x ] No
    - [ ] Yes
    - [ x] No
    - [ ] Yes
    - [x ] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [x ] No 


daveh150:
Tony,

 

I think your solution is good, given how CICE6 opens and reads files internally. The only other way I can think to get around this is to loop over the ‘y’ and read into the array, but that is not compatible with how open and read are used throughout the code. “Stream” access may work, but then we need to worry about byte swapping. 

 

I’ll implement what you have outlined and run the tests again. That should work. 

 

Thanks,
Dave

 

David A. Hebert

Naval Research Laboratory 
(228)688-5846; DSN 828-5846 
 <http://www.nrl.navy.mil/> http://www.nrl.navy.mil

daveh150:
Actually, now that I think a bit more, if we want to avoid integer8 altogether, then I think the original method I submitted would solve that. But a question for the group is what do you prefer from a ‘clean code’ standpoint. So for reference I had (adding bits_per_byte):

 

integer(kind=int_kind) :: nbytes

integer (kind=int_kind),parameter :: bits_per_byte = 8 ! this can be module data or even added to ice_constants

 

nbytes = nbits / bits_per_byte

RecSize = nxglobal * nyglobal * nbytes

 

Both methods will work, and to me the above is a bit cleaner from coding (with comments). But for documentation, if it is better to use what Tony suggested I’m good with that. 

 

Let me know what y’all think!

 

Thanks,
Dave

 

 

David A. Hebert

Naval Research Laboratory 
(228)688-5846; DSN 828-5846 

cicecore/cicedynB/infrastructure/ice_read_write.F90
```diff
@@ -78,18 +78,28 @@ module ice_read_write
 ! nbits indicates whether the file is sequential or direct access.
 !
 ! author: Tony Craig, NCAR
+! 
+! update: David Hebert, NRLSSC, Feb 2020
+!         added nx8, ny8, RecSize defined as 8 byte integers
+!         to allow for large, high resolution gloabl grids.
```
eclare108213: spell global

daveh150:
I just committed the method by computing nbytes first, then RecSize = nx*ny*nbytes. To me, typecasting the 8 byte integers and then converting back to 4 byte seemed too complicated at this stage, and didn't really do anything extra since we don't use the RecSizeI8 variable after the data is computed. Please let me know if  it is preferred to go back to the using RecSizeI8. 

Thank  you!
David

apcraig:
I'm fine with this @daveh150.  But how about making the bits_per_byte module data for now.  We also need to retrigger the PR because readthedocs gagged, so a quick push should do that.  Have you run any tests with this update?  Thanks!

daveh150:
Attached is the results.log with the 'nbytes' method. So far all passed, I'm waiting for 2 in the queue.

I moved the bits_per_byte to a module level variables. Those tests are underway now. I will send log when complete. 
[results.log](https://github.com/CICE-Consortium/CICE/files/4269059/results.log)

apcraig:
This looks good.  I think we can merge it when testing is done.  I would like to merge this today after the JRA55 PR.  My only final comment is that if nbits/bits_per_byte does not divide evenly, the new computation could produce a different result.  That is partly why I was suggesting integer*8, but I don't think we'll run into that issue.  

Keep us posted on test results.  I hope to include this in the weekend testing.

daveh150:
Interesting, hadn’t thought if nbits is not divisible by 8. Are we aware if anyone is using a nbits that is not divisible by 8? I guess the same problem would occur in the original method, since it was divided by 8? The integer math would not result in an accurate record length?

 

Thanks again. Jobs are in the queue now. Gordon queue is slow at the moment. Hopefully by the weekend I’ll have results. 


Dave

 

David A. Hebert

Naval Research Laboratory 
(228)688-5846; DSN 828-5846 

phil-blain:
@apcraig for this PR it would have been useful in my opinion to edit the commit message generated by the "Squash and Merge" button, because now the commit message mentions several commits that were not the subject of this PR, but were included because the branch `iceopen` had merge commits coming from merging the consortium master (and probably Dave's master also ? )...

PR #418, opened 2020-03-17
conda: Macros: add '-L' argument to the linker command

phil-blain:
## PR checklist
  add '-L' argument to the linker command under conda
- [X] Developer(s): 
    P. Blain
- [x] Suggest PR reviewers from list in the column to the right. @apcraig 
    No test suite run. Tested this on my Ubuntu 17.04 system.
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 

On Linux, compiling CICE in serial mode under the conda environment
currently fails as the linker does not find the NetCDF libraries.

Compiling in parallel mode does work as the MPI compilation wrappers
seem to pass the necessary flag to the linker under the hood.

Explicitely tell the linker the location of the libraries using the
`SLIBS` variable in the Macros file. To be on the safe side, also do it
on macOS.

### DIFF
--- a/configuration/scripts/machines/Macros.conda_linux
+++ b/configuration/scripts/machines/Macros.conda_linux
@@ -40,7 +40,7 @@ LD:= $(FC)
 MODDIR  += -I$(CONDA_PREFIX)/include
 
 # Libraries to be passed to the linker
-SLIBS   := -lnetcdf -lnetcdff
+SLIBS   := -L$(CONDA_PREFIX)/lib -lnetcdf -lnetcdff
 
 # Necessary flag to compile with OpenMP support
 ifeq ($(ICE_THREADED), true)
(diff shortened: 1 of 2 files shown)


phil-blain:
@apcraig since I now have "triage" permission I can add labels and directly ask for reviews, super!

phil-blain:
Actually this is kind of a workaround, as it would work out of the box if our script were not using Tcsh: if you take a look at `/path/to/miniconda/envs/cice/etc/conda/activate.d/` you will see that the conda compiler packages have activation scripts that set the appropriate standard environment variables (FC, LD, FFLAGS, LDFLAGS) when the environment is activated, but these scripts are just for Bash and do not have a Csh version.

This means that if we do `conda activate cice && ./cice.build` (instead of relying on the environment activation being done in the env file) it will work.

I'll see if I can open an issue on the repo for the conda-forge compilers to ask if they would be willing to add activation scripts for Csh (one can always ask !).

phil-blain:
asked here: https://github.com/conda-forge/compilers-feedstock/issues/20#issuecomment-600666554

Issue #238, opened 2018-11-15
tracer index improvements


apcraig:
There are several tasks that should be considered moving forward.  This is related to #115 and #212. 

1. test arbitrary tracer index order for bit-for-bit.  understand tracer dependencies and order requirements better.  allow tracers to be indexed in any order as much as possible.
2. get rid of the "unused" tracer and set n_[tracer] to 0 (or -1) for tracers that are inactive.  make sure that works (see count_tracers)
3. get rid of the icepack_max parameters in CICE to define array sizes, use sizes computed at runtime.  may require upgrades to both icepack and cice to allow for dynamic sized arrays to be passed between cice and icepack.  might require addressing non-assumed shape/size arrays in interfaces.
4. make sure arguments to subroutines passed in as array sections are working ok for non-active tracers with invalid or arbitrary index (related to 2)
5. make subroutine argument sizing more robust and extensible, move to assumed size/shape where needed, make sure if an array size is passed as an argument that the implementation is correct, avoid mixing dynamic and static arrays on both sides of a subroutine interface.
6. review icepack_tracer routine where tracer index and count are determined from max values.  this is inconsistent with dynamics tracer counts in cice.

eclare108213:
During the isotopes implementation #423 I noticed a few tracer indices needing to be cleaned up when their respective capability is turned off.  In particular, ntrcr=the total number of tracers+1 (item 2 above), and these tracer indices are set to ntrcr:  nt_aero, nt_bgc_S, nt_fsd, nt_fbri (there are probably others).  nt_zbgc_frac is (still?) not initialized properly and gets set to some arbitrarily large integer rather than 0 or ntrcr.

PR #423, opened 2020-03-26
Isotopes for CICE

eclare108213:
This PR is not yet ready to be merged - further review, testing and documentation is required.

Addition of isotope diagnostics following @dabail10's implementation for CESM.
- [x] Developer(s): 
@dabail10 @eclare108213 
extensive testing is still needed
    - [ ] bit for bit
    - [ ] more substantial 
    - [x] Yes
    - [ ] No
    - [x] Yes
    - [ ] No
    - [ ] Yes
        - [x] Yes
        - [ ] No 

- Icepack is not updated in this PR.  To run this code, grab the associated Icepack code from
https://github.com/CICE-Consortium/Icepack/pull/307
- This initial version compiles and runs.  Isotope restarts are failing.  It's still a work in progress, but close enough for review and regression testing with isotopes turned off.
- The implementation of the isotope optional arguments in CICE's Icepack interface are probably not consistent and need to be checked and fixed.
- One thing that I do not like about this implementation is that the ccsm aerosols, and therefore now the isotopes which mimicked the aero code, are spread out between the bgc code and the 'regular' physics code.  E.g. diagnostics are calculated in ice_diagnostics.F90 (not ice_diagnostics_bgc.F90) while history is written in ice_history_bgc.F90 (not ice_history.F90), and there appear to be similar issues for ice_flux.F90 and ice_flux_bgc.F90.  This needs to be cleaned up, and my opinion is that both aero and iso should be in the bgc code.


eclare108213:
For @dabail10 and @apcraig, an update on the current state of my tests.

CICE v6.1.1 baseline on badger:
https://github.com/CICE-Consortium/Test-Results/wiki/9040fe2be7.badger.intel.20-03-29.114803
The 40-processor tests failed, as usual.  The gx3 JRA tests failed because I don’t yet have the 2008 data on badger.  None of these failures are concerning.

CICE v6.1.1 master with my Icepack iiso2 branch:
https://github.com/CICE-Consortium/Test-Results/wiki/9040fe2be7.badger.intel.20-03-29.122009
There are a lot of regression failures, so I've not done something correctly.  It could be the interfaces, but it could also be the tracer indexing.  All of the runs completed successfully (except for the same failures above) and all of the debug tests passed, so I'm guessing there's a non-BFB initialization problem.  If either of you notice other patterns in the failures, let me know!

apcraig:
I will duplicate some of the CICE testing on conrad and try to do some debugging of the regression failures.  Right now, I'm running Icepack, but I should be able to do the full CICE test suites tonight then have a closer look tomorrow.

eclare108213:
THANK YOU

apcraig:
Test results on conrad for 4 compilers are [here]( https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#f831086a74f9324b229c216274531626d4585fdc) (#f831086a).  My results are similar on intel, more than half the regression tests fail with intel.  But with pgi, gnu, and cray compilers, those all pass.  In addition, all intel tests with debug turned on pass regression.  That has no compiler optimization.  My feeling is the failing intel regression tests are due to compiler optimization related to the code changes.  We are seeing the [same behavior in Icepack](https://github.com/CICE-Consortium/Icepack/pull/307).  We could look into this further.  I could try running a base and new test suite changing -O2 to -O1.  Or I could try other optimization flags to see if I can get bit-for-bit regression passing, but we probably wouldn't want to leave the optimization that way.  Instead, I propose we just run a qc test with the intel compiler vs a baseline to verify the climate is the same.  I will setup and run that qc test next.  Regardless of the qc results, we still can decide to do further analysis on the failed regressions tests if we wanted.

There is one failure that is problematic.  The isotope restart test is failing.  It runs fine, but it doesn't restart exactly.  Has the restart functionality for isotope been debugged?

eclare108213:
Thanks @apcraig.  It's worrying that the compiler optimization would have such troubles, but I agree that it's probably not worth tracking down at this point.  The QC testing is a great idea.

I was already aware of the restart problem and am planning to do some debugging for that.  With this current batch of tests I was hoping to get a bead on how much more needs to be done beyond that, e.g. to the icepack interfaces.  It sounds like the code is in good shape from a technical standpoint and is now ready for @dabail10 to check the physics.  Cool.

apcraig:
I just ran the current CICE master with the iiso2 Icepack version with the intel compiler and everything ran fine, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#9ac1863e6b346a954d74110474e3b58d0f2345e4.  That means the interface changes are backwards compatible as we previously tested.  The same regression tests failed as with the CICE iso2 version.  That suggests whatever non bit-for-bit results we're getting are likely coming from Icepack changes, not CICE changes.  I may look into that a bit further.

apcraig:
The qc test between master and iso2 passed on conrad with the intel compiler.

eclare108213:
The restart tests are failing in the horizontal transport, which means that the tracer indexing for isotopes is not correct.  Probably a loop over n_iso is missing or the indices within it are incorrectly implemented, but I'm not seeing it at the moment.  Will continue looking.

eclare108213:
The isotope restarts seem to be working now, so this code is ready for @dabail10 to test.  You'll have to grab Icepack in the iiso2 branch in my fork, since it's not linked here.

@apcraig One thing I noticed while debugging are very small values that ought to be flushed to zero but aren't, in my configuration.  Could this be contributing to the regression problems?  E.g. in icepack_isotopes.F90,

> ! condensation of vapor onto snow and ice
> 
>       TsfK = Tsfc + Tffresh
> 
>       if (evaps > c0) then   ! condensation to snow
>          do k = 1, n_iso
>             ratio = c1   ! ratio between 18O(HDO) and 16O in humidity
>             alphai = c1  ! fractionation coefficient
>             if (isotope_frac_method.ne.'nfrac' .and. Qref_iso(2)>puny) &
>                ratio = Qref_iso(k)/Qref_iso(2)
>             if (Qref_iso(2) <= puny) ratio = c0
>             if (isotope_frac_method.ne.'nfrac' .and. k==1) alphai = wiso_alpi(3,TsfK)
>             if (isotope_frac_method.ne.'nfrac' .and. k==2) alphai = wiso_alpi(2,TsfK)
>             if (isotope_frac_method.ne.'nfrac' .and. k==3) alphai = wiso_alpi(4,TsfK)
>             work = alphai*ratio*rhos*evaps*aicen
>             fiso_evapn(k) = fiso_evapn(k) + work/dt
>             isosno(k) = isosno(k) + work
>          enddo
>          dzsno = dzsno + evaps
>       endif
> print*,'update A',isosno(1),work
> print*,'update A',ratio, alphai,evaps,aicen,Qref_iso(:)
> 
> 
> 
>  update A   0.0000000000000000        6.9531242993815443E-310
>  update A   2.1813440625567039E-314   2.2473962091190323E-314  -2.1645554965428471E-005   1.6849178281029695E-002   0.0000000000000000        0.0000000000000000        0.0000000000000000     
>

apcraig:
The small numbers could be affecting the results if they are supposed to be zero.  I assume the case above is one with isotopes on.  When isotopes are off, are we getting some isotope code to activate when it shouldn't?

We don't automatically underflow to zero with all compilers.  I also don't think we want to rely on flush to zero to zero out stuff that should be zero.  We need to explicitly deactivate code that should be off thru logicals or other checks.  We also need to make sure if something is computed that should be identically zero, the implementation does that, not the flush to zero.

1e-314 may or may not be below the flush to zero limit.  Can you tell me the configuration you're running.  I'd like to try to duplicate those results and do a bit of additional debugging.  But I still think we need to try to have implementations that compute zeros directly as much as possible.

eclare108213:
That case was with isotopes on.  I don't know if any of the isotopes code is activated when tr_iso=F.  It shouldn't be, but I'm not sure how to test for that. We need to look for that in during the code review.  This particular run got 10^-314, but other runs got 10^-50, which is definitely not below the ftz cutoff, and the values were definitely different on every run.  Here is the directory name ~ my test configuration:
conda_macos_restart_gx3_1x1_isotope.iso7
NB I don't know if the underflows are still happening in that code block, after I fixed the restart problem.

My thought was that this kind of thing might be happening in other configurations of the code without us being aware of it.  I agree we ought to code these problems out rather than rely on compiler flags, but in some cases (like the stress routine in the dynamics), the extra coding slows down the code too much (so in stress I've included the coding but commented it out as the default).

apcraig:
I'll take a look.  Just to confirm, that was with CICE, not Icepack?

eclare108213:
yes

apcraig:
OK, I did some tests on conrad with the intel compiler.  If you remember, this machine/compiler led to many regression failures while conrad with pgi, gnu, and cray did not.  In addition, conrad+intel does have a flush-to-zero (ftz) flag turned on by default.  

I ran cases with the ftz flag on and off at conrad+intel, and I got bit-for-bit results including comparisons of the "update A" output that I manually added and checked.  That suggests two things.  First is that the ftz is probably NOT playing a role in the regression failures with the intel compiler.  I continue to believe that is compiler optimization.  Second is that the crazy numbers (E-50 to E-300) are probably OK computationally.  I see those kinds of numbers in the "update A" output in my testing as well.  I'm not suggesting the algorithm that generates such tiny numbers is ideal, I'm just suggesting it's probably not leading to any technical issues that we can detect at this point.

As noted in #238, there are still a number of open questions about the tracer indexing and control.  That may (or may not) be contributing to the small numbers.  I suspect it isn't, but we still don't know.  At this point, many machines and compilers are fully passing regression testing with the isotope mods.  Where they don't, the most likely reason is compiler optimization and the fact that the debug tests do pass makes that case even stronger.  I think we should continue to evaluate the algorithms that are producing the tiny numbers, but would treat that as a separate task at this point.

eclare108213:
Good, I think we can move forward assuming it's a compiler issue.

One thing about the tiny numbers that bothers me:  In the code block above, Qref_iso is identically zero, which means that ratio is set to c0.  But c0 apparently is not identically zero.  Why not?

apcraig:
That is a good question.  I'm not sure I'm seeing that but will have a look.

klindsay28:
@eclare108213 , it looks like `evaps` is negative, so the block is not getting executed, and some of the printed values are non-initialized local variables.

apcraig:
good catch @klindsay28, I was just coming to the same conclusion.  In my testing, I shifted the print statement into the do loop to pick up more details.  I suspect what @klindsay28 suggested is the issue in this case.

apcraig:
Just another data point.  I ran the same case in master and the iso2 versions with -O0, -O1, and -O2.  -O0 were bit-for-bit, -O1 and -O2 were not.   I don't know whether to draw any particular conclusions, but wanted to report it since I had the data.  

The initial divergence in the log files in the O1 case is

```
1283c1293
< arwt frsh h2o kg in dt =   -5.40653119731186719E+12   1.04545216036125586E+13
---
> arwt frsh h2o kg in dt =   -5.40653119731186133E+12   1.04545216036125586E+13
1288c1298
< arwt water flux        =    6.55326937280221680E+12  -9.84273844887030273E+12
---
> arwt water flux        =    6.55326937280221094E+12  -9.84273844887030273E+12
1290c1300
< water flux error       =    1.51587407740070224E-17  -9.72191257532538586E-17
---
> water flux error       =    1.47490450774122382E-17  -9.72191257532538586E-17
1293c1303
< arwt ocn heat flux (W) =   -7.01774634835119844E+13  -6.35513940521804000E+14
---
> arwt ocn heat flux (W) =   -7.01774634835119531E+13  -6.35513940521804000E+14
1296c1306
< arwt net heat      (J) =   -1.71118194416788659E+18   2.88453583201328998E+18
---
> arwt net heat      (J) =   -1.71118194416788685E+18   2.88453583201328998E+18
1298c1308
< arwt heat error        =    2.59724157989657469E-11   1.28489799381921792E-09
---
> arwt heat error        =    2.59724158560667225E-11   1.28489799381921792E-09
1307,1308c1317,1318
< arwt salt flx in dt(kg)=   -2.25539389111198578E+10   3.51042708101931839E+10
< arwt salt flx error    =    1.19243440478172217E-16   1.78616380907256175E-17
---
> arwt salt flx in dt(kg)=   -2.25539389111198273E+10   3.51042708101931839E+10
> arwt salt flx error    =    1.19802284956101569E-16   1.78616380907256175E-17
1364,1365c1374,1375
< arwt frzl h2o kg in dt =    5.96201132673940674E+11   5.55663299001190290E+07
< arwt frsh h2o kg in dt =   -5.39926154809379199E+12   1.03483486973514160E+13
---
> arwt frzl h2o kg in dt =    5.96201132673939209E+11   5.55663299001190290E+07
> arwt frsh h2o kg in dt =   -5.39926154809379395E+12   1.03483486973514160E+13
```

The initial divergence in the O2 case is 

```
1703c1713
< arwt ocn heat flux (W) =   -6.75616246231895547E+13  -6.13739866966776750E+14
---
> arwt ocn heat flux (W) =   -6.75616246231895625E+13  -6.13739866966776750E+14
1857c1867
< arwt frsh h2o kg in dt =   -5.31949959984610840E+12   1.02487066973116191E+13
---
> arwt frsh h2o kg in dt =   -5.31949959984610156E+12   1.02487066973116191E+13
1862c1872
< arwt water flux        =    6.43998729252299902E+12  -9.63346022416123828E+12
---
> arwt water flux        =    6.43998729252299219E+12  -9.63346022416123828E+12
1864c1874
< water flux error       =   -6.97679890361102138E-17  -2.21789562657725422E-16
---
> water flux error       =   -7.02444533514787689E-17  -2.21789562657725422E-16
1867c1877
< arwt ocn heat flux (W) =   -6.67571433450328594E+13  -6.08365856387301625E+14
---
> arwt ocn heat flux (W) =   -6.67571433450333125E+13  -6.08365856387301625E+14
1869,1872c1879,1882
< arwt tot energy    (J) =   -4.49564195778666902E+21  -6.71187951565688275E+21
< arwt net heat      (J) =   -1.79456705788091981E+18   2.89003801432144230E+18
< arwt tot energy chng(J)=   -1.79456679704159846E+18   2.89004502843077427E+18
< arwt heat error        =    5.80204837914669132E-11   1.04502908844029485E-09
---
> arwt tot energy    (J) =   -4.49564195778666955E+21  -6.71187951565688275E+21
> arwt net heat      (J) =   -1.79456705788091827E+18   2.89003801432144230E+18
> arwt tot energy chng(J)=   -1.79456679704212275E+18   2.89004502843077427E+18
> arwt heat error        =    5.80203668284158145E-11   1.04502908844029485E-09
1881,1882c1891,1892
< arwt salt flx in dt(kg)=   -2.21796319783276215E+10   3.37827894890573425E+10
< arwt salt flx error    =    1.07377154358655962E-16   2.57526465137704803E-16
---
> arwt salt flx in dt(kg)=   -2.21796319783275948E+10   3.37827894890573425E+10
> arwt salt flx error    =    1.07864599287651159E-16   2.57526465137704803E-16
1920c1930
< heat used (W/m^2)      =        3.07477563406503895      16.17082914631258106
---
> heat used (W/m^2)      =        3.07477563406503762      16.17082914631258106
1924c1934
< total ice area  (km^2) =    1.56276679519851003E+07   1.55261392267098539E+07
---
> total ice area  (km^2) =    1.56276679520084020E+07   1.55261392267098539E+07
```

eclare108213:
I think we have a bfbflag (or something like that) in the namelist.  Would setting it hide this issue and make the regression tests work?

apcraig:
I don't believe the bfbflag would make a difference.  That feature operates only on the global sum diagnostics and deals with identical data distributed in memory in different orders (like different pe counts or decompositions).  In this case, we are getting different answers with identical decompositions.  I believe the code is just being optimized differently by the compiler so the order of operations is changed in some parts of the code.  We could try to do things like set max_iso=0 and see if that reverts back to prior results.  If that doesn't work, I could also try to debug the divergence and identify what parts of the code are leading to the divergence.  This would be time consuming, but I'm willing to do it.  It would not necessarily lead to a solution, but we would probably gain confidence that it's not a coding error, although I think that confidence is high already since non-intel compilers are bit-for-bit and reduced optimization with the intel compiler is also bit-for-bit.

eclare108213:
Thanks, both of you.  I'm just throwing out ideas for dealing with the nonBFB issue.  I don't think it needs to be debugged at this point - I agree it's probably a compiler issue and we can continue to move forward on this PR.

eclare108213:
@apcraig I updated icepack in this PR, and many of the travis suite tests work.  But there are a couple of errors that I don't understand.  I did relaunch it with the same results.  Please, could you take a look?  Thanks.  

I also ran a full base_suite on badger.  Many of the regression comparisons failed, but all of the debug cases passed.  
https://github.com/CICE-Consortium/Test-Results/wiki/7ab48ebeec.badger.intel.20-04-06.130500

apcraig:
The travis error is the same one we saw this weekend on master.  I have relaunched again.  My sense of the error we're seeing is that there is something flakey with the file system or similar.  I don't think it has anything to do with our scripts or code.

The badger results are what we expect.  I will fire off a full suite again on conrad just to further confirm everything is still as expected.

eclare108213:
It looks like the checks passed this time.  Thanks @apcraig.
I think all that's left to do for this PR is making the driver changes in standalone/cice for all of the other drivers.

apcraig:
The conrad test results with the latest iso2 mods (including Icepack) are here, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#7fb8d693c87fe0b4a26f1da030daa91c2e1876e5

Several intel tests fail as expected.  There is also one cray regression test failing.  I have looked at the diff in the restart file and it's just one field, frzmlt.  I'm not sure how to interpret that failure for one field for one test.

eclare108213:
I prefer to make the [hadgem3 change](#430) in a separate PR.

apcraig:
I normally try to keep all the drivers updated when a change is made.  I also generally do not test those.  It's usually safe to do that as changes are purely internal and not related to anything coupling.  Some of the driver files should be very close to each other.  At some point, I tried to sync them up as much as I could.  The caps are different but the Init, Run, and Final code should be similar.  That is likely where the changes go so it should not be difficult or dangerous to update the other drivers.

I will have a look later today and create a PR to the iso2 branch.  If there are other driver changes that have been missed in prior updates, I'll try to address those as well.

TillRasmussen:
It would be nice if drivers are updated along with the code otherwise it can be hard to tell what a potential difference may make. Regarding the Hadley driver try to ask @phil-blain or @JFLemieux73 if they are using it.

apcraig:
Quick question, what do people think about getting rid of the CICE_RunMod.F90_debug file in the non-standalone drivers?  This exists in the coupled model directories largely to mimic the standalone driver, but I do not think they are ever used.  This would reduce the maintenance burden in the drivers.

TillRasmussen:
I/we are not using it.

daveh150:
We do not use the CICE_RunMod.F90_debug. I am okay if it is removed. 

 

Thanks!

 

David A. Hebert

Naval Research Laboratory 
(228)688-5846; DSN 828-5846 

eclare108213:
That is my primary in-code debugging tool.  If there is a better way to implement it, let's do it, but let's not lose the capability.  We could also just maintain it in the standalone/cice directory and not the others.

phil-blain:
I would suggest merging it with the main file and using a preprocessor macro that would be activated with the existing `./cice.setup -s debug` option to compile these parts of the file. I think this is one good use of preprocessor macros.

--
my two cents (haven't looked at the code)

apcraig:
I'm proposing we keep it in the standalone model and maintain it there but remove it from the coupler model driver directories.

I think there probably are better ways to implement it in the regular file.  Even a cpp that is turned on at the top of the code would be reasonable approach.  Then we don't have to maintain two files.  Lets take that on as a separate issue.

eclare108213:
I added @apcraig's changes to the other drivers.  I think this one is ready to merge.

apcraig:
I am going to merge the PR and then kick off a bunch of testing.

Issue #437, opened 2020-05-04
Improve testing code coverage


apcraig:
The codecov tool is working for Icepack and CICE, (see #434, #436) and a handful of people can run the testing.  We should next evaluate the codecov results and see if we can improve testing coverage.  There is an equivalent issue in https://github.com/CICE-Consortium/Icepack/issues/194, but I thought I'd add one specifically for CICE because the efforts may not overlap.

The codecov results for CICE can be found here,

https://codecov.io/gh/apcraig/Test_CICE_Icepack

and the current results by file are

https://codecov.io/gh/apcraig/Test_CICE_Icepack/list/3d1a95bc42877d9ab1a5ddf17d5dd255ce4d929b/

apcraig:
There is a google doc that is trying to track some of the coverage issues

https://docs.google.com/document/d/1OMqg8u66uobqD_sbIHSM6ZXLEbv7FFRnQKSk2TH1c7M

apcraig:
We added an io_suite to test io_binary, io_netcdf, and io_pio as well as a histall option to turn on all history fields in #447.

apcraig:
The coverage testing moved to lcov and output can be found here, https://apcraig.github.io/.

rallard77:
Tony,

Your charts look great and are very intuitive.

Rick

eclare108213:
@CICE-Consortium/devteam 
I have gone through the worst offenders in the code (modules highlighted red in the [lcov output](https://apcraig.github.io/lcov_cice_201120-001611:d95a4f35ee:7:first,base,travis,decomp,reprosum,io,quick/index.html)) and made notes for the things  we could test better in a [spreadsheet](https://docs.google.com/spreadsheets/d/12tIfm5OzvzH_LF_ie9WbOn_w5aY1hv4SrqTQQ8bqM-s/edit?usp=sharing).  Text in red is high priority; text in purple is BGC (which needs a concerted effort all by itself).  I've assigned many of you to take a look at and (maybe) address particular items in this list (negotiable - let me know if you won't be able to do this).  Options include adding a test, choosing not to test, deleting the unexercised code, etc.  

[...]

Issue #411, opened 2020-02-28
scripts: copy cice.settings to ICE_RUNDIR


phil-blain:
It could be useful if the scripts would copy `cice.settings` to the run directory. 

One of the reason would be so that there exist a back reference to the test case directory (`ICE_CASEDIR`) in the run directory, so that we can easily jump back and forth between a test case (possibly in a test suite) and its run directory. (If we just go back and forth we can use `cd - ` to do that, but if we `cd` someplace else then we can't use this trick anymore). 

Having `cice.settings` in the run directory, one could easily create an shell alias that greps `cice.settings` looking for `ICE_CASEDIR` and `cd`'s there.
Right now I have these:
~~~
$ alias cdrun
alias cdrun='cd ~/local/CICEDIRS/CICE_RUNS/${PWD##*/}'
$ alias cdcase
alias cdcase='cd ~/local/CICEDIRS/cases/${PWD##*/}'
~~~

These work fine for cases and `cdrun` works for standalone tests and tests in a test suite, but once I'm in `ICE_RUNDIR` there's no easy way to go back to `ICE_CASEDIR` if `cd -` does not work because I `cd`'ed somehwere else (eg.  `ICE_RUNDIR/{restarts,compoile,history}`) and then came back to  `ICE_RUNDIR`.

@apcraig

apcraig:
This seems to be related to something mentioned in #412

apcraig:
This can be closed, both the model .settings file and the env machine file are copied to the run directory in Icepack and CICE now.

Issue #298, opened 2019-03-13
FSD:  miscellaneous


eclare108213:
- [x] Move l_conservation_check into namelist
- [x] Add/clean up FSD sanity checks
- [x] Add FSD timers

apcraig:
@eclare108213, do you have a preference what the namelist should be called or in what namelist group it should appear.  By default, I'll use l_conservation_check and put it in dynamics.  But maybe we need to change the name to something a little clearer, maybe "conservation_checks" or "tracer_conservation_checks"?  Should it be in the dynamics namelist group?

eclare108213:
It could just be conservation_check.  The l_ just indicates it's a logical flag, not something we have to have.  It could also be shorter, like conserv_check.  I could go in the dynamics namelist, but in icepack it also is used for zbgc, so maybe it should go in the general-use namelist (setup_nml).

apcraig:
Thanks @eclare108213, I'll probably go with conserv_check and put it in setup_nml.

apcraig:
https://github.com/CICE-Consortium/Icepack/pull/318 and #450 have been merged.  @eclare108213, please close if you feel all tasks are complete.

Issue #412, opened 2020-03-02
scripts: add support for launching CICE in a debugger


phil-blain:
When I want to launch CICE in a debugger (like Totalview), what I do in my case directory is to copy `cice.run` as `cice.debug` and prefix the command for launching cice with the debugger command, i.e.
~~~diff
diff --git 1/picard-1x1-fev2020/cice.run 2/picard-1x1-fev2020/cice.debug
index 8b3fd44..bd7f5a0 100755
--- 1/picard-1x1-fev2020/cice.run
+++ 2/picard-1x1-fev2020/cice.debug
@@ -36,7 +36,7 @@ echo "CICE rundir is ${ICE_RUNDIR}"
 echo "CICE log file is ${ICE_RUNLOG_FILE}"
 echo "CICE run started : `date`"
 
-./cice >&! $ICE_RUNLOG_FILE
+totalview -newui ./cice >&! $ICE_RUNLOG_FILE
 echo "CICE run finished: `date`"
 echo " "
~~~

It would be great if the scripts would support this use case, either through a separate script `cice.debug` or a flag to `cice.run`, i.e. `./cice.run --debug`.

The command to launch the debugger could be added as an `ICE_MACHINE_DEBUGGER` variable in the env file, as well as any `module load` command (or similar) necessary to access it (and we could leverage `ICE_BLDDEBUG` to only execute these commands when compiling/running in debug mode).

This would make it easy to debug CICE, at least in an interactive session, either in serial mode or with a small number of MPI processes. Making it also work for jobs submitted to a cluster would probably be a lot more work if we want it to be generic enough for multiple job schedulers / debuggers.

eclare108213:
I use totalview too.  My workflow is to request an interactive session from the scheduler/queuing system, then cd to the run directory and type 

> totalview mpirun -a -np [numprocs] cice

(or leave out the mpirun part for serial tests)

Seems a lot more straightforward, and then we don't have to worry about users wanting other debuggers.  I often cd to the run directory for other reasons anyhow -- we can't do everything from the case directory.  My wish would be to make that cd easier, rather than trying to increase capabilities in the case directory.  But I'm open to others' opinions about this.

phil-blain:
@eclare108213 when #450 is merged, `cice.settings` will be copied to the run directory, so these aliases would work to go easily from `ICE_CASEDIR` to `ICE_RUNDIR` and back :

**For Bash / compatible**
```shell
[...]

PR #450, opened 2020-05-22
Move conserv_check to namelist, update test suite to improve coverage

apcraig:
## PR checklist
    Move conserv_check to namelist, update test suite to improve coverage
- [X] Developer(s): 
    apcraig
    https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#52b875401168ae7591cee21a99e680e51329eaa0
   This changes alt03 and alt04 configurations to extend test coverage, will change answers of alt03 and alt04.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

- rename l_conservation_check to conserv_check and move to namelist (#298)
- copy cice.settings to run directory (#411)
- modify alt03 and alt04 to improve code coverage, changes answers (#437)
   - add tr_aero (with calc_tsfc true and false)
   - test ncat=6
   - test evp 1d kernel
   - add conserv_check
- add abort for ew/ns boundary_type=closed
- fix some fsd logic
- update a few error messages


eclare108213:
Ah, the conservation check in the CICE repo is for horizontal transport; there is another one in the Icepack repo for the ice thickness distribution (mostly but not necessarily ridging).

apcraig:
I was thinking just this morning whether we want a name a little more specific.  conserv_check seems pretty broad.  This conserv_check only checks horizontal transport.  Should we call it remap_conserv_check or something more specific like that?  I can make this change quickly if it's helpful.  Ultimately, it seems we might have different levels of conservation checks in the code, it would be good to be able to distinguish them.

apcraig:
Or maybe we should use conserv_check to trigger all conservation checks we might have in the system.  Do we want one flag to turn on checks or do we want to have multiple flags to support different checks?

eclare108213:
At the moment, they are coded separately but the logical flag is the same, so it's easier to grep.  In practice, I think it would be nicer for one flag to turn them all on, to make debugging a little more straight-forward.  Just make sure the output distinguishes which instance is printing.  IMHO

apcraig:
> At the moment, they are coded separately but the logical flag is the same, so it's easier to grep. In practice, I think it would be nicer for one flag to turn them all on, to make debugging a little more straight-forward. Just make sure the output distinguishes which instance is printing. IMHO

Looking in icepack, there are two places where l_conservation_check occurs, in icepack_mechred.F90 and icepack_therm_itd.F90.  Both are hardwired values in each subroutine.  I propose we rename l_conservation_check to conserv_check in icepack.  The I propose we also add a optional argument to icepack (say in icepack_parameters) that allows a driver (CICE or other) to turn on that check.  I would also update the output to make sure the error is better identified.  Does that sound OK?

I'll try to work on this today as it's involve an update to Icepack that needs to come in before CICE can use it.  We could still merge this now if that was desired.  Lets see how timing works out getting an update to Icepack before we merge this PR.

configuration/scripts/cice.run.setup.csh
```diff
@@ -46,6 +46,7 @@ cd \${ICE_RUNDIR}
 setenv OMP_NUM_THREADS ${nthrds}
 
 cp -f \${ICE_CASEDIR}/ice_in \${ICE_RUNDIR}
+cp -f \${ICE_CASEDIR}/cice.settings \${ICE_RUNDIR}
```
phil-blain, line 49: Minor point : in fact, I think *also* copying the env file could be a good idea in the context of #412, since then one could do from the run directory:
```bash
$ csh # if not already using csh
% source env.<machine>_<compiler>
% source cice.settings
% gdb ./cice  # or another debugger
```

This is a little bit easier than having to start csh in the case in the run directory, source both files, then `cd` to the run directory...

apcraig: Sounds good, I just checked that mod in.

apcraig:
I will include https://github.com/CICE-Consortium/Icepack/pull/318 once it's merged in icepack and update the conserv_check icepack parameters setting from CICE.  This PR should be on hold for the moment and I have assigned it to me for the time being.

apcraig:
I have updated icepack and rerun the full test suite on cheyenne with 3 compilers and compared to the prior master version, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#52b875401168ae7591cee21a99e680e51329eaa0.  All tests pass as expected.
  - alt03 and alt04 tests have changed answers.  that represents 16 comparison failures per compiler.
  - pgi bgc cases still fail, that represents 9 failures with cheyenne pgi.

I believe this is ready to merge if others agree.

apcraig:
I also ran on izumi last night and the test suite results are consistent with prior versions, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#587b972ff83d5e69815172faa59ee025ce41ddf6

phil-blain:
Hi @apcraig ! 

What was the reason to add aborts for `{ns,ew}_boundary_type == closed` in this PR ? I'm realizing that we were using this setting in our NEMO3.6/CICE4 systems and I'm transitioning these to CICE6 and I'm bit puzzled...

eclare108213:
Hi @phil-blain !
I could never get the 'internal' closed boundary conditions to work correctly on corners of the grid (probably when the B.C.s were mixed, e.g. with closed overlapping cyclical).  The only way to properly (and safely) have closed boundary conditions in CICE's current implementation is to mask them out with a land mask, and I think the code now enforces that.

phil-blain:
Hi Elizabeth, thanks for your answer. Maybe we were doing things in a hacky way then... we have a lot of internal code that was added at different places in CICE4 and 5, and I'm trying to get rid of most of it, just coupling to CICE6 "out-of-the-box"...
I'll keep investigating. Maybe we were abusing this setting in some way and we could do it in a different way.

eclare108213:
You might not have noticed the problem in your tests.  If I remember correctly, the errors were round-off level and only visible when doing symmetry tests.

PR #460, opened 2020-06-06
update izumi and implement intel20 workaround

apcraig:
## PR checklist
    Update izumi and implement intel20 workaround
- [X] Developer(s): 
    apcraig

    Full test suite run on izumi
    -  izumi had a recent upgrade
    -  izumi only supports one version of each compiler right now
    -  half the intel tests fail without the INTEL20_WORKAROUND fix
            - believe this is a compiler bug
            - recommend this be reviewed further and hopefully removed at a later time
    -  pgi tests are not bit-for-bit and currently do not restart exactly
            - have verified results are reproducible run-to-run
            - have reviewed restart read/write diagnostic, they look fine
            - requires more debugging
    - https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#4df9a8f90aa7f1e992e2c83d61caf96aa8bfc4cd  shows results both with and without the INTEL20_WORKAROUND fix

    - [X] bit for bit except izumi pgi
    - [ ] more substantial 
    - [X] Yes, Icepack is updated which includes bug fix for debug tests
    - [ ] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

- Updated izumi due to upgrade, old compilers no longer available
- Added feature to turn on machine/compiler dependent cpps via the machine files in the scripts.
- Fixed nt_zbgc_frac initialization error (thought we fixed this before)
- Updated icepack, includes a fix from https://github.com/CICE-Consortium/Icepack/pull/321 needed to pass some debug mode tests.

    Full test suite run on izumi
    -  izumi had a recent upgrade
    -  izumi only supports one version of each compiler right now and they are the latest compilers which we have little experience with
    -  half the intel tests fail without the INTEL20_WORKAROUND fix
            - believe this is a compiler bug
            - recommend this be reviewed further and hopefully removed at a later time
    -  pgi tests are not bit-for-bit and currently do not restart exactly
            - have verified results are reproducible run-to-run
            - have reviewed restart read/write diagnostic, they look fine
            - requires more debugging

### DIFF
--- a/cice.setup
+++ b/cice.setup
@@ -572,6 +572,13 @@ EOF
       continue
     endif
 
+    # unset env variables that might not exist in env machine file
+    # to avoid any carry over during multi compiler suites
+    unsetenv ICE_MACHINE_MAXTHREADS
+    unsetenv ICE_MACHINE_MAXPES
+    unsetenv ICE_MACHINE_QUIETMODE
+    unsetenv ICE_MACHINE_CPPDEFS
+    unsetenv ICE_MACHINE_QSTAT
     source ${ICE_SCRIPTS}/machines/env.${machcomp} -nomodules || exit 2
 
     # Obtain the test name, sets, grid, and PE information from .ts file
@@ -782,6 +789,11 @@ EOF
       set quietmode = ${ICE_MACHINE_QUIETMODE}
     endif
 
+    set cppdefs = ""
+    if ($?ICE_MACHINE_CPPDEFS) then
+      set cppdefs = ${ICE_MACHINE_CPPDEFS}
+    endif
+
     if (${acct} == ${spval}) then
       if (-e ~/.cice_proj) then
         set acct = `head -1 ~/.cice_proj`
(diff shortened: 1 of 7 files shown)


eclare108213:
@CICE-Consortium/devteam (or anyone reading this):
Do you have access to a machine with the intel20 compiler, to test this PR and see whether the problem is izumi-specific or if it's a real bug, either in the code or in the compiler?

apcraig:
I can't say for sure whether all the array need to be allocated to workaround this issue.  I'm pretty sure that's not the case, but I did try to pin down whether it was one or a few arrays that were causing the problem and couldn't find any array alone that created the error.  My sense is that it's a memory issue, not a problem with a particular array declaration or usage.  I played around a lot with the argument declarations, local variable declarations, and source code.  I even did things like turning off all the code in the subroutine while leaving all the declarations on, I degraded the interface down to no arguments, and many other things to try to isolate the issue.  I also tried a number of compiler flags to try to increase the memory management.  The error was surprising robust and came down to these local arrays.

What I'll do is merge this PR and start testing across a wider set of machines.  I'll also try to find another intel20 compiler on another machine to see if the problem is seen on another machine.  I think that's a very good idea.

Once we have a bit more information, we can decide what the next steps should be.

PR #491, opened 2020-07-20
dynamics: add implicit VP solver

phil-blain:
## PR checklist
    Add an implicit solver for the VP rheology based on Picard iteration
- [X] Developer(s): 
    P. Blain, J-F Lemieux.
   Here are the results from the `base_suite` ([Test-Results link](https://github.com/CICE-Consortium/Test-Results/wiki/c1caa647ad.daley.intel.20-07-20.163306.0)):
    ```bash
    249 measured results of 249 total results
    216 of 249 tests PASSED
    0 of 249 tests PENDING
    2 of 249 tests MISSING data
    31 of 249 tests FAILED
    ```
    This is comparing the tip of this branch to c22c6d5 (clarify CICE citation info in doc (#487), 2020-07-13).

    The two missing data tests are:
    
    ```bash
    MISS daley_intel_smoke_gx3_4x1_diag1_run5day_thread bfbcomp daley_intel_smoke_gx3_8x2_diag1_run5day missing-data
    MISS daley_intel_smoke_gx3_4x1_dynpicard_medium compare c22c6d5  -1 -1 -1 missing-data
    ```
    
    The second one is expected, it's the test I add in this branch so it does not yet exist on `master`. The first one is unexpected, I'll look into that. **EDIT 24/07 :** see https://github.com/CICE-Consortium/CICE/pull/491#issuecomment-663544517, this was simply  a timing issues between the two tests.

    The 31 failing tests are all "different-data". I expected this branch to not be bit for bit since did some refactoring of the dynamics code to increase code reuse. I've not yet investigated whether the answers change a lot or just at machine precision level. I'll update the PR when that's done.

  **EDIT 22/09/20** The failing tests seem to be those where OpenMP (threading) is turned on and compiler optimization is also turned on, see https://github.com/CICE-Consortium/CICE/pull/491#issuecomment-663226841.

  I've also run the QC test, **comparing an EVP run and a VP run** (with 20 nonlinear iterations instead of the default of 4). The QC test passes:
    ```python
    $ ./configuration/scripts/tests/QC/cice.t-test.py ~/data/site4/cice/runs/daley_intel_smoke_gx1_32x1_medium_qc.qc/     ~/data/site4/cice/runs/daley_intel_smoke_gx1_160x1_dynpicard_medium_nonlin20_qc.qc/
    INFO:__main__:  /home/phb001/data/site4/cice/runs/daley_intel_smoke_gx1_32x1_medium_qc.qc/
    INFO:__main__:  /home/phb001/data/site4/cice/runs/daley_intel_smoke_gx1_160x1_dynpicard_medium_nonlin20_qc.qc/
    INFO:__main__:Creating map of the data (ice_thickness_daley_intel_smoke_gx1_32x1_medium_qc.qc.png)
    INFO:__main__:Creating map of the data (ice_thickness_daley_intel_smoke_gx1_160x1_dynpicard_medium_nonlin20_qc.qc.png)
    INFO:__main__:Creating map of the data (ice_thickness_daley_intel_smoke_gx1_32x1_medium_qc.qc_minus_daley_intel_smoke_gx1_160x1_dynpicard_medium_nonlin20_qc.qc.png)
    INFO:__main__:
    ```

  **EDIT 13/08/2020** The QC test is bit-for-bit in the default decomposition ( `-p 44x1`). The QC test passes in a 160x2 decomposition.

    - [ ] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [X] Yes: **I've added a test to the base_suite with the settings `dynpicard` (i.e. using the implicit solver).**
    - [ ] No
    - [x] Yes. **Here are links to the specific sections of the documentation that I changed**:
      - CICE index: https://cice-consortium-cice--491.org.readthedocs.build/en/491/cice_index.html?highlight=kdyn
      - Dynamics in the Science guide: https://cice-consortium-cice--491.org.readthedocs.build/en/491/science_guide/sg_dynamics.html (**this is the biggest changes**)
      - Dynamical solvers in the Developer guide: https://cice-consortium-cice--491.org.readthedocs.build/en/491/developer_guide/dg_dynamics.html#dynamical-solvers
      - Case settings in the User guide: https://cice-consortium-cice--491.org.readthedocs.build/en/491/user_guide/ug_case_settings.html#dynamics-nml
        - [ ] Yes
        - [ ] No 

We've been talking about it for a long time, and it's finally ready for review!

This PR adds the implicit VP solver that JF and I have been working on for the last 2 years. 

The VP solver is implemented in a new module in `dynamics/`, `ice_dyn_vp.F90`. Here are some points that I would like to highlight:
- The subroutine `imp_solver` is the main driver (equivalent to subroutine `evp` for EVP), and should be read side-by-side with `evp` for ease of reviewing. We are aware that there is a lot of repetition between `imp_solver` and `evp`, but there was also already a lot of repetition between `evp` and `eap` in `ice_dyn_eap.F90`. If everyone is on-board, I could work on refactoring this to increase code reuse (although that would be in a subsequent PR).
- The nonlinear equations resulting from the VP rheology discretization are solved using Picard iteration. This is done in subroutine `anderson_solver`. This subroutine also implements Anderson Acceleration, which is an acceleration method for fixed point iterations (Picard iteration is a fixed point iteration). Picard iteration is implemented as a special case of Anderson acceleration. At the moment the Anderson solver is not parallelized, so it is not documented. The Picard solver is parallelized and so it is the only one that is documented in the doc. Also, I've added an abort in the Anderson code if this method is used in parallel. The namelist setting `algo_nonlin` can be used to change from 'picard' to 'anderson', this is also not documented and it defaults to 'picard'. 
- The namelist setting `use_mean_vrel` is also not documented because it should always be set to `.true.` for the Picard solver for faster convergence. It is present in the namelist because it has to be set to `false` for the Anderson solver.
- We set a default of 4 nonlinear iterations (`maxits_nonlin`), 1E-2 for the tolerance of the FGMRES solver (`reltol_fgmres`) and 5 PGMRES iterations (`dim_pgmres`, `maxits_pgmres`). @JFLemieux73 tells me this is in line with how implicit VP solvers are used in other models (ex. MITgcm). This means that by default, the code does *not* iterate until the solution is "converged". For numerics studies, I've added an options setting `set_nm.nonlin5000` that sets `maxits_nonlin` to 5000 so that the code iterates until the solution reaches the desired tolerance (`reltol_nonlin`). 
- We are aware there is a lot of code duplication between some subroutines in `ice_dyn_vp`, and also between the `stress` and `stress_eap` subroutines in `ice_dyn_[ea]vp`, relating to the different computations for rheology. This is documented (briefly) in https://github.com/phil-blain/CICE/issues/36. I plan to refactor the code to reduce code duplication in a subsequent PR.
- The subroutines `fgmres` and `pgmres` are also very similar. `fgmres` implements the FGMRES linear solver, and `pgmres` implements the right-preconditioned GMRES linear solver, which is used as a preconditioner for FGMRES. In theory GMRES is a special case of FGMRES, with a fixed preconditioner, so the code could be refactored so that they are both implemented in the same subroutine. However, to have only one subroutine would mean that this subroutine would have to be a recursive subroutine (since `fgmres` calls `pgmres`). I'm not sure of the performance implication of that, so as a first step I used separate subroutines.
- The Anderson solver needs LAPACK at the moment, but **the Picard solver does not**. The code compiles out of the box without LAPACK since I've protected the LAPACK calls with preprocessor macro, `CICE_USE_LAPACK`. This preprocessor macro can be activated using the setting `set_env.lapack`, although this is not documented (since the Anderson solver itself is not documented). 
- I've tried to make clear commit messages that explain *why* I made the changes I made in each commit. If you want to know why did a particular change, you can use the following commands to read commit messages for changes touching specific files:
    ```bash
    git fetch https://github.com/phil-blain/CICE.git parallel-picard
    git log upstream/master..FETCH_HEAD <path/to/file>
    ```
   Each message can also be read in the "Commits" tab of the PR by clicking in the `...` button next to the summary line.


phil-blain:
I've updated the PR description with link to the specific parts of the doc that I modified.

phil-blain:
I've updated the PR description with the result of the `base_suite`.

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -1017,6 +1068,36 @@ subroutine input_data
          endif
       endif
 
+      ! Implicit solver input validation
```
eclare108213, line 1071: shouldn't these error checks and aborts be wrapped in kdyn=3?

phil-blain: yes, I guess that doesn't hurt. Good idea, thanks.

phil-blain:
> Awesome!
> 
> ice_HaloUpdate_vel is much more elegant than the loading/unloading of fld2. Is this new routine usable for any vector quantity? If so, then it could be named ice_HaloUpdate_vec and put in the boundary module as a generic routine.
> 

Good point. I think it depends on what kind of halo updates are needed. I replaced all occurrences of `fld2` by calls to `ice_haloUpdate_vel`, and they were all for the velocity field. One reason I prefer to keep it in `ice_dyn_shared` instead of `ice_boundary` is that this way the code is not duplicated between the serial and MPI versions..


> Are you sure about the factors of 2 in the comment lines for shearing strain rate? I would think that someone else would have noticed that, after all this time, if it were wrong. (But it might be wrong!)
> 

I *think* my change is correct. This is in commit fb4a49c2d322e310f334349e4244e855d66f2b8b: 
```
All occurrences of 'shearing strain rate' as a comment before the
computation of the quantities `shear{n,s}{e,w}` define these quantities
as

    shearing strain rate  =  e_12

However, the correct definition of these quantities is 2*e_12.

Bring the code in line with the definition of the shearing strain rate
in the documentation (see the 'Internal stress' section in
doc/source/science_guide/sg_dynamics.rst), which defines

    D_S = 2\dot{\epsilon}_{12}
```
The formatted doc that I refer to is here : https://cice-consortium-cice.readthedocs.io/en/latest/science_guide/sg_dynamics.html#internal-stress


> Rather than imp_solver, please spell it out as implicit_solver. ('imp' means devil or rascal or troublemaker, etc, not exactly what I'd want in the code. 

Will gladly change that, 'imp'  was not in my English vocabulary :P

> This reminds me of @njeffery's flag called 'solve_sin' which at least had a positive connotation...)
> 

Haha that made me smile :P


> Now I'm going to be a bit lazy: what is the difference between global_sum and global_sums in ice_global_reductions.F90?
> 

`global_sum` is for a single scalar, whereas `global_sums` is for several scalars, i.e. on each proc (i) I have a table [a_i,  b_i, c_i, ...] and I want to compute the distributed sums of all members of this "vector" : [\sum_i a_i, \sum_i b_i, \sum_i c_i, ...]


> If the Anderson solver isn't working yet, then why add -llapack to the macros now?

The Anderson solver is working well in serial mode, but it's not parallelized yet. I'm just adding it to the macros for my own workstation at work, and for the conda port. Adding it to the conda port does not have any downside that I can see since it's just a additional package that gets installed in the conda environment. Plus, the code is protected by a CPP so even if `-llapack` is there at link time, the linker won't link any subroutines from the library if the CPP was not activated.


> Would the set_nml.diagimp and set_nml.dynpicard and set_nml.nonlin5000 options ever be used with kdyn /= 3? Should they have kdyn=3?

`set_nml.dynpicard` does set `kdyn=3`. As for `set_nml.diagimp` and `set_nml.nonlin5000`, it's true that they only make sense with `kdyn=3`, but I prefered to create "narrow-purpose" options that can then be combined, i.e.
```bash
./cice.setup ... -s dynpicard             # default : 4 nonlin. iterations, no convergence diagnostics
./cice.setup ... -s dynpicard,diagimp     # add convergence diagnostics
./cice.setup ... -s dynpicard,nonlin5000  # max 5000 nonlin. iterations
# etc
```
I'm open to changing that however.

Thanks for this review!

dabail10:
Dumb question. Does this restart on a tripole grid? There are sometimes issues with vector halo updates and the tripole seam.

Dave

phil-blain:
@dabail10 good point! I actually have just tested it on the gx3 and gx3 grids. It is on my todo list to test it with the tx1 grid.

I will add a caveat to the documentation.

cicecore/cicedynB/dynamics/ice_dyn_eap.F90
```diff
@@ -392,6 +372,9 @@ subroutine eap (dt)
          call ice_HaloMask(halo_info_mask, halo_info, halomask)
       endif
 
+      ! velocities may have changed in dyn_prep2
+      call ice_HaloUpdate_vel(uvel, vvel, halo_info_mask)
```
apcraig, line 376: halo_info_mask is not defined unless maskhalo_dyn is true.  I find it a little awkward that halo_info_mask is passed in here and then in ice_haloUpdate_vel, halo_info_mask is only used if maskhalo_dyn is true, otherwise halo_info is used (which has been passed into the subroutine via a "use" statement).

I'm wondering if a clearer implementation might not be

```
if (maskhalo_dyn) then
  call ice_HaloUpdate_vel(uvel,vvel,halo_info_mask)
else
  call ice_HaloUpdate_vel(uvel,vvel,halo_info)
endif

```

and then remove the if check inside ice_HaloUpdate_vel.  Also, would it make sense to generalize the ice_HaloUpdate_vel so it looks more like

`call ice_HaloUpdate_vel(f1, f2, halo_info, field_loc)
`

which would make it more usable and more closely match the other HaloUpdate interfaces.  I'm also torn about whether this should be in ice_boundary or not.  If I had a choice, I'd put it in ice_boundary now and then create a separate issue that targeted improved reusability (or even merged) the mpi and serial comm directories.

configuration/scripts/options/set_env.lapack
```diff
@@ -0,0 +1 @@
+setenv ICE_CPPDEFS -DCICE_USE_LAPACK
```
apcraig, line 1: I would prefer "USE_LAPACK" as the CPP as this follows the format of some of the latest changes.

cicecore/cicedynB/infrastructure/comm/mpi/ice_global_reductions.F90
```diff
@@ -55,6 +56,12 @@ module ice_global_reductions
                       global_sum_scalar_int
    end interface
 
+   interface global_sums
```
apcraig, line 59: I too find this interface a little perplexing.  All the public interfaces in this subroutine are for fields.  This is for a vector of scalars where all scalars are added together across all pes (if I understand correctly).  I would like to suggest that a global_scalar_sum interface would be better and for the local vector sum to be passed into the global_scalar_sum.  I would also call it global_scalar_sum instead of global_sums.  Or if it remains a vector, then we should be able to clearly differentiate an interface where the sum of the vectors produces a vector vs a scalar.  I think we want to differentiate these not just thru different module procedure interfaces, but actually via clearer interface names.  It may be that we never implement the global sum of vectors to vector, but it makes sense to me to put in place a naming convention that makes it clear to do.

phil-blain:
Thanks @apcraig. I'll look into your comments in the coming days. 

I'm investigating the non b4b tests in the base_suite, trying to find a pattern.  I suspect this is due to the refactoring I did in 6af142e, but I'm not 100% sure.

I noticed all failing tests are threaded (i.e. `-p NxM`, `M`>1, except `daley_intel_smoke_gx3_4x1_diag1_run5day_thread`, but this job also sets `ICE_THREADED` to true.) 
However, some passing tests are also threaded (`./results.csh | grep -E 'PASS .*[0-9]+x[2-9]_.* compare'`). Among those, only the following have `kdyn=1` (and  `kevp_kernel` unset):
```
$ ./results.csh | \grep -E 'PASS .*[0-9]+x[2-9]_.* compare'|\grep -v boxadv |\grep -v boxdyn | \
\grep -v alt01 |\grep -v alt03 |\grep -v alt04
PASS daley_intel_smoke_gx3_1x4_debug_diag1_run2day compare c22c6d5 118.90 63.76 28.25
PASS daley_intel_restart_gx3_8x2_debug compare c22c6d5 114.49 59.09 26.38
PASS daley_intel_restart_gx3_8x2_alt02_debug_short compare c22c6d5 84.69 69.08 7.89
PASS daley_intel_smoke_gx3_4x4_alt05_debug_short compare c22c6d5 71.00 43.55 12.86
PASS daley_intel_smoke_gbox128_4x4_boxrestore_debug_short compare c22c6d5 521.69 480.85 10.54
PASS daley_intel_smoke_gx3_8x2_bgcz_debug compare c22c6d5 61.68 9.87 20.30
PASS daley_intel_smoke_gx3_4x2_debug_diag24_fsd1_run5day compare c22c6d5 165.32 86.84 45.15
PASS daley_intel_restart_gx3_4x2_debug_fsd12_short compare c22c6d5 195.67 85.97 64.70
```

and these are all compiled with debug flags. None of the failing tests are compiled in debug mode. However, some passing tests with EVP dynamics turned on are compiled in non-debug mode (`$ ./results.csh | \grep -E 'PASS .* compare'|\grep -v boxadv|\grep -v boxdyn |\grep -v alt01 |\grep -v alt03 |\grep -v alt04 | \grep -v boxslotcyl |\grep -v debug`). 

So I'm thinking of a compiler optimization that is only triggered when threading is on... but maybe my logic is wrong. 

I ran the QC test with the decomposition suggested in the doc (`-p 44x1`) and it was bit4bit. I'm re-running it again with `-p 40x4` to see if this passes QC (I'm expecting this one to not be b4b according to the above).

cicecore/cicedynB/general/ice_init.F90
phil-blain: I've pushed a new commit that adds this.

phil-blain:
I checked the unexpected missing-data test, `MISS daley_intel_smoke_gx3_4x1_diag1_run5day_thread bfbcomp daley_intel_smoke_gx3_8x2_diag1_run5day missing-data`. This test does a bit for bit compararison (`bfbcomp`) with another test, and this other test was not finished when this one ran, and so it was indeed "missing-data". I re-ran the test and it passed.

apcraig:
> This test does a bit for bit compararison (`bfbcomp`) with another test, and this other test was not finished when this one ran, and so it was indeed "missing-data". 

This happens more than I'd like.  I've tried to introduce delays and other things to try to make sure the first test is finished before other tests that need the results, but haven't gotten this to work well yet.  I may need to rethink the implementation at some point.

cicecore/cicedynB/infrastructure/comm/mpi/ice_global_reductions.F90
apcraig: One other comment.  Is the "global_sums" used for diagnostics or is it needed in the solver?  In either case, I think we need to target the ability to get "bit-for-bit" sums on different decompositions and pe counts using the bfbflag or some other way if possible.  This bit-for-bit capability is important at least for testing.  I haven't looked closely at the branch, but I'd be happy to see if we can leverage the reprosum algorithm to do this.  I think what it also implies is that the per pe values in global_sums cannot already be a local reduction of a field.

phil-blain:
> This happens more than I'd like. I've tried to introduce delays and other things to try to make sure the first test is finished before other tests that need the results, but haven't gotten this to work well yet. I may need to rethink the implementation at some point.

I think PBS has a syntax for jobs that depend on other jobs, and probably other schedulers do too. But then do we want to invest time in developing our own meta-syntax that can abstract away the implementation differences is another question...

configuration/scripts/options/set_env.lapack
phil-blain: Good idea. I've changed it (will push later.)

cicecore/cicedynB/dynamics/ice_dyn_eap.F90
phil-blain: I agree this is a little awkward. Today I've been working on refactoring that, it's not as easy as it looks.

I tried just moving the subroutine to module **`ice_boundary`**, but that creates a circular dependency between **`ice_domain`** and **`ice_boundary`** (because I have to `use ice_domain` in the subroutine to get access to `halo_info`, `maskhalo_dyn`, and `nblocks`.

So I'll try a different approach tomorrow, keeping the subroutine in **`ice_dyn_shared`** but generalizing the interface a bit, as you suggest.

phil-blain: I tried another approach in af8d03e. Since as I explained above moving the `ice_Haloupdate_vel` subroutine to **`ice_boundary`** does not appear feasible, I think that leaving it in **`ice_dyn_shared`** makes more sense. Then it is clear that it's only for the velocity, so `field_loc_NEcorner` stays hardcoded.

As I explain in the commit message of af8d03e, I chose to move the declaration of `halo_info_mask` to **`ice_dyn_shared`** as a module variable. I'm open to undoing that if we feel it's not something we want. It does simplify the interfaces in **`ice_dyn_vp`** since `halo_info_mask`  does not have to be passed down the call stack.

cicecore/cicedynB/infrastructure/comm/mpi/ice_global_reductions.F90
phil-blain: Ok I'll try to explain what `global_sums` does in a clearer way (I don't mind changing the name):
 In fact it is simply an MPI_ALL_REDUCE. Referencing the interface:

https://github.com/CICE-Consortium/CICE/blob/af8d03e42a4111ee2045f8bf5d7f7a7188637953/cicecore/cicedynB/infrastructure/comm/serial/ice_global_reductions.F90#L713-L714

Here is the computation that's taking place:
```
processes |      "vector"      |  "globalSums"
----------------------------------------------------------------------------------
p1        |  a1, b1, ..., x1   |  a1+a2+...+an, b1+b2+...+bn, ..., x1+x2+...+xn  |
----------------------------------------------------------------------------------
p2        |  a2, b2, ..., x2   |    same as above
----------------------------------------------------------------------------------
...       |    ...             |    same as above
----------------------------------------------------------------------------------
pn        |  an, bn, ..., xn   |    same as above
----------------------------------------------------------------------------------
```


I already use the `compute_dbl_sums` subroutine, so any bfbflag option would already work.
I looked at the code of  `compute_dbl_sums`  and understood that it receives a 2D array, reduces it to a 1D vector by summing along columns, and passes the resulting vector to `MPI_ALL_REDUCE`. So this is exactly what I want, and I simply create a 2D array "`work`" of shape (1,m) in `global_sums_dbl` to accomodate the interface of `compute_dbl_sums`:

https://github.com/CICE-Consortium/CICE/blob/af8d03e42a4111ee2045f8bf5d7f7a7188637953/cicecore/cicedynB/infrastructure/comm/serial/ice_global_reductions.F90#L761-L766

To answer your question, this subroutine is used in the solver if one choose classical Gram-Schmidt as the orthogonalization method.

As for naming, there are already subroutines `global_sum_scalar_{dbl,real,int}` (in the interface `global_sums`) that do a global reduction for scalar values (i.e. a single scalar value on each proc, all summed together across procs to obtain the same scalar values on all procs). So I would suggest that I rename the subroutine to `global_sum_vector_dbl` and the interface to `global_sum_vector` so that everything is clear. Would that work ?

phil-blain:
About the non bit-for-bit-ness: I did a QC test with a 160x2 configuration, and it passes. I'll gladly test other decomposition if we feel it is necessary.

cicecore/cicedynB/dynamics/ice_dyn_eap.F90
apcraig: I had a quick look at af8d03e.  I am pretty uncomfortable with halo_info_mask becoming reusable module data.  halo_info is set at initialization and is static.  halo_info_mask is very dynamic.  In fact, the correct version may vary in different parts of the code depending what you want to mask.  It certainly varies in time.  My preference continues to be that halo_info_mask be local, computed when needed locally, used locally, passed into other routines, and then destroyed.  There is too much risk someone is going to use/resuse that module data when it's going to contain stale data or nothing at all.  I also don't like the disconnect between computing halo_info_mask in one place and how delicately that interacts with the halo update in another part of the code.  

I continue to believe if we are going to want an ice_haloUpdate_vel, it should be more general (call ice_HaloUpdate_vel(uvel, vvel, halo_info, field_loc)) and in ice_boundary.F90.  If there is some circular logic issues, then maybe we need to think about moving some stuff out of ice_domain.F90 so it can be used in ice_boundary.F90.

I understand that aggregating uvel and vvel into a single variable might be part of the problem.  That's done for performance.  We can always haloUpdate each variable separately if we don't like the look of it.  All of this (creating the maskhalo and copying the two fields into one) is all done for performance.  I feel like the new implementation is creating more confusion/problems than it fixes.  I'm not sure what problem we're trying to fix.

The other thing is maybe we could create a method that creates "f2" (call copy2to1(f2,uvel,vvel)), then a call ice_haloUpdate as before with an if test on the maskhalo, then a copy1to2(f2,uvel,vvel) .  Is the problem that you don't like to have to copy in/out of f2 not in a reused subroutine. 

Having said all that, I think this can maybe be fixed by changing some names.  I think use of ice_HaloUpdate_vel and halo_info_mask are part of my problem.  Those are way too generic and are giving me heartburn.  If ice_HaloUpdate_vel were renamed to something like ice_dyn_velhalo and the module data, halo_info_mask were renamed to something like ice_dyn_velhalomask, I could handle this a lot better.  I still think ice_dyn_velhalomask should not be module data and prefer an implementation more like

```
if (maskhalo_dyn) then
  call ice_dyn_velhalo(uvel,vvel,halo_info_mask,field_loc_NEcorner)
else
  call ice_dyn_velhalo(uvel,vvel,halo_info,field_loc_NEcorner)
endif
```
which means we could keep the name halo_info_mask as it's no longer module data and just local.  We also get rid of the mask_halo_dyn check inside the "ice_HaloUpdate_vel" call.  

I guess in general, I still prefer to see a more generic interface created in this case.  If we want to continue down this path, I think at least we need to come up with names that don't feel out of place.

cicecore/cicedynB/infrastructure/comm/mpi/ice_global_reductions.F90
apcraig: Sounds fine.  I think it definitely needs a more descriptive interface name.  Looking at what we have now,

global_sum_dbl    ! sums a field across all pes to form a single scalar
global_sum_scalar_dbl  ! sums a scalar across all pes to form a single scalar

global_sum_vector_dbl might suggest that it is summing a vector across all pes to form a single scalar, but it's not.  What it's really doing is global_sum_scalar_dbl but for multiple scalars at the same time.  In other words, summing a vector across all pes to form a vector.  How about if we call it global_sum_vector_vector_dbl and put it under the generic "global_sum" interface.  Would there be a problem with uniqueness in the interface if we tried to do that?  At the same time, you could implement the global_sum_vector_dbl to acknowledge the difference?

phil-blain: My thinking is that since it performs a different kind of action, it makes sense for it to be a separate interface... that's why I'm suggesting not putting it in the generic `global_sum` (there would be no interface conflict if we do, though). Maybe `global_allreduce_sum` (this is closer to `MPI_ALLREDUCE` ) ? This way we would have : 

`global_sum` -> interface to reduce a distributed something (presently, a 2D array or a scalar) to a single scalar
`global_allreduce_sum` -> interface to reduce a distributed something (in this case a 1D vector) to something of the same shape (a 1D vector)

and then the subroutine itself could be `global_allreduce_sum_vector_dbl`. 

What do you think?

cicecore/cicedynB/dynamics/ice_dyn_eap.F90
phil-blain: > I am pretty uncomfortable with halo_info_mask becoming reusable module data.

I had a feeling you would not be comfortable with this change. In fact I was unsure myself if I liked my solution because as you say it makes it easy for callers of "ice_HaloUpdate_vel" to forget that the `halo_info_mask` has to be previously correctly created. I'll revert that change.

>  I feel like the new implementation is creating more confusion/problems than it fixes. I'm not sure what problem we're trying to fix.

My main goal was reducing code duplication.  

> I continue to believe if we are going to want an ice_haloUpdate_vel, it should be more general (call ice_HaloUpdate_vel(uvel, vvel, halo_info, field_loc)) and in ice_boundary.F90. If there is some circular logic issues, then maybe we need to think about moving some stuff out of ice_domain.F90 so it can be used in ice_boundary.F90.

OK. I'll try to work on that. The other thing that felt awkward to me when I tried this solution is that the `fld2` array used to combine uvel and vvel needs to be allocated somewhere (and I think for performance we wouldn't want to allocate it inside the subroutine each time it is called, but I could be wrong). If we move the subroutine to ice_boundary, then it makes more sense for `fld2` to be private module data in ice_boundary, but there is no existing `alloc_boundary` subroutine, so it would have to be added and all drivers updated to call it, which I wanted to avoid...

apcraig: I think there are a number of issues to get this refactoring right.  The allocation issue is a good point.  It's better if fld2 is local and we can reuse it during the loop.  I understand there could be a bit more reuse, and I'm supportive of that.  Sorry for being a pain, I just want to make sure the implementation is relatively clean if we do change it.

cicecore/cicedynB/infrastructure/comm/mpi/ice_global_reductions.F90
apcraig: I could be convinced that there should be a separate generic interface.  One other thing we could do is update the internal module interface names (not global_sum) to make them clearer.  We could come up with a subroutine naming scheme that is more general and extensible.  Maybe 

  global_sum_dbl would become global_sum_field_scalar_dbl
  global_sum_scalar_dbl would become global_sum_scalar_scalar_dbl
  
and the new method could be called

  global_sum_vector_vector_dbl

with room to add

  global_sum_vector_scalar_dbl

in the future if needed.  This should only affect the module and nothing else.  I guess we'd do it in both the serial and mpi versions.  Otherwise, the global_allreduce_sum seems OK for the generic interface name if we keep those methods separate.

cicecore/cicedynB/dynamics/ice_dyn_eap.F90
phil-blain: Yes I understand. I'll come back to this the week after next.

cicecore/cicedynB/infrastructure/comm/mpi/ice_global_reductions.F90
phil-blain: I went for a separate interface since I feel it's simplest: e30da2b. We can always revisit this in the future.

cicecore/cicedynB/dynamics/ice_dyn_eap.F90
phil-blain: I re-immersed myself in this today and came to the conclusion that I a big refactor would involve spending more time on this than I want to/can invest. So what I did was reverting the changes I made (removing `ice_HaloUpdate_vel`) and simply introduce two new subroutines, `stack_velocity_field` and `unstack_velocity_field`, like you suggested above. 

There is still a lot of copied code but at least the amount of code reuse is increased a bit.

The changes are here: 6afd6f4.

phil-blain:
There were some conflicts in the namelist, I just rebased to fix those.

phil-blain:
I reran the base suite after the rebase and I get the same failures as before (non b4b results for some non-debug, threaded cases). Just to confirm my above, hypothesis I'm rerunning the base suite with all cases in debug mode, just to see if it will be b4b this way. I'll update the PR with these results when the runs are done.

phil-blain:
So it seems that what I wanted to do is not quite possible "out-of-the-box" with our test suite infrastructure. I created a new base_suite, adding the `debug` setting:
```
./cice.setup -m daley -e intel --suite base_suite --testid base-e5ff1f2-dbg  \ 
--bgen cice.e5ff1f2.dbg --tdir ~/cice-dirs/suites/base_daley_e5ff1f2_dbg -s debug
```
and had a couple problems:

- The 2 test cases that do a bfbcomp test (`smoke_gx3_4x2_debug_diag1_run5day` and `smoke_gx3_4x1_debug_diag1_run5day_thread`) got "Missing data" for the "base_data" because the name of the case they are supposed to be compared against is hardcoded in the test suite to "`smoke_gx3_8x2_diag1_run5day`", but since I added `-s debug` the case name in the suite directory is `smoke_gx3_8x2_debug_diag1_run5day`.
- 8 tests failed to complete because they hit walltime since the model runs slower when compiled in debug mode (`-O0`).
- I also got 2 failures ("floating invalid") in the bgc tests: `daley_intel_restart_gx1_4x2_bgcsklclim_debug_medium` and `daley_intel_restart_gx1_8x1_bgczclim_debug_medium`. This is more concerning and maybe warrants a separate issue.

---

@eclare108213, @JFLemieux73, @apcraig is the testing I've done sufficient for this PR ? Do you think of additional test that I'd need to run before we feel comfortable merging ?

phil-blain:
I just pushed two commits (e96bdda and 2fb99d1) for a change that was suggested by @JFLemieux73 offline (a little refactoring and renaming of the way the "init" subroutines for the dynamics are called in the drivers).

phil-blain:
I've completed my run of the base_suite (baseline+test) in debug mode. **All test cases that failed the regression test in the regular base_suite now are b4b with the baseline when compiled in debug mode.**

apcraig:
I will wait for @eclare108213 to approve.  Unless something else comes up, it sounds like this is ready to merge.  I will probably wait until next week to merge just thinking about weekend testing and our recent release.  I also want to have time to manually run master testing after the merge just to generate some new baselines since answers change.  Thanks @phil-blain

apcraig:
@eclare108213, are we ready to merge this PR?  thanks.

eclare108213:
If you understand the failures and why they are BFB in debug mode, then I think this is ready to merge.  If it's due to refactoring, then I suggest identifying the affected lines in this PR, if possible, or at least describe what changed and why.  (maybe you did already, up above - is it because the refactored code is BFB until optimization is done?  do you know which loops etc?)

apcraig:
@phil-blain, do you have a pretty good idea what aspect of the refactoring/new code is changing the results?  If so, could you summarize in this conversation or in the PR checklist section.  If not, just let us know, and I think we can merge.  Thanks.

phil-blain:
@eclare108213, @apcraig :

I'm still not 100% sure why the code is not BFB with threading and optimization turned on... I went through the output of `git diff --stat upstream/master...` on this branch to see which files (outside `ice_dyn_vp.F90`) I touched, and I still think that the only culprit that makes sense is this commit: https://github.com/CICE-Consortium/CICE/commit/6af142ef23e19b2b773f52f1607f1bb3cb223e76. However there is no OpenMP in the code being changed in that commit, so I'm not convinced...

phil-blain:
I've updated the PR description with the link to the message (https://github.com/CICE-Consortium/CICE/pull/491#issuecomment-663226841) where I explain my reasoning about the non BFB cases being those with threading and optimization turned on.

apcraig:
thanks @phil-blain, I don't think any of this is unexpected.  It can be very difficult to pinpoint compiler optimization changes.  I've found cases where a code change results in answer changes somewhat away from where the code change took place.  I think the fact that it's bit-for-bit for all tests when they are all run in debug (no optimization) is a really solid result and about as much as can be done in these situations.  I will merge now.

apcraig:
I ran full test suites of the current master (after this commit) on cheyenne and onyx and there is one test that is failing on both machines on all compilers, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash#b0a7ee6a85ba2283ff2d12ac20c01a92595644e4.  The test "restart gx3 8x2x8x10x20 droundrobin maskhalo" runs but fails exact restart.  This seems to be a robust result.  I am doing some additional testing now to see if I can learn any additional information, but we'll want to figure it out and fix it.

apcraig:
I can reproduce the failure.  I have confirmed that it diverges on the first timestep.  It's the only restart test with padded blocks with the maskhalo on.  That may be a clue or it may not be.  I'll see if I can make some more progress.

phil-blain:
Thanks for looking into that. Tell me if I can be if any help.

phil-blain:
I can at least confirm that the same test also fails on our XC-50 machine (`daley`).

apcraig:
#517 addresses the latest bug.  An unmasked halo update was recoded to a masked halo update and this seems to have created the issue.  This was fixed in evp, eap, and vp implementations.

apcraig:
One other comment.  I did run the problematic case (padded blocks with maskhalo) with kdyn=3, and it aborted with a "bad departure points" error.  I suspect there are edge configurations that we are testing with evp and/or eap that have not yet been tested with vp.  It might be good to run a full test suite (all suites) with kdyn=3 to see what happens or to otherwise expand/add the test coverage with kdyn=3.  It raises a questions about how to best test big new options.  Running the current (even full) test suite may not be adequate in general.  For something like this, kdyn=3 probably needed to be run under many different configurations (maybe technical and science) and at least a few new test cases probably need to be added.  And this highlights again the question about how comprehensive we are in our current testing.  We test kdyn=1 (evp) with lots of configurations (like padded blocks), but we don't do that with kdyn=2 (eap).  We have identified some missing tests now (testing with additional decomps with eap and vp), I will try to create a new PR to add those tests as well.

apcraig:
I ran a decomp test suite with evp, eap, and vp-picard.  I did not test vp-anderson as that is not ready to use out of the box.  Below are the results.  evp and eap pass all tests with the various decomps.  However, vp-picard is more of a mixed bag.  Some configurations don't run at all and some don't restart exactly.  The test suite is running the same tests just changing the dynamics option.  I will create a new issue and that issue can be closed if it's not of interest.  I also understand that I may not have setup the tests properly or that vp-picard may still be under development, but this gives us a data point.  Looking at one of the failed runs, I see

 ```
(abort_ice)ABORTED: 
 (abort_ice) error = (horizontal_remap)ERROR: bad departure points
```

```
PASS cheyenne_intel_restart_gx3_4x2x25x29x4_dslenderX2 run 10.46 2.72 5.35
PASS cheyenne_intel_restart_gx3_4x2x25x29x4_dslenderX2 test 
PASS cheyenne_intel_restart_gx3_1x1x50x58x4_droundrobin_thread run 38.01 8.49 21.03
PASS cheyenne_intel_restart_gx3_1x1x50x58x4_droundrobin_thread test 
PASS cheyenne_intel_restart_gx3_4x1x25x116x1_dslenderX1_thread run 12.50 2.70 6.91
PASS cheyenne_intel_restart_gx3_4x1x25x116x1_dslenderX1_thread test 
PASS cheyenne_intel_restart_gx3_6x2x4x29x18_dspacecurve run 12.03 4.06 4.37
PASS cheyenne_intel_restart_gx3_6x2x4x29x18_dspacecurve test 
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin run 7.95 2.58 2.51
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin test 
PASS cheyenne_intel_restart_gx3_6x2x50x58x1_droundrobin run 11.20 2.46 6.08
PASS cheyenne_intel_restart_gx3_6x2x50x58x1_droundrobin test 
PASS cheyenne_intel_restart_gx3_4x2x19x19x10_droundrobin run 11.25 3.11 4.81
PASS cheyenne_intel_restart_gx3_4x2x19x19x10_droundrobin test 
PASS cheyenne_intel_restart_gx3_1x20x5x29x80_dsectrobin_short run 30.55 13.72 7.23
PASS cheyenne_intel_restart_gx3_1x20x5x29x80_dsectrobin_short test 
PASS cheyenne_intel_restart_gx3_16x2x5x10x20_drakeX2 run 4.87 1.67 1.78
PASS cheyenne_intel_restart_gx3_16x2x5x10x20_drakeX2 test 
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_maskhalo run 7.32 2.31 2.42
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_maskhalo test 
PASS cheyenne_intel_restart_gx3_1x4x25x29x16_droundrobin run 28.46 9.31 17.47
PASS cheyenne_intel_restart_gx3_1x4x25x29x16_droundrobin test 

PASS cheyenne_intel_restart_gx3_4x2x25x29x4_dslenderX2_dyneap run 32.14 24.53 5.23
PASS cheyenne_intel_restart_gx3_4x2x25x29x4_dslenderX2_dyneap test 
PASS cheyenne_intel_restart_gx3_1x1x50x58x4_droundrobin_dyneap_thread run 110.39 80.04 21.20
PASS cheyenne_intel_restart_gx3_1x1x50x58x4_droundrobin_dyneap_thread test 
PASS cheyenne_intel_restart_gx3_4x1x25x116x1_dslenderX1_dyneap_thread run 36.50 26.83 6.77
PASS cheyenne_intel_restart_gx3_4x1x25x116x1_dslenderX1_dyneap_thread test 
PASS cheyenne_intel_restart_gx3_6x2x4x29x18_dspacecurve_dyneap run 46.75 38.86 4.30
PASS cheyenne_intel_restart_gx3_6x2x4x29x18_dspacecurve_dyneap test 
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dyneap run 24.54 19.22 2.75
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dyneap test 
PASS cheyenne_intel_restart_gx3_6x2x50x58x1_droundrobin_dyneap run 33.14 24.53 5.95
PASS cheyenne_intel_restart_gx3_6x2x50x58x1_droundrobin_dyneap test 
PASS cheyenne_intel_restart_gx3_4x2x19x19x10_droundrobin_dyneap run 36.26 28.27 4.64
PASS cheyenne_intel_restart_gx3_4x2x19x19x10_droundrobin_dyneap test 
PASS cheyenne_intel_restart_gx3_1x20x5x29x80_dsectrobin_dyneap_short run 116.90 100.69 7.68
PASS cheyenne_intel_restart_gx3_1x20x5x29x80_dsectrobin_dyneap_short test 
PASS cheyenne_intel_restart_gx3_16x2x5x10x20_drakeX2_dyneap run 14.25 11.05 1.84
PASS cheyenne_intel_restart_gx3_16x2x5x10x20_drakeX2_dyneap test 
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dyneap_maskhalo run 23.86 18.80 2.37
PASS cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dyneap_maskhalo test 
PASS cheyenne_intel_restart_gx3_1x4x25x29x16_droundrobin_dyneap run 102.08 82.70 38.28
PASS cheyenne_intel_restart_gx3_1x4x25x29x16_droundrobin_dyneap test 

PASS cheyenne_intel_restart_gx3_4x2x25x29x4_dslenderX2_dynpicard run 15.13 7.46 5.30
FAIL cheyenne_intel_restart_gx3_4x2x25x29x4_dslenderX2_dynpicard test 
PASS cheyenne_intel_restart_gx3_1x1x50x58x4_droundrobin_dynpicard_thread run 47.71 18.42 20.95
PASS cheyenne_intel_restart_gx3_1x1x50x58x4_droundrobin_dynpicard_thread test 
PASS cheyenne_intel_restart_gx3_4x1x25x116x1_dslenderX1_dynpicard_thread run 15.70 6.00 6.92
PASS cheyenne_intel_restart_gx3_4x1x25x116x1_dslenderX1_dynpicard_thread test 
FAIL cheyenne_intel_restart_gx3_6x2x4x29x18_dspacecurve_dynpicard run
FAIL cheyenne_intel_restart_gx3_6x2x4x29x18_dspacecurve_dynpicard test 
FAIL cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dynpicard run
FAIL cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dynpicard test 
PASS cheyenne_intel_restart_gx3_6x2x50x58x1_droundrobin_dynpicard run 14.13 8.56 6.04
PASS cheyenne_intel_restart_gx3_6x2x50x58x1_droundrobin_dynpicard test 
FAIL cheyenne_intel_restart_gx3_4x2x19x19x10_droundrobin_dynpicard run 
FAIL cheyenne_intel_restart_gx3_4x2x19x19x10_droundrobin_dynpicard test 
FAIL cheyenne_intel_restart_gx3_1x20x5x29x80_dsectrobin_dynpicard_short run
FAIL cheyenne_intel_restart_gx3_1x20x5x29x80_dsectrobin_dynpicard_short test 
PASS cheyenne_intel_restart_gx3_16x2x5x10x20_drakeX2_dynpicard run 15.44 12.28 2.09
FAIL cheyenne_intel_restart_gx3_16x2x5x10x20_drakeX2_dynpicard test 
FAIL cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dynpicard_maskhalo run
FAIL cheyenne_intel_restart_gx3_8x2x8x10x20_droundrobin_dynpicard_maskhalo test 
FAIL cheyenne_intel_restart_gx3_1x4x25x29x16_droundrobin_dynpicard run
FAIL cheyenne_intel_restart_gx3_1x4x25x29x16_droundrobin_dynpicard test 
```

phil-blain:
You are right that I did limited testing with the various options. It's on my list to do more.

PR #496, opened 2020-07-29
Remove "coupled" CPP flag, add ssh_stress namelist

apcraig:
## PR checklist
    Remove "coupled" CPP flag, add ssh_stress namelist
- [X] Developer(s): 
    apcraig
    Passed standalone test suite on gordon, 4 compilers, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#b055c7f01ad4ad8f5d946d4456188a8f8b8eb55d.  Coupled testing will have to be carried out by different groups.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

The main issue was to remove the coupled flag from the non-driver part of the code.  This was done by adding the ssh_stress namelist and removing the "fake MPI" stuff from the serial ice_communicate file.  For the standalone CICE driver, the coupled and CESMCOUPLED Macros were removed as if they were never set.  This is correct for the standalone model.  For other drivers, CPPs were removed where it seemed safe and old CPP code was left in but commented out.  Each coupled system should test their updated driver and do further cleanup as needed.

- Remove "coupled" CPP flag
  - Remove "fake MPI" capability in serial/ice_communicate.F90
  - Add ssh_stress variable/namelist with options 'geostrophic' or 'slope' to specify the computation of the sea surface height stress term.
  - Add ssh_stress to ice_in, set to 'geostrophic'.
  - Add ssh_stress to namelist documentation
  - Change "coupled" to CESMCOUPLED in ice_init on check for oceanmixed_ice setting
  - Clean up drivers
    - remove "coupled" and CESMCOUPLED from standalone/cice as if neither were set
    - remove "coupled" macro end_run call in nuopc/cmeps, CESMCOUPLED left mostly alone
    - remove "coupled" and CESMCOUPLED from mct/cesm1 as if both were set
    - remove "coupled" and CESMCOUPLED from direct/hadgem as if both were set
    - no changes to nuopc/dmi, prefer dmi make those changes
- Fix error in coriolis namelist documentation where Cstar was in the wrong place

NOTE: Coupled systems should leverage the new ssh_stress namelist to turn on the 'slope' option.  The coupled CPP flag is no longer in use.


apcraig:
I have implemented the new namelist 'ssh_stress' with options 'geostrophic' and 'slope'.  I am happy to change these names or anything about how they're implemented after review.

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -341,6 +341,8 @@ dynamics_nml
    "``ndte``", "integer", "number of EVP subcycles", "120"
    "``Pstar``", "real", "constant in Hibler strength formula (N/m\ :math:`^2`)", "2.75e4"
    "``revised_evp``", "logical", "use revised EVP formulation", "``.false.``"
+   "``ssh_stress``", "``geostropic``", "computed from ocean velocity", "``geostrophic``"
+   "", "``slope``", "computed from coupled sea surface height tilt", ""
```
eclare108213, line 345: I'd use the term "gradient" rather than tilt here

dabail10:
Or maybe 'dynamic'?

eclare108213:
geostrophic is dynamic too

dabail10:
Haha. True that. Maybe 'prognostic'.

eclare108213:
geostrophic is prognostic too...

eclare108213:
it's just a different approximation, a simple parameterization versus a full ocean model

dabail10:
Not exactly. The geostrophic approximation is effectively diagnostic from the velocities. The full sea surface height terms are computed prognostically in the ocean component.

eclare108213:
yeah, okay, that's a pretty subtle distinction.  I'd still prefer something more intuitive, to indicate that the terms are coming directly from the ocean model.  "Prognostic" to me indicates that the sea ice model is producing them, as opposed to their appearing as input data.

dabail10:
Maybe, 'actual' then? I was thinking 'gradient' too, but the problem is that the "gradient" winds or current have a definition that is just a step above geostrophic. Funny how such a small detail can generate this much discussion.

apcraig:
Is there a decision about what the new namelist variable name and possible values should be?

eclare108213:
Not really, we're still throwing out ideas.  'actual' isn't true (this is a simulation, not reality).  How about 'coupled-ocean'?

apcraig:
I'm good with 'coupled-ocean'.  I will also change "tilt" to "gradient" in the documentation.  I went thru several iterations trying to decide what might work with both the namelist and documentation.  I do think it's worth the effort to discuss and choose good names because these are things that will be with us a long time in most cases.  Thanks!

dabail10:
Haha. Or 'coupled', since we are removing the ifdef ... :)

apcraig:
"coupled" would be a good choice too.  I guess I would ask if it's possible we'll add other options in the future.  If so, it would be good to think about what those might look like and to pick names now that allow for some extensibility.  If ultimately we have just two options, it would also be possible to make this a logical flag.  I'm not advocating that, I think the char string thing works well.  But just saying.

eclare108213:
I think it's better to go with the character string in case some other option comes up - I don't know what it would be.

dabail10:
I am fine with 'coupled-ocean' or 'coupled'.

apcraig:
I just committed a change to use "coupled" instead of "slope".  I ran a quick suite on gordon with 4 compilers to make sure it was working and all looks fine.  Happy to update again if we want to continue to discuss or further changes are requested.

TillRasmussen:
Maybe the warning could be changed to an abort if both coupled and the oceanmixed model is turned on.

apcraig:
Regarding the check on "coupled" and "oceanmixed_ice"; I suspect there are 100s of other checks that could/should be implemented but that don't exist.  That doesn't mean this check should be removed, but I assume this was added by someone at some point because they set something up and didn't catch it.   Having said that, I would be happy to change the logic to 

```
 if (trim(ssh_stress) == 'coupled' .and. oceanmixed_ice) then
   ...
   abort
 endif
```

Or is a warning message really what we want here?  I suspect the warning message is not going to be very helpful in practice.  Thoughts?

eclare108213:
I agree we aren't going to be able to catch every mistake, we can only try to head them off.  In this case, the warning was just supposed to be helpful, if a user couldn't figure out why their results looked weird and so they looked at the diagnostic output.  I was probably the one who got caught by this, when coupling POP and CICE long ago!  I guess my inclination would be to make sure this is well documented:  say somewhere that if on, the mixed layer model will overwrite information coming in from the ocean (SST, frzmlt, maybe other things).  We could have a section explicitly for coupled modeling, saying things like "make sure you have this namelist flag (not) set as..."

eclare108213:
In fact, you could keep the warning in the code as it is, but without the CESMCOUPLED cpp, so that it writes a slightly more generic warning any time that the ocean mixed layer is turned on.  Hopefully, most of the time people will be running coupled without the mixed layer model, and so it wouldn't print anything.

apcraig:
I do like that idea.

apcraig:
I update the warning message to the following,

```
         if (oceanmixed_ice) then
            write(nu_diag,*) '     WARNING: ocean mixed layer ON'
            write(nu_diag,*) '     WARNING: will impact ocean forcing interaction'
            write(nu_diag,*) '     WARNING: coupled forcing will be modified by mixed layer routine'
         endif
```

If anyone prefers different words or format, just let me know.

eclare108213:
looks fine to me

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -314,9 +314,9 @@ dynamics_nml
    "``basalstress``", "logical", "use basal stress parameterization for landfast ice", "``.false.``"
    "``Cf``", "real", "ratio of ridging work to PE change in ridging", "17.0"
    "``coriolis``", "``constant``", "constant coriolis value = 1.46e-4", "``latitude``"
-   "``Cstar``", "real", "constant in Hibler strength formula", "20"
    "", "``latitude``", "coriolis variable by latitude", ""
    "", "``zero``", "zero coriolis", ""
+   "``Cstar``", "real", "constant in Hibler strength formula", "20"
```
phil-blain, line 319: Thanks for fixing that, I noticed recently and then forgotten to mention it!

PR #507, opened 2020-08-18
Moved calculation of sicen and trcrn_bgc to the loop where they are used

TillRasmussen:
Moved calculation of sicen and trcrn_bgc to the loop where they are used. The current construction did not use the calculated values as they were defined private and overwritten at each i/j

## PR checklist
    Resolve warning 
- [ ] Developer(s): 
tar
@apcraig @dabail10 
   This is a bug fix of the init_bgc. I have not been running a test with bgc but the code compiles on gfortran, intel and cray. It will change the initialization of bgc.
 I hope that @dabail10 will review that the code work as intended. I have one concern this is that the initial calculation was imbedded in a "if (.not. restart_bgc)" clause. This is not the case anymore, however it is only used here.
    - [ ] bit for bit
    - [x ] more substantial (anticipated)
    - [ ] Yes
    - [ x] No
    - [ ] Yes
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [ ] No 

### DIFF
--- a/cicecore/shared/ice_init_column.F90
+++ b/cicecore/shared/ice_init_column.F90
@@ -877,7 +877,7 @@
 
       endif     ! .not. restart
 
-      !$OMP PARALLEL DO PRIVATE(iblk,i,j,k,n,ilo,ihi,jlo,jhi,this_block,sicen,trcrn_bgc)
+      !$OMP PARALLEL DO PRIVATE(iblk,i,j,n,ilo,ihi,jlo,jhi,this_block)
       do iblk = 1, nblocks
 
          this_block = get_block(blocks_ice(iblk),iblk)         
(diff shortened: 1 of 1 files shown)


apcraig:
Do we want to do a qc test with bgc?  It looks like this could have significant impacts on the initial condition.  Does the initial condition change when using a restart file?  We should probably check exact restart as well.

TillRasmussen:
> Do we want to do a qc test with bgc? It looks like this could have significant impacts on the initial condition. Does the initial condition change when using a restart file? We should probably check exact restart as well.

This will impact the initial condition as the value of sicen and trcrn_bgc were undefined in the previous version.
If  a restart file is used it is read afterwards (line 953), thus I dont think that this will change the result.

apcraig:
I ran a full test suite on conrad with this change and all results including the bgc configurations are bit-for-bit.  Looking at a bgcz test, the initial condition seems to be "runtype=initial, ice_ic=none".  So, just for further clarification, is this what we expect?

eclare108213:
If the BGC is being started from a no-ice condition, then I wouldn't expect the answers to change.  It sounds like the modifications here only affect initializations from a state with existing ice, which aren't restarting BGC (restarts should overwrite the modified initialization).  I don't think (but we should check) that we have test cases for BGC starting from existing ice, mainly because of the headaches associated with getting the initial conditions correct - @njeffery and I have generally always started from no-ice or from restarts with the BGC-relevant info in them, to let things evolve naturally.  @njeffery please correct me if I'm wrong about that...

njeffery:
There is an option to activate BGC from a spin-up run with no BGC. Set restart_bgc = false and scale_bgc = true.  This option assumes that initial concentrations of nutrients in existing sea ice are proportional to the salinity profile and scaled by the ocean SSS divided by the respective ocean nutrient.  I'd still disregard the first 10 years or so of the output.

apcraig:

eclare108213:
It would be good to add a test for the case that @njeffery describes above, but that does not have to be done for this PR.  Better to get the bug fixed now.

Issue #267, opened 2018-12-03
tracers in upwind transport


eclare108213:
Upwind transport should use the tracer dependency arrays instead of doing calculations explicitly in work_to_state and state_to_work.  Should we deprecate the upwind scheme?

dabail10:
Does anyone use upwind? We should probably do a survey before getting rid of it.

eclare108213:
This could be a discussion topic during the town hall -- whether to deprecate some things, and if so, what.  Upwind and 0-layer thermo come immediately to mind.  A general survey is an interesting idea.  I suspect that there will always be someone who thinks we should keep things in the code.  E.g. last fall one of our German colleagues strongly urged me not to get rid of the 0-layer thermo.

JFLemieux73:
I know I said before we should deprecate upwind but I have changed my mind. I am involved in an intercomparison project where we do idealized experiments (similar to box2001). My colleagues just asked me to do a run with upwind...I think it would be a good idea to keep for this kind of idealized experiments.

eclare108213:
@JFLemieux73 we can certainly keep the upwind functionality, but the issue with the tracers remains.  I suggest that we revert the changes in #508 and make sure there is an abort if there are any tracers other than those that upwind is able to advect (e.g. by checking that ntrcr is not larger than the number of required tracers).  I've updated the [code deprecation list](https://github.com/CICE-Consortium/CICE/wiki/Deprecated-Code-Features).

apcraig:
I will create a PR to revert #508 and to add a check on ntrcr.  Can someone clarify what that check should be?  I will also try to figure it out.

eclare108213:
Thank you @apcraig. I was thinking that we could just check that ntrcr is less than or equal to the number of required tracers.  But looking at the upwind advection routines, it's more complicated than that.  The issue lies in the routines state_to_work and work_to_state, which load and unload arrays to be advected.  It looks like work_to_state is probably okay. In state_to_work, the question is whether all of the potential tracers are captured in the manually loaded works array.  I don't think so, although many of them are.  There is this code check at the end of that routine:
`      if (narr /= narrays) write(nu_diag,*)      &
           "Wrong number of arrays in transport bound call"`
which hopefully then aborts if something is missing.  So maybe this is okay for now, although I'd still prefer that this code be rewritten so that works is not manually loaded.  If the narr check is working, we can revert the deprecated code and the re-writing can continue to wait.

apcraig:
[...]

PR #508, opened 2020-08-21
deprecate upwind advection

eclare108213:
## PR checklist
First phase of code deprecation for upwind advection, following https://github.com/CICE-Consortium/CICE/wiki/Deprecated-Code-Features 
- [x] Developer(s): 
    E. Hunke

https://github.com/CICE-Consortium/Test-Results/wiki/eb77517470.badger.intel.20-08-21.210840.0

> 243 measured results of 243 total results
> 237 of 243 tests PASSED
> 0 of 243 tests PENDING
> 0 of 243 tests MISSING data
> 6 of 243 tests FAILED
> 
> [eclare@ba-fe1 testsuite.dep3]$ results.csh | grep FAIL
> FAIL badger_intel_restart_gx1_40x4_droundrobin_medium run 
> FAIL badger_intel_restart_gx1_40x4_droundrobin_medium test 
> FAIL badger_intel_restart_tx1_40x4_dsectrobin_medium run
> FAIL badger_intel_restart_tx1_40x4_dsectrobin_medium test 
> FAIL badger_intel_restart_gx3_4x4_alt04 compare bs_20200821 8.37 2.23 6.37 different-data
> FAIL badger_intel_smoke_gx3_4x4_alt04_debug_short compare bs_20200821 11.05 4.94 3.18 different-data

Four failures are the usual 40-processor fails on badger, the other two are because the alt04 test changed to use remap instead of upwind.

    - [x] bit for bit
    - [x] more substantial 
This question is harder to answer because we are removing a capability.  In that sense, answer changes are indeed substantial, but this does not change the answers for the rest of the code.
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No, but alt04 changes.
    - [x] Yes
        - [ ] Yes
        - [ ] No 

Commented out most of the upwind related code, flagging it with ``!deprecate upwind`` to make it easier to find and remove later.  Changed subroutine names that aren’t completely commented out by adding the suffix ``_deprecated``, to catch instances where they might be called from elsewhere in the code. 

Changed the advection parameterization in alt04 to ``remap``.  This changes the answers for that test only, as expected.

Updated readthedocs documentation to its final version, since commenting out rst with identifying flags doesn’t appear to be obvious for later removal during the final deprecation stage.  If we need to reinstate the deprecated code instead of removing it completely, we can resurrect the documentation from earlier versions.

I'd like feedback on whether this process is the way we want to continue deprecating code, or if there are better ways than what's done here.


eclare108213:
If upwind were used in any run (coupled or not) that executes the input_data subroutine, the code should abort because it now only allows 'remap' and 'none' as options for advection.  However I added an additional error message if advection='upwind', to warn that it has been deprecated.  If a coupled code does not execute input_data and doesn't update its own interface appropriately for the deprecation changes, then the code would probably fail to compile because the upwind subroutines are not available.  This is part of the reason for doing this in two stages, in case a user group is using those subroutines and would prefer that we not deprecate them but wasn't aware of it - this way we can reinstate them fairly easily.

I ran a test using advection='upwind', to make sure it did abort, and here's the output:

>  (abort_ice)ABORTED: 
>  (abort_ice) called from ice_init.F90
>  (abort_ice) line number         1640
>  (abort_ice) error = (input_data) ABORTING on input ERRORS
> (input_data) ERROR: abort_list = :3

It did not print the full error message, either the original one or the 'upwind deprecated' message that I just added. 
@apcraig is this what's expected from the warning system now?  Here is the code block from ice_init.F90:

```
!deprecate upwind      if (advection /= 'remap' .and. advection /= 'upwind' .and. advection /= 'none') then                       
      if (advection /= 'remap' .and. advection /= 'none') then
         if (trim(advection) == 'upwind') then
            if (my_task == master_task) write(nu_diag,*) subname//' ERROR: upwind advection has been deprecated'
         endif
         if (my_task == master_task) write(nu_diag,*) subname//' ERROR: invalid advection=',trim(advection)
         abort_list = trim(abort_list)//":3"
      endif


```

apcraig:
There must be some additional error messages before the "ABORTED:" line?

eclare108213:
Yes!  All the way at the top of the file - sorry I just skipped to the end to make sure it aborted.  There's a lot of other diagnostic stuff in between the warning lines 
```
(input_data) ERROR: upwind advection has been deprecated
(input_data) ERROR: invalid advection=upwind
```

and the abort message.

apcraig:
Good.  So what's happening is that the input_data routine is trapping errors but not aborting until the end of the routine.  That way, you can see all the errors rather than having to run multiple times to trap all the errors.  The "abort_list" indicates which errors were triggered.  In this case it was just "3".  There are about 30 right now if I remember.  When an error is trapped, the error message is written out and then other diagnostic messages can be written in between before the ABORT message.  Maybe I should fix this so each error trap produces an error message that can also be written when the run is stopped.  That's probably a valuable feature to have, but is a separate issue from this.

apcraig:

eclare108213:
@apcraig 
@JFLemieux73 has requested that we keep upwind in the code, which is fine with me as long as we make sure users don't expect the tracers to work (see #267).  So far this PR only commented things out.  Here's a git question:  it's possible to revert a PR, but it looks like this will happen directly in master rather than on a fork/branch, which would break our workflow.  Is there an elegant way to do this, or should we revert the changes by hand?

phil-blain:
@eclare108213 about reverting: you seem to be talking about the "Revert" button on GitHub. From my account, the hover text for this button on this here PR shows "Create a new pull request to revert these changes". Maybe because you have write access to the repo this option is not the same for you and in your case it would directly revert on master. 

In any case, anybody can create a local branch, use ` git revert 164fcfc` and open a PR for that branch, following our normal workflow.

eclare108213:
Thanks @phil-blain, that's what I was wondering.  I looked at the online documentation for git revert, which wasn't clear.

apcraig:
@eclare108213, do you want me to create a PR to undo #508 and also add a check as noted in #267 on ntrcr?

apcraig:
I'll take care of the PR to revert this.  Lets move discussion to #267

PR #510, opened 2020-08-25
doc: correct description of box2001 test

phil-blain:
## PR checklist
doc: correct description of box2001 test
- [X] Developer(s): 
        P Blain.
    **no tests; doc fix.**
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

    The documentation states that this test case sets 'coriolis' to zero,
    but 'configuration/scripts/options/set_nml.box2001' sets 'coriolis' to 'constant'.
    Fix the documentation to be in line with the code.

### DIFF
--- a/doc/source/user_guide/ug_case_settings.rst
+++ b/doc/source/user_guide/ug_case_settings.rst
@@ -364,7 +364,7 @@ dynamics_nml
    "``brlx``", "real", "revised_evp value", "300.0"
    "``basalstress``", "logical", "use basal stress parameterization for landfast ice", "``.false.``"
    "``Cf``", "real", "ratio of ridging work to PE change in ridging", "17.0"
-   "``coriolis``", "``constant``", "constant coriolis value = 1.46e-4", "``latitude``"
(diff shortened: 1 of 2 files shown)


phil-blain:
@JFLemieux73 good idea. I've added it and also in the Case settings section.

apcraig:

JFLemieux73:
In Hunke 2001 it is written: For these simulations the Coriolis parameter is taken to be constant...
In the table: f Coriolis parameter 1.46 x 10-4 s-1

We could set f to 0 as long as it is consistent with the doc. I think the 3 islands are not there either (am I right?). We could just say this test case is inspired by the test case in Hunke 2001.

eclare108213:
Interesting.  Thanks for checking!  Then let's keep it as the constant value.  You're right, the islands aren't there, but they aren't the important part of the test - that's getting the forcing the same.  So I think this is fine.  Thank you!

PR #517, opened 2020-09-25
Fix maskhalo bug in dynamics

apcraig:
PR checklist
    Fix maskhalo bug in dynamics introduced in #491
- [X] Developer(s): 
    apcraig
    Full test suite passed on cheyenne, this fixes issue with "restart gx3 8x2x8x10x20 droundrobin maskhalo" on the current trunk.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#b1b1f9d49b3a228cff3fc9b07ac4a22d13f2da39
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Fix maskhalo bug in dynamics, only appears when running padded decompositions.  In #491, an unmasked halo update was changed to a masked halo update.  This affects only padded decompositions with maskhalo_dyn=true and was picked up by an exact restart failure.

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_eap.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_eap.F90
@@ -365,6 +365,11 @@
       call ice_timer_start(timer_bound)
       call ice_HaloUpdate (strength,           halo_info, &
                            field_loc_center,   field_type_scalar)
+      ! velocities may have changed in dyn_prep2
+      call stack_velocity_field(uvel, vvel, fld2)
+      call ice_HaloUpdate (fld2,               halo_info, &
+                           field_loc_NEcorner, field_type_vector)
+      call unstack_velocity_field(fld2, uvel, vvel)
       call ice_timer_stop(timer_bound)
 
       if (maskhalo_dyn) then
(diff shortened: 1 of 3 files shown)


apcraig:
It would be nice to get this reviewed and approved today to get it into weekend testing.  But this bug is unlikely to affect anyone in the short term, so it can also wait if needed.

dabail10 (approved):

phil-blain:
Thanks @apcraig. This looks like the right solution, but I'm not sure I understand why the change from an unmasked to a masked halo at the point in code would cause exact restart to fail... Can you explain a little more ? just so I understand the code better. 

The original commit where I made this change was: e85bca8ade2abefa9de732effbe77cc940457ec8, and I remember thinking "oh, the masked haloupdate is not done there, with my change it will be done but I guess that's an optimization..." so something was clearly missing from my understanding.

apcraig:
@phil-blain, I was sort of asking myself the same question when I identified the difference.  I can't remember the details about why that first halo update should not be masked.  It's obviously a subtle issue that seems to only create problems is very special cases.  The other thing is that haloupdate is not part of the subcycling (in evp and eap anyway), so the cost savings is relatively small.  If we want to dig deeper into this, we could.  It would probably require comparing various fields within the subroutine as the model advances to see where and how the differences are introduced.  I'm not sure it's worth the effort right now, but am open to raising the priority and pursuing this question.  

When I implemented the maskhalo feature, I tried to identify the proper mask and test for bit-for-bit focusing especially where a mask could be reused for multiple halo updates.  Since this was a single halo update outside the dynamics subcycling, the cost benefit was relatively small.  I suspect I recognized the masked halo was a problem in this haloupdate, but I may not have ever tried to understand why because the performance implication was small.  It may be as simple as forcing zeros into the unused halo to ensure it behaves properly, but that is just speculation.  Again, we could look into this more if there was interest.

phil-blain:
OK I understand. Let's not pursue this too far, we can always revisit if need be, "premature optimization is the root of all evil" after all.

eclare108213:
I don't completely understand this either, but here's a hypothesis:
If it's only happening for padded runs where there are extra grid cells outside of the physical domain, beyond the halos, and the code in question is only shifted from before the maskhalo update to afterward, then the halo masking must not treat the padded cells the same way that the regular halo update does.  This could affect restarts when the entire grid is written to the restart file (as opposed to just the physical domain).  I'm not sure if that's the case here, and this is just a hypothesis.  But the fix looks fine to me, and so I'll approve it for weekend testing.

eclare108213 (approved):

PR #526, opened 2020-10-27
box2001: compute aiu with aice instead of aice_init

JFLemieux73:
## PR checklist
    Correcting my bug...because get_forcing_atmo is called after ice_step, aiu in get_forcing_atmo (for box2001) should be calculated with aice instead of aice_init (that way the aiu for aiu*tair and aiu*twater are at the same time level).

This does not change the result with the box2001 exp because the advection is turned off. It only matters when someone is running modified box2001 exp with advection turned on (like I am doing...). 

- [ ] Developer(s): 
    @JFLemieux73 
@eclare108213 @TillRasmussen 
    249 measured results of 249 total results
249 of 249 tests PASSED
0 of 249 tests PENDING
0 of 249 tests MISSING data
0 of 249 tests FAILED
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 

### DIFF
--- a/cicecore/cicedynB/general/ice_forcing.F90
+++ b/cicecore/cicedynB/general/ice_forcing.F90
@@ -5067,7 +5067,7 @@
       use ice_blocks, only: nx_block, ny_block, nghost
       use ice_flux, only: uocn, vocn, uatm, vatm, wind, rhoa, strax, stray
       use ice_grid, only: uvm, to_ugrid
-      use ice_state, only: aice_init
+      use ice_state, only: aice
 
       ! local parameters
 
@@ -5082,7 +5082,7 @@
       call icepack_query_parameters(pi_out=pi, pi2_out=pi2, puny_out=puny)
       call icepack_query_parameters(secday_out=secday)
 
-      call to_ugrid(aice_init, aiu)
+      call to_ugrid(aice, aiu)
 
       period = c4*secday
 


phil-blain:
I know we do not currently have a strict policy about commit messages, but in my opinion I think we should. Here are some references on the subject:

- https://chris.beams.io/posts/git-commit/
- https://github.com/git-for-windows/git/wiki/Good-commits and links at the bottom of that page.

It's nice to limit the commit title to about 60 characters (some references say 50, but I think it's often not enough), this way `git log --oneline` gives a nice, readable output.

For this PR, I think the commit message should have the same title has the PR, and the rest of the message should describe why this change is needed:


```
box2001: compute aiu with aice instead of aice_init

<description of what the code does at the moment, and why it's wrong>

<an order to the code base to improve itself>
```

The idea is that the rationale behind each change to the code base is encoded in the Git history; we do not rely on discussions on GitHub PRs to understand why each change to the code base was made. This, in my opinion, is good because we are always in control of the repository (since Git is distributed), but we don't know what can happen to GitHub: in theory it could go under at any moment, and we would loose crucial information contained in PR discussions.

(just my 2 cents - we could discuss this at the next teleconference.)

eclare108213:
@phil-blain I agree, it's better to include informative content in the git commits themselves, so that we don't have to dig into PRs or elsewhere to remember what it was about.  The messages need to be concise, though, and I think it's fine to have more verbose content in the PR/discussion.  Most PRs will have multiple commits, and it's helpful for the commit messages to be informative, but I don't think it's necessary or even always desirable for the commit message and the title of the PR to be the same.  Let's talk about this at the next meeting -- please remind me to put it on the agenda if I forget!

eclare108213 (approved):

TillRasmussen (approved):

PR #528, opened 2020-11-02
Add DMI Cray XC50 hpc (freya) to machines. Based on Banting

TillRasmussen:

## PR checklist
    This adds environment and Macros for DMI Cray XC50 computer to machines. Based on Banting XC50.
- [ ] Developer(s): 
    tar
@phil-blain @apcraig 
    Do not change the code itself 
    - [x ] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x ] No
    - [ ] Yes
    - [x ] No
    - [ ] Yes
        - [ ] Yes
        - [ ] No 
- [ x] Please provide any additional information or relevant details below:
For DMI runs big endian flag should be removed.
Big endian flag only in Macro to use test examples from CICE consortium.
Only implemented for intel compiler

### DIFF
--- a/configuration/scripts/cice.batch.csh
+++ b/configuration/scripts/cice.batch.csh
@@ -185,7 +185,15 @@ cat >> ${jobfile} << EOFB
 #SBATCH --qos=standby
 EOFB
 
-else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting* || ${ICE_MACHINE} =~ freya* ) then
+else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting*) then
+cat >> ${jobfile} << EOFB
+#PBS -N ${ICE_CASENAME}
+#PBS -j oe
+#PBS -l select=${nnodes}:ncpus=${corespernode}:mpiprocs=${taskpernodelimit}:ompthreads=${nthrds}
+#PBS -l walltime=${batchtime}
+EOFB
+
+else if (${ICE_MACHINE} =~ freya* ) then
 cat >> ${jobfile} << EOFB
 #PBS -N ${ICE_CASENAME}
 #PBS -j oe
--- a/configuration/scripts/cice.batch.csh
+++ b/configuration/scripts/cice.batch.csh
@@ -185,7 +185,7 @@ cat >> ${jobfile} << EOFB
 #SBATCH --qos=standby
 EOFB
 
-else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting*) then
+else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting* || ${ICE_MACHINE} =~ freya* ) then
 cat >> ${jobfile} << EOFB
 #PBS -N ${ICE_CASENAME}
 #PBS -j oe
--- a/configuration/scripts/cice.launch.csh
+++ b/configuration/scripts/cice.launch.csh
@@ -130,7 +130,7 @@ EOFR
 endif
 
 #=======
-else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting*) then
+else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting* || ${ICE_MACHINE} =~ freya*) then
 if (${ICE_COMMDIR} =~ serial*) then
 cat >> ${jobfile} << EOFR
 ./cice >&! \$ICE_RUNLOG_FILE
--- a/configuration/scripts/cice.launch.csh
+++ b/configuration/scripts/cice.launch.csh
@@ -130,10 +130,22 @@ EOFR
 endif
 
 #=======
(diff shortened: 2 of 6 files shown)


eclare108213 (approved):

configuration/scripts/cice.batch.csh
```diff
@@ -185,7 +185,7 @@ cat >> ${jobfile} << EOFB
 #SBATCH --qos=standby
 EOFB
 
-else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting*) then
```
phil-blain, line 188: Hi Till! I don't really mind that you add your machine here to the same `else if` statement as ours (the PBS directives should not change often), but if you (or I) want to change them in the future, maybe it makes more sense for you to have a separate entry for your machine ?

configuration/scripts/cice.launch.csh
```diff
@@ -130,7 +130,7 @@ EOFR
 endif
 
 #=======
-else if (${ICE_MACHINE} =~ daley* || ${ICE_MACHINE} =~ banting*) then
```
phil-blain, line 133: same comment here. in fact I'm realizing that this is wrong for the serial case, as if we don't use `aprun` the model will run on the MOM node instead of the computes...

configuration/scripts/machines/Macros.freya_intel
```diff
@@ -0,0 +1,75 @@
...
+#INCLDIR := -I. -I/usr/include
+#SLIBS   :=
+
+#--- Compiler/preprocessor ---
+FC         := ftn 
+CC         := cc
+CXX        := CC
+CPP        := /usr/bin/cpp
+CPPFLAGS   := -P -traditional  # ALLOW fortran double backslash "\\"
+SCC   := gcc
```
phil-blain, line 16: any reason you use `gcc` instead of `cc` for the serial case ?

configuration/scripts/machines/Macros.freya_intel
```diff
@@ -0,0 +1,75 @@
...
+# FFLAGS     := -O2 -fp-model precise -assume byterecl -ftz -traceback -xHost
+#endif
+# Preprocessor flags
+#CPPDEFS    := -DLINUX $(ICE_CPPDEFS)
+
+# Linker flags
+
+# Additional flags
+
+ifeq ($(DITTO), yes)
```
phil-blain, line 50: the is probably obsolete.. greping either 'DITTO' and 'REPRODUCIBLE' does not return any result for me..

configuration/scripts/machines/Macros.freya_intel
```diff
@@ -0,0 +1,75 @@
...
+#--- NetCDF ---
+#ifeq ($(IO_TYPE), netcdf)
+#   
+#endif
+#
+#ifeq ($(IO_TYPE), netcdf_bin)
+#   CPPDEFS :=  $(CPPDEFS) -Dncdf
+#endif
+
+### if using parallel I/O, load all 3 libraries.  PIO must be first!
```
phil-blain, line 70: this part should have been removed long ago from our Macro files, PIO is not installed on our machines... So I would suggest removing it also here, to not carry uneeded stuff.

TillRasmussen: gcc: The short answer makdep fails if I use cc. It works when I use gcc. I think, but I am not sure, that the reason is that the XC50 computer at DMI consist of login nodes and compute nodes, which have different architectures. The cc command is used when compiling for compute nodes and gcc for login nodes.
I will separate the computers within the if statement and update.
I agree about the DITTO and REPRODUCIBLE but I left them in there to be sure.
There is an installation of parallel netcdf on the DMI computer but I am not sure if it works. I have never tried CICE or other model systems with pio.

phil-blain: ah yes, makes sense to use `gcc` if the login and computes have different architectures. I think you also could use the `CFLAGS_HOST` macro for that : https://cice-consortium-cice.readthedocs.io/en/latest/user_guide/ug_running.html#cross-compiling

This would be a little "cleaner" I think, since this way the C code (ice_reprosum..) would still be compiled by `cc` (I don't know if it would really make a difference though.)

TillRasmussen:
I have now separated the two machines (Bainting and Freya) in cice.batch.csh and cice.launch.csh

apcraig (approved):
Some effort was made to clean up the Macros a few months ago.  They are now documented, https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_case_settings.html#table-of-c-preprocessor-cpp-macros.  I agree that DITTO and REPRODUCIBLE should be removed at this point, although it's not a big deal if they're not.  I think they were removed from several other machine files at that time (although maybe not all).

In general, I think it's fine to share machine settings in the "if" tests for multiple machines.  It's pretty easy to split them up later if needed.  I agree with @phil-blain's comment that if multiple machine are sharing an "if" section, care must be taken if something needs to change for one machine.

apcraig:
@TillRasmussen, I'll give you a chance to make any additional revisions.  Once you feel you're ready to merge (and maybe that's now), just add a comment to the PR or email me if you prefer.  thanks!

TillRasmussen:
I added option for gnu as well for DMI Freya. I am done with my changes. These can be merged

PR #535, opened 2020-12-01
Revert "deprecate upwind advection (#508)"

apcraig:
This reverts commit 164fcfcfa4c126c9779c087dd1267b08449ddb24.

## PR checklist
    revert removal of upwind advection in #508 
- [X] Developer(s): 
    apcraig
    full test suites on cheyenne with 3 compilers, all bit-for-bit except alt04 which turns on the upwind scheme again.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#c5c1a837084a3492e545df6af9e0c18a917f4f15
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No, but changes alt04
    - [ ] Yes
        - [ ] Yes
        - [X] No 

This is a simple reversion of #508 with subsequent testing.  There are still some outstanding issues that need to be addressed in the upwind advection (see #508, #267), but these will be addressed separately.  Done by

git revert 164fcfcfa4c

Also update ktransport/advection interaction.
  - advection has valid values of remap and upwind, can take value of none only internally.
  - ktransport namelist turns on/off advection, sets advection to none if off.

Separately, add tmpstr2 values in ice_init for cases where it's underfined.


doc/source/developer_guide/dg_dynamics.rst
```diff
@@ -50,9 +50,8 @@ abort if set.  To override the abort, use value 102 for testing.
 Transport
 -----------------
 
-The transport (advection) methods are found in **cicecore/cicedynB/dynamics/**.  Only the incremental
-remapping method is supported at this time, and is set in namelist via the ``advection`` variable.
-Transport can be turned off by setting ``advection = none`` or ``ktransport = -1``.
```
eclare108213, line 55: It would be good to keep this last sentence in the docs, about how to turn transport off.

eclare108213: and maybe reinstate the double quotes around advection

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -356,7 +356,7 @@ dynamics_nml
 
    "", "", "", ""
    "``advection``", "``remap``", "linear remapping advection scheme", "``remap``"
-   "", "``none``", "advection off", ""
```
eclare108213, line 359: need to check whether advection=none works correctly in the reverted code

apcraig: I tried "none" and it worked, but then noticed that the advection timer was still pretty large.  The issue is that in step_dyn_horiz, we have

```
      if (advection == 'upwind') then
         call transport_upwind (dt)    ! upwind                                                                                                                     
      else
         call transport_remap (dt)     ! incremental remapping                                                                                                      
      endif
```

which is a bit of a problem.  Unless you set 'upwind', the advection option will be remap.  In ice_init, remap, upwind, and none are the only accepted options.  So, either we need to remove none as an option or we need to update the code.  Is it as simple as

```
      if (advection == 'upwind') then
         call transport_upwind (dt)    ! upwind                                                                                                                     
      elseif (advection == 'remap') then
         call transport_remap (dt)     ! incremental remapping                                                                                                      
      endif
```

We might as well fix this now while we're thinking about it.  How do you want me to proceed?

eclare108213: I like this:

> elseif (advection == 'remap') then

eclare108213: We also have ktransport, which is supposed to turn it off when it is -1.  is it not wrapped around that 'if advection' code block?

apcraig: Happy to have either/both.  In any case, we either need to remove "none" as a supported advection option or switch to the elseif assuming there are no other issues.  I will test the elseif and you can think about it.

apcraig: Just to clarify, the entire if block is

```
      if (ktransport > 0) then
      if (advection == 'upwind') then
         call transport_upwind (dt)    ! upwind                                                 
      elseif (advection == 'remap') then
         call transport_remap (dt)     ! incremental remapping                                  
      endif
      endif

The only other place ktransport seems to show up is in ice_init.  It seems like we can get rid of ktransport completely and rely only on advection?  We can leave it in the namelist for backwards compatibility.  Maybe ktransport<0 should set advection to 'none' in ice_init (overriding which we try to avoid in general, but it seem required here) and then we get rid of ktransport if check above?
```
so maybe we don't need to support both?

apcraig: Adding the else if seems to have turned advection off with advection=none.  And it completed with Advection timer = 0.00.   So no advection does seem to work for an out of the box case.

The only test we have right now with advection off is "box2001" which sets ktransport=-1.  Do we need a test with advection='none' with the gx3 grid?

eclare108213: I don't think we need a test with advection = 'none'.  We didn't used to have ktransport, but if I remember correctly, @TillRasmussen added it to make the on/off flags more consistent, and we kept 'advection' for backwards compatibility.  Maybe we don't need 'none', but it helps clarify the namelist when ktransport=-1 or 0.  I still like the elseif change, so that remap isn't the default when ktransport > 0.  Although I'm not imagining a situation where we'd have ktransport>0 and advection='none'.  I agree with you, we either need to get rid of the 'none' option for advection or else have this longer elseif statement.  I'm fine either way.

apcraig: If you leave it up to me, I'll deprecate ktransport except as a namelist and have ktransport=-1 override the advection setting, but otherwise have 'upwind', 'remap', and 'none' be working advection options.  I will implement this and run the test suite. @TillRasmussen or anyone, is there a reason not to do that?

eclare108213: Sounds fine to me.  ktransport was added in #151 and I was wrong about who did it -- @mattdturner not @TillRasmussen (sorry Till)

doc/source/developer_guide/dg_dynamics.rst
apcraig: This has been updated.

doc/source/user_guide/ug_case_settings.rst
apcraig: The lastest commit, 6aeb366, implements the new proposed handling of advection='none' and deprecation of ktransport.

apcraig:
The latest commit, 6aeb366 was tested on cheyenne using the full test suite with one compiler and all the tests still pass except the alt04 cases as before.  https://github.com/CICE-Consortium/Test-Results/wiki/c5c1a83708.cheyenne.intel.20-12-02.020322.0.

JFLemieux73:
Thanks @eclare108213 and @apcraig for taking care of this. I think it is better if @eclare108213 does the review.

eclare108213 (commented):
I think this (getting rid of ktransport completely) is simpler, but @mattdturner had a good reason for implementing ktransport. In general, CICE's namelist is not consistent in how parameterization options are implemented, which makes it confusing for new users.  Opinions, anyone, on whether we should keep ktransport as just an on/off switch with 'advection' determining the scheme used (or not)?

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -355,8 +355,9 @@ dynamics_nml
    :widths: 15, 15, 30, 15 
 
    "", "", "", ""
-   "``advection``", "``remap``", "linear remapping advection scheme", "``remap``"
-   "", "``none``", "advection off", ""
+   "``advection``", "``none``", "advection disabled", "``remap``"
```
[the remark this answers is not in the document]
apcraig: It's there.  The next two lines define remap and upwind.  remap in the first line defines the default in the table.  We are listing the options alphabetically, each on a separate line and the default in on the first line.

JFLemieux73:
I think we should make things simpler and have only one. Do we prefer to have options as strings (like advection) or as integer (like kstrength)?

doc/source/science_guide/sg_horiztrans.rst
```diff
@@ -33,9 +33,10 @@ introductory comments in **ice\_transport\_remap.F90**. Prognostic
 equations for ice and/or snow density may be included in future model
 versions but have not yet been implemented.
 
-One transport scheme is available, the incremental
+Two transport schemes are available: upwind and the incremental
 remapping scheme of :cite:`Dukowicz00` as modified for sea ice by
-:cite:`Lipscomb04`. The remapping scheme has several desirable features:
+:cite:`Lipscomb04`. The CICE model also allow the transport scheme
```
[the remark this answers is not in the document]
apcraig: Sorry about that, I just fixed it.

doc/source/user_guide/ug_case_settings.rst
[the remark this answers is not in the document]
eclare108213: okay, thanks.  I should have expanded the code listing.

mattdturner:
> I think this (getting rid of ktransport completely) is simpler, but @mattdturner had a good reason for implementing ktransport.

I created the PR that added `ktransport`, but @rallard77 did the development.  It might be good to bring him into the discussion.

eclare108213:
> I think we should make things simpler and have only one. Do we prefer to have options as strings (like advection) or as integer (like kstrength)?

We have both, for historical reasons (it's an old code!  originally Fortran77...).  My personal preference is for strings, since they carry more information content.  I think ktransport was added so we could turn it all off using -1, like we do for other things like kdyn, kridge, ktherm, etc.  The benefit of having a separate on/off flag is being able to turn off several related parameterizations at once regardless of how the individual flags are set (e.g. turning off dynamics could turn off momentum, stress, transport and ridging -- NB the code might not actually do this now).  

I'm hesitant to let this become a big issue for this PR (and I know I brought it up!), but it would be good to move our namelist values gradually toward the format(s) that we prefer, cleaning up as we go.

apcraig:
@eclare108213, I agree that a case can be made for having both integers and strings.  If we were starting from scratch, I'd probably advocate strings.  I'd also probably advocate for logicals instead of integers for turning things on and off.  

I think this issue came up here because we were supporting "none" for advection but it was not working as expected, so we had to do something.  I'm certainly willing to deprecate "none" and put ktransport back into the code if folks prefer that.

rallard77:
I have no preference. I can support either proposed options...

Rick

eclare108213:
Thanks @rallard77.  I'm open to other ways of doing this, but here's my preference:
- don't deprecate ktransport completely, but only use it in ice_in and ice_init to set advection='none' when needed (so it overrides the advection setting).  Then only the advection string is used during the timestepping.

apcraig:
@eclare108213, should 'none' also be a valid option for advection?  

I will un-deprecate ktransport in the documentation and otherwise but it will only be used to set advection=none in ice_init.

eclare108213:
If ktransport isn't used during the timestepping, then I think you'll need the advection = 'none' setting.

apcraig:
Right.  Sorry I wasn't clear.  Do we want advection='none' to be a valid and documented option for the advection namelist.  It sounds like we'll continue to support the ktransport "flag" in the namelist.  Do we want the advection options to be just "upwind" and "remap"?

eclare108213:
I see.  I think it's fine to have 'none' as an option in the namelist, but the argument against it is to not have 2 different ways to turn off the advection in there, which could be confusing.  So (it feels like we've come full circle), let's just let ktransport control off/on, allow advection to take the value 'none' in the code but not in the namelist, and during the initialization check whether the namelist values of advection are either remap or upwind.  If not, print a warning, and also print a warning when resetting to 'none'.  Does that make sense?

apcraig:
that all makes sense.  will do.

TillRasmussen:
I think that arguments can be found for both. Strings explain better.
Integers allows for combinations. For instance the basal stress from icebergs is currently being implemented.
One could define an integer for this and the old basal  stress
0 = none included
1 = basal stress
2 = iceberg grounding
3 = both basal stress and icebergs

There are workarounds for both choices.

Till

JFLemieux73:
I am not sure I follow all these comments but let's make sure (whatever if it is ktransport or advection) that we have an option so that we do NOT advect (or transport) the ITD...good for idealized tests.

eclare108213:
@JFLemieux73 please clarify:  are you talking about horizontal transport of the ice, or transport in thickness space (i.e. the ITD)
?  The ktransport and advection flags only control horizontal transport.

JFLemieux73:
I mean horizontal transport.

apcraig:
I just pushed an update that I think meets the latest requirements, 

> So (it feels like we've come full circle), let's just let ktransport control off/on, allow advection to take the value 'none' in the code but not in the namelist, and during the initialization check whether the namelist values of advection are either remap or upwind. If not, print a warning, and also print a warning when resetting to 'none'.

The test results look fine for a full suite on cheyenne with 1 compiler, everything passes bit-for-bit except the alt04 case which invokes upwind again.

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#610f440ba3455d3e57e8cfdd77fdf2082c7eb41e

apcraig:
I did one other thing while I noticed it.  In the ice_init namelist diagnostics, there are cases where tmpstr2 could be undefined (or stale) and then written.  In cases where tmpstr2 is set by an if test without a final "else", I think we should define that final "else" case, even if we don't expect it to trigger.  It's easy to end up with a mismatch in the namelist checks and the namelist diagnostics and then for tmpstr2 to write something that makes no sense.

eclare108213 (approved):
Looks good.  Thanks for fixing the tmpstr2 issues, straightening out the namelist options and for reverting the upwind code, @apcraig.

Issue #521, opened 2020-10-13
Unelegant crash writing initial condition when 'histfreq' is more frequent then any 'f_*' output variable


phil-blain:
I noticed that if I change `histfreq` to `d` or `1` in the namelist, and I don't change nothing else from the defaults (i.e. all `f_*` output variables are either set to `'m'` or `'x'`), then  the model crashes very inelegantly in `accum_hist` when called in `cice_init` to write the initial condition.

Here is the backtrace from the Intel compiler runtime (code compiled with `-traceback`):

```bash
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable A2D when it is not allocated
Image              PC                Routine            Line        Source
cice               0000000001F6AE76  Unknown               Unknown  Unknown
cice               0000000000E0DFDA  ice_history_mp_ac        1893  ice_history.F90
cice               000000000040383B  cice_initmod_mp_c         224  CICE_InitMod.F90
cice               0000000000401E83  cice_initmod_mp_c          52  CICE_InitMod.F90
cice               0000000000401671  MAIN__                     43  CICE.F90
cice               00000000004015F2  Unknown               Unknown  Unknown
cice               000000000203A08F  Unknown               Unknown  Unknown
cice               00000000004014DA  Unknown               Unknown  Unknown
```

The call to `accum_hist` is here:
https://github.com/CICE-Consortium/CICE/blob/e662c7908931c5f812ac2feb66b5415fbf1fc36c/cicecore/drivers/standalone/cice/CICE_InitMod.F90#L224

And this is the line where it crashes:

https://github.com/CICE-Consortium/CICE/blob/e662c7908931c5f812ac2feb66b5415fbf1fc36c/cicecore/cicedynB/analysis/ice_history.F90#L1893

It's not super important, but it's a little bit disconcerting. I think we could improve that and abort early, with a message like:
```
(WARNING): 'histfreq' set to 'd', but no outputs are set for this frequency
```

or something like that... at least we should not cause a runtime crash like this.

dabail10:
I'll take responsibility for this one.

apcraig:
#548 may address this in part.

phil-blain:
I just had this issue again, relaunching the exact same case as two weeks ago... I guess it depends on the state of the memory on the system... 

I could not find a core dump though....

~~~
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable A2D when it is not allocated

Image              PC                Routine            Line        Source
[...]

PR #548, opened 2020-12-18
zsal option for testing zsalinity code

eclare108213:

## PR checklist
Provides -s zsal option for testing zsalinity code and improving code coverage
- [x] Developer(s): 
@eclare108213 

https://github.com/CICE-Consortium/Test-Results/wiki/9d9b65ce8f.badger.intel.20-12-18.233253.0

193 measured results of 193 total results
189 of 193 tests PASSED
0 of 193 tests PENDING
0 of 193 tests MISSING data
4 of 193 tests FAILED

The 4 failures are the usual badger 40-proc fails.  The smoke test in this batch of test results is for a 1-year run, which took so long that I changed it to a 5-day test with debugging turned on in the base_suite modification.

    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [x] Yes
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [ ] No 

I'm  not sure whether documentation must be manually updated when new tests are added.

See https://github.com/CICE-Consortium/Icepack/pull/344 includes roughly equivalent tests for Icepack.
Note that z_tracers=T is not needed for Icepack, but is for CICE.


eclare108213:
[plots.tar.gz](https://github.com/CICE-Consortium/CICE/files/5718484/plots.tar.gz)

Here are annual timeseries plots and cice.runlog files for the zsal run with the new option and for the default (mushy).   The initial conditions are different but overall the behavior is similar and more or less as expected.  I'm not sure why the zsal salinity starts so high; it decreases through the year and does not reach the low values seen in mushy.

@njeffery if you would please take a look at cice.runlog.zsal in the attached file, where the salinity profile is printed out through the run.  It looks okay to me but you're more of an expert on this.  If you think it's okay, I'll add this as part of our regular test suite.

njeffery:
It looks like it's working this time.  That's great. It's not a confirmation that everything's working precisely as it should, but I'm comfortable with including as a testsuite.


njeffery (approved):
The initial run looks reasonable.  There still may be issues with the code, but at least the zsalinity portion is running.

apcraig (commented):
I think the restart test should be done with threading (say 8x2).  Does the smoke test with debug for 5 days run so slowly that it needs medium?

eclare108213:
I'll try these tests on badger and report back.

eclare108213:
My testing ran into problems where, for history fields defined outside of ice_history.F90 (mechred, ponds, bgc), the code was trying to accumulate arrays that had not been allocated because there were no similarly shaped fields turned on in the main history namelist.  It shouldn't have been trying to accumulate them in the first place -- I think I fixed this.  Since the changes also affect mechanical redistribution and melt pond output, I'm running a full base_suite again. Note: base_suite might not fully test the history functionality, something we should probably look into as part of our code coverage work.  A particular case:

histfreq = 'd'
f_aice = 'd' and no other history variables = 'd'
f_fbri = 'm' caused the brine height tracer to be accumulated but aD3C was not allocated since aice is 2D.  

[Test results](https://github.com/CICE-Consortium/Test-Results/wiki/a98f3adf92.badger.intel.20-12-23.190632.0). The fails are the usual 40-proc tests.  

I'm comfortable with this PR and the related [Icepack PR](https://github.com/CICE-Consortium/Icepack/pull/344) being merged now.

apcraig:
Do the history fixes relate at all to #521?

apcraig (approved):

eclare108213:
Yes, #521 is definitely related.  That is the same error I was getting, but for the a3Dc array instead of a2D.  My changes might have fixed that too, but additional conditionals might still be needed in ice_history.F90.  I only fixed the ones in the other ice_history_*.F90 modules.

apcraig:
@eclare108213, Testing on other compilers picked up a problem with the debug zsal test, https://github.com/CICE-Consortium/Test-Results/wiki/f5f487f973.cheyenne.gnu.20-12-23.235355.0.  The test fails with the gnu compiler at line 1013 in icepack/columnphysics/icepack_algae.F90.  Some quick debugging indicates bphin_N is not initialized because nbtrcr is zero.  This only fails with the gnu compiler, I assume the gnu compiler with debug flags trap uninitialized values while the other compilers might only do so if they are initialized to invalid values.

I don't have any further information, but could debug further if you like.  I guess this is an interaction of the bio and algae settings?  Maybe algae shouldn't be called at all unless bgc is on or maybe there is some other logic that belongs here?  Let me know if you want me to create a separate issue.

eclare108213:
Algae shouldn't be called unless the bgc is turned on, either skeletal or zbgc.   Looks like it's getting called because z_tracers=T, which @njeffery suggested to turn on in order to initialize zsalinity.  So we either need to increase nbtrcr or else debug the initialization of zsal...

I noticed that 'zbio' is used in two different ways in the code, as a keyword and as a subroutine name.  That's probably not the problem in this case but still ought to be fixed.

Issue #61, opened 2017-12-26
uvm initialization


eclare108213:
the NAG compiler reports a missing initialization in the uvm halo, line 1399 of ice_grid.f90
(issue noted by DMI collaborators)

dabail10:
I'm confused by this one. We do a halo update of uvm right after we compute it. Also, the NAG compiler with -nan on was not catching this.

eclare108213:
Okay, let's not worry about it for now.  They've been working on their version of the code, which we will try to merge into the main CICE sooner or later.  I'll ask them to run the new code through their tests, and we'll see if it comes up again.  I'll take this off of the project board but keep the issue open.

eclare108213:
@mhrib @TillRasmussen 
This is one of the issues that I entered based on your original text describing the various tests that you'd run, which turned up some problems.  If you can, please run those tests again with the current code and see which problems persist.   See also #59.

eclare108213:
@dabail10 please test with your Nag compiler - if it comes back clean then we can close this issue.

TillRasmussen:
I have looked into the compile flags that was used. For IEEE, which were the one that displayed the issue, the flags were "-g -00 -ieee=stop -nan". Unfortunatly we dont have the compiler inhouse at DMI anymore.

apcraig:
I have been playing with the nag compiler this morning and have not been able to duplicate this error.  There is an interesting line at 1584,
`!     uvm = c0                                                                                   
`which seems to have been modified (commented out) in 2018 by Dave.  Not sure why, but given we are not able to duplicate the error from DMI, I think we can close this.

There is one other thing though.  When I add -nan to the compile, I get a separate error,

> /home/tcraig/cice-consortium/cice.nagclean/cicecore/cicedynB/dynamics/ice_dyn_evp_1d.F90: In function ‘dmi_omp_MP_domp_get_domain_rlu’:
> /home/tcraig/cice-consortium/cice.nagclean/cicecore/cicedynB/dynamics/ice_dyn_evp_1d.F90:110:58: error: ‘dmi_omp_MP_rdomp_iamTPI’ undeclared (first use in this function)
>    subroutine domp_get_domain_rlu(lower,upper,d_lower,d_upper)
>                                                           ^
> /home/tcraig/cice-consortium/cice.nagclean/cicecore/cicedynB/dynamics/ice_dyn_evp_1d.F90:110:58: note: each undeclared identifier is reported only once for each function it appears in
[...]

PR #560, opened 2021-01-28
cicecore: fix some use of uninitialized variables

phil-blain:
## PR checklist
   cicecore: fix some use of uninitialized variables
- [x] Developer(s): 
    P.Blain
   **in progress**
    - [ ] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

This is the first one of a series of small PRs I'm preparing to upstream some changes I've had to make to CICE6 during my work on coupling with NEMO3.6.  This one deals with use of uninitialized values, which I've detected by compiling with `-init=snan,arrays` (Intel). I will start the base_suite right away and update the description with the test results once they are finished running.

See the [commit messages](https://github.com/CICE-Consortium/CICE/pull/560/commits) for complete details.

### DIFF
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -384,7 +384,7 @@
       ! T-grid cell and U-grid cell quantities
       !-----------------------------------------------------------------
 
-!     tarea(:,:,:) = c0
+      tarea(:,:,:) = c0
 
       !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block)
       do iblk = 1, nblocks
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -1636,7 +1636,7 @@
       !-----------------------------------------------------------------
 
       bm = c0
-!     uvm = c0
+      uvm = c0
 
       !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block)
       do iblk = 1, nblocks
--- a/cicecore/drivers/standalone/cice/CICE_RunMod.F90
+++ b/cicecore/drivers/standalone/cice/CICE_RunMod.F90
@@ -627,11 +627,12 @@
       
       real (kind=dbl_kind)    :: &
           puny, &          !
+          Lsub, &          !
           rLsub            ! 1/Lsub
 
       character(len=*), parameter :: subname = '(sfcflux_to_ocn)'
 
-      call icepack_query_parameters(puny_out=puny)
+      call icepack_query_parameters(puny_out=puny, Lsub_out=Lsub)
       call icepack_warnings_flush(nu_diag)
       if (icepack_warnings_aborted()) call abort_ice(error_message=subname, &
          file=__FILE__, line=__LINE__)
--- a/cicecore/cicedynB/general/ice_flux.F90
+++ b/cicecore/cicedynB/general/ice_flux.F90
@@ -654,9 +654,7 @@
       enddo
       enddo
 
-#ifndef CICE_IN_NEMO
       sst   (:,:,:) = Tf(:,:,:)       ! sea surface temp (C)
-#endif
       qdp   (:,:,:) = c0              ! deep ocean heat flux (W/m^2)
       hmix  (:,:,:) = c20             ! ocean mixed layer depth (m)
       hwater(:,:,:) = bathymetry(:,:,:) ! ocean water depth (m)


eclare108213 (commented):
These changes might affect other coupled models, e.g. for restarts, so I'll ask @apcraig and @dabail10 to take a look.

TillRasmussen:
Generally fine. A few comments.
Tarea: loops from 1 to nx_block when calculating northern or southern hemisphere average. This is used for diagnosticks and global sums. My question is more why it is necessary to include halo zones in a global sum of e.g ice concentration. Hence should the loop indices be changed instead? 
uvm: There is a call to ice_halo_update with uvm just after its initialization. I would assume that the halos are updated there.
An offline topic: 'we should compare sbcice_cice between the nemo3.6 and the nemo 4

phil-blain:
Thanks for taking  a look Till!
> Tarea: loops from 1 to nx_block when calculating northern or southern hemisphere average. This is used for diagnosticks and global sums. My question is more why it is necessary to include halo zones in a global sum of e.g ice concentration. Hence should the loop indices be changed instead?

I think I agree. I did not look into it further when I saw that the initialization was commented, I just uncommented it. What you suggest seems more sane. 

> uvm: There is a call to ice_halo_update with uvm just after its initialization. I would assume that the halos are updated there.

You mean that the initialization should not be necessary ? Well at least on our "CREG025-Pacific extension" grid, some values were uninitialized in the condition [at line 1673](https://github.com/CICE-Consortium/CICE/pull/560/commits/c136267896615793c42ec959d03698d5483a1ac8#diff-5c8e815b29d7bb8283f9ed89427d8271562e374797b1a57fe8174825ff736b2fR1673) if I did not initialize `uvm` to zero... maybe something else is hiding here with the halos on that grid... 

> An offline topic: 'we should compare sbcice_cice between the nemo3.6 and the nemo 4

Yes, I looked at the changes to sbc_ice_cice in your nemo4cice project and they were mostly the same as mine, although our version has a lot of in-house modifications, that I don't yet fully understand. Sometimes stuff was implemented in a not so clean way because of lack of time, and it's one of the mid-term goals of my project to clean that up. We definitely will let you know.

cicecore/cicedynB/infrastructure/ice_grid.F90
```diff
@@ -384,7 +384,7 @@ subroutine init_grid2
       ! T-grid cell and U-grid cell quantities
       !-----------------------------------------------------------------
 
-!     tarea(:,:,:) = c0
+      tarea(:,:,:) = c0
```
dabail10, line 387: I think this is unnecessary as tarea is in a HaloUpdate shortly after it is computed. I don't think you want any zero values in tarea at all.

cicecore/cicedynB/infrastructure/ice_grid.F90
```diff
@@ -1636,7 +1636,7 @@ subroutine makemask
       !-----------------------------------------------------------------
 
       bm = c0
-!     uvm = c0
+      uvm = c0
```
dabail10, line 1639: Similarly here. The HaloUpdate should take care of uvm values in the ghost cells.

cicecore/drivers/standalone/cice/CICE_RunMod.F90
```diff
@@ -627,11 +627,12 @@ subroutine sfcflux_to_ocn(nx_block,   ny_block,     &
       
       real (kind=dbl_kind)    :: &
           puny, &          !
+          Lsub, &          !
           rLsub            ! 1/Lsub
 
       character(len=*), parameter :: subname = '(sfcflux_to_ocn)'
 
-      call icepack_query_parameters(puny_out=puny)
+      call icepack_query_parameters(puny_out=puny, Lsub_out=Lsub)
```
dabail10, line 635: This is done in ice_forcing.F90. Why does it need to be here?

cicecore/cicedynB/infrastructure/ice_grid.F90
phil-blain: Thanks for the review. I agree that it should not be necessary, but as I wrote, if I do not initialize here, an uninitialized value gets used... So that would mean that not all halo cells are initialized by the halo update. I'll have to investigate that further on the specific grid we are using.

cicecore/drivers/standalone/cice/CICE_RunMod.F90
phil-blain: If you mean what is the point of the whole subroutine, I'm not yet familiar with all of the intricacies of how coupling works, so I don't know the specifics. The comments at the top of the subroutine state the following:

~~~
! If surface heat fluxes are provided to CICE instead of CICE calculating
! them internally (i.e. .not. calc_Tsfc), then these heat fluxes can 
! be provided at points which do not have ice.  (This is could be due to
! the heat fluxes being calculated on a lower resolution grid or the
! heat fluxes not recalculated at every CICE timestep.)  At ice free points, 
! conserve energy and water by passing these fluxes to the ocean.
~~~


Here I'm just making sure the code compiles fine when `CICE_IN_NEM` is defined

cicecore/cicedynB/infrastructure/ice_grid.F90
dabail10: It could be that your compiler is being very picky. What line does it abort on exactly with the uninitialized variable? I've done the check with NAG and Intel and it runs fine. As you say, it might be a particular feature of your grid. There was an issue awhile ago from one of the European groups that trapped this same error. I am fine with initializing uvm to zero, but tarea should never be zero.

dabail10 (commented):
No, I meant the Lsub initialization. This is done in ice_forcing.F90.

cicecore/drivers/standalone/cice/CICE_RunMod.F90
phil-blain: > No, I meant the Lsub initialization. This is done in ice_forcing.F90.

Ok. Here I'm ont initializing Lsub, just using it. The variable was not declared in that subroutine.It had not been catched because none of our tests compile with CICE_IN_NEMO.

dabail10: Got it. I'm not familiar with the CICE_IN_NEMO code pieces.

TillRasmussen: @dabail10 This entire subsection only applies when calc_Tsfc is false and CICE_IN_NEMO is on

TillRasmussen:
The European group may have been DMI. I should look at the report Jacob wrote some time ago

dabail10:
It was Issue #61 . I have reopened it and linked it here.

cicecore/cicedynB/infrastructure/ice_grid.F90
phil-blain: It aborts at line [1673 in `makemask`](https://github.com/CICE-Consortium/CICE/blob/a76893a0d47c3c2d0c8fbe83238aa3b000dabd1b/cicecore/cicedynB/infrastructure/ice_grid.F90#L1694)

cicecore/cicedynB/infrastructure/ice_grid.F90
phil-blain: It aborts at line [1673 in `makemask`]( https://github.com/CICE-Consortium/CICE/blob/a76893a0d47c3c2d0c8fbe83238aa3b000dabd1b/cicecore/cicedynB/infrastructure/ice_grid.F90#L1673)

cicecore/cicedynB/infrastructure/ice_grid.F90
dabail10: Ok. A cleaner initialization would be as follows:

do j = 1, ny_block
   do i = 1, nx_block

      im = max(i,ilo)
      jm = max(j,jlo)
      ip = min(i,ihi)
      jp = min(j,jhi)

Then replace all of the i,j indexes on the right with ip/jp. This will extrapolate the tarea values to the ghost cells including the four corners. Thoughts?

dabail10: Sorry, these should all be im/jm or ip/jp.

apcraig: If the maskhalo is turned on, it's possible that is also masking some part of the initialization.  Just speculation on my part, we'd have to look into it further.

apcraig: Actually, the maskhalo is not being used for the halo update of tarea, which is correct, so forget I said that.   I assume tarea is not initialized on the entire halo because of the land mask?  We could also fix the loop in makemask, but that's likely to affect other loops in other parts of the code in the same way.  This is almost certainly a case of variables not being initialized in places they are not needed, but where they are referenced at some point later.   The proper fix, of consistently implementing all loops, is probably not worth the effort, but if we want to update a loop here or there, that's probably OK too.  I'm not opposed to this type of fix in this case.

apcraig (approved):

apcraig:
@phil-blain, is testing complete, is this ready to merge?

phil-blain:
not ready to merge at all. I've yet to go back into the code and come up with a better solution for `tarea`, and understanding why this issue comes up only on the grid we're using. I'm working on that.

TillRasmussen:
This problem with uvm was also seen at DMI a year or two ago with the NAG compiler.

phil-blain:
Ok I'm beginning to understand a little bit more. It seems it is due to the fact that the domain decomposition we are using has some padded blocks. 

I was able to plot `tarea(:,:,iblk)` from a core dump for some values of iblk and indeed some values are missing (here iblk = 33 and 34):

tarea_33
tarea_34

So my guess at this point is that the halo updates are not correctly updating padded blocks ? The darker line around the edge of the plots have value 1.0, I'm don't know yet where this value is coming from...

TillRasmussen:
I think that 1. originates from the icehalo update
     call ice_HaloUpdate (tarea,              halo_info, &
                           field_loc_center,   field_type_scalar, &
                           fillValue=c1)

dabail10:
Good catch. Cell padding has always been a headache for me. I try to avoid it. I think that the cells in the padded area are not even treated as ghost cells. So, that means halo updates do not touch them. However, the model should never reach in here. In my opinion, the compiler is just being pedantic about uninitialized variables here.

TillRasmussen:
If I understand the code correct then block size (nx_block and ny_block) are constant wheras jhi and ihi are adjusted to the global domain size. Then in the last block towards east and north it may run out of the the global domain when it loops from 1, ny_block or 1, nx_block. (padding from e.g ihi ot nx_block).

apcraig:
I believe that is all correct.  Each block has it's own ilo, ihi, jlo, jhi defined.  That's what this is all about,

```
      do iblk = 1, nblocks
         this_block = get_block(blocks_ice(iblk),iblk)         
         ilo = this_block%ilo
         ihi = this_block%ihi
         jlo = this_block%jlo
         jhi = this_block%jhi

```
The padded areas should not be used in general.  If they are, we might want to try to fix that.  I assume the halos do work correctly on the padded blocks.  We do have some bit-for-bit checks for different decomps, some padded.  Is the padded cell halo index ihi+1 or is it still nx+1?  I actually don't know.

eclare108213:
Not looking at the code, so someone needs to confirm this, but what I recall is that the padded cells start outside of the halo cells, and they should be treated as 'missing'.  Entire missing (land) blocks are never accessed, but it's possible that 'missing' padded cells still get noticed by compilers, even if they are never accessed.  If they are being accessed, then there's definitely a bug.

dabail10:
Very good points. Don't we use nghost as we might have more than one ghost cell in each direction? The ilo, ihi, jlo, jhi and nx_block, ny_block are unique for each block including the padded areas. There could be loops that only go to +/- nghost relative to the lo/hi indices. The padded areas should still be "interior" cells, but just masked out right? This has always confused me.

eclare108213:
The padded cells are assigned a global index of zero, which can never occur for physical cells.  This is how they are identified and avoided for boundary conditions, parallelization, etc.   ilo:ihi, jlo:jhi index only the physical domain, and nx_block, ny_block are those physical cells plus ONE row or column of ghost cells.  The code was written in a general way to allow more ghost cells, but this has never been tried and there are comments in the code saying that some routines will need to be updated to use more than one.  Padded cells lie beyond nx_block, ny_block for the eastern- and northern-most blocks, and they should never be accessed in the main code, once the decomposition has been done.  It's possible that the haloes along those edges aren't right (especially the two tripole forms), but the code has been exercised enough that I trust those pretty well.

TillRasmussen:
I do think that nx_block and ny_block includes the padded celles. The function create_blocks in cicecore/cicedynB/infrastructure/ice_blocks.F90 reduce jhi and ihi in the northernmost and westernmost blocks. It appears to be the case in the plots @phil-blain attached.  I have not been able to find where (if) nx_block and ny_block are altered in the case where padding is required.

phil-blain:
@TillRasmussen, coming back to your first commment up top : 
> Tarea: loops from 1 to nx_block when calculating northern or southern hemisphere average. This is used for diagnosticks and global sums. My question is more why it is necessary to include halo zones in a global sum of e.g ice concentration. Hence should the loop indices be changed instead?

I'm not exacty sure what you mean in fact.

The loop in makemask that uses `tarea` initializes `tarean` and `tareas`:
https://github.com/cice-consortium/CICE/blob/c136267896615793c42ec959d03698d5483a1ac8/cicecore%2FcicedynB%2Finfrastructure%2Fice_grid.F90#L1693-L1699

These arrays are only used in module **`ice_diagnostics`**.

They are used as the multiplicative mask for computing several quantities, like first the total ice area:
https://github.com/cice-consortium/CICE/blob/d0084ddf5697fdae6e22bb798c7ece19088dd72d/cicecore%2FcicedynB%2Fanalysis%2Fice_diagnostics.F90#L221-L223

  This is OK as the loop in `global_sum` uses ilo,ihi,jlo,jhi (same thing for `global_sum_prod`):
https://github.com/cice-consortium/CICE/blob/c136267896615793c42ec959d03698d5483a1ac8/cicecore%2FcicedynB%2Finfrastructure%2Fice_grid.F90#L1693-L1699

---

Back to the padding/halo discussion:


@TillRasmussen , you write just above:
> I do think that nx_block and ny_block includes the padded celles. 

I also think that's the case.

@apcraig, you write:
> The padded areas should not be used in general. If they are, we might want to try to fix that. I assume the halos do work correctly on the padded blocks. We do have some bit-for-bit checks for different decomps, some padded. Is the padded cell halo index ihi+1 or is it still nx+1? I actually don't know.

@eclare108213 you write:

> it's possible that 'missing' padded cells still get noticed by compilers, even if they are never accessed. If they are being accessed, then there's definitely a bug.

> Padded cells lie beyond nx_block, ny_block for the eastern- and northern-most blocks, and they should never be accessed in the main code, once the decomposition has been done.

From what I understand (correct me if I'm wrong), how it works is that all blocks have the same `nx_blocks`, `ny_blocks`, so blocks at eastern and northern edges of the grid might get padded depending on the size of the global domain. So for the padded blocks, the padding cells are those with indices [ihi+1, nx_block]  and [jhi+1, ny_block]. This is what I see in the figures: for both blocks `jhi = 70` and `ny_blocks = 76` and we have 6 cells of padding. The right-most colored cells have `j = 70 = jhi`. Note that `jhi + 1` cells are NaN (shows as white), and so even though the halo updates have already taken place, they did not fill `jhi + 1`. They did fill `ny_block`, the right most darker line, with value 1.0, which like Till wrote comes from the `fillValue  = c1`: https://github.com/cice-consortium/CICE/blob/c136267896615793c42ec959d03698d5483a1ac8/cicecore%2FcicedynB%2Finfrastructure%2Fice_grid.F90#L442-L444

I don't know if that is expected behaviour or not for halo updates of padded cells.

Nonetheless, the loop in makemask goes from 1 to `nx_block`, `ny_block`, so it *will* definitely access the padded cells. And since I initialize with NaN, it crashes the model. So I think that maybe initializing ̀`tarea` to some value (be it zero or one or whatever) is actually OK, since apart from that loop the padded cells should not accessed anywhere ...

---

For `uvm`, it's kind of the same situation: in makemask we initialize it on a loop going from `jlo,jhi` and `ilo,ihi`, then call HaloUpdate, then access it in a loop on `1, ny_block` and `1,nx_block`, so again we access the padded cells.

----

I will try to run the decomp suite and initialize with NaNs, see if I can trigger the same behaviour there.

eclare108213:
You're right, the padded cells lie inside nx_block, ny_block.  They are outside of nx_global, ny_global, and in fact they should be outside of nx_global+1, ny_global+1.  If I understand the conversation so far, the only problem is in makemask?  Or is it more general, that the halo updates aren't doing the halos inside of padded regions correctly?  I agree with @TillRasmussen that for diagnostics we should only need to use the physical domain (ilo:ihi,jlo:jhi), but on the other hand, in some cases we might be using the full block to check for bugs (e.g. at one point we weren't getting proper tarea values in the haloes due to land masking, so the diagnostics might be looking for that).  From @phil-blain 's analysis it seems that there is a bug for the halo cells ihi+1 (similarly for j).  When I looked at the code that defines the padded cells, I think i_global=0 for the cells beyond ihi+1 in the eastern-most blocks, which is correct---the halo cells (ihi+1) should not be padded cells.  But maybe some of the logic or loops after that don't match, so there's a bug.  I'll be in a day-long meeting all day today so maybe can come back to this tomorrow.

TillRasmussen:
It is used several places without checking for the global domain
Many places it does not do any harm as it is initializing to zero, however in my opinion we CICE should newer loop through padding cells. I think that the correct loop would be i=1,ihi+nghost and j=1,jhi+nghost.

apcraig:
Strictly speaking, that should probably be ilo-nghost,ihi+nghost, jlo-nghost,jhi+nghost.

I have mixed feelings about how to proceed.  We obviously should fix the problems that have arisen.  There are several ways to do that, like initializing the padded areas if they aren't or avoiding computations in the padded regions.  I don't feel strongly about taking one approach vs another.  I think we could separately focus on fixing loops where computations are done in padded regions to avoid those padded regions, but it seems like that hasn't been a problem in general.  Most critical loops already using the lo/hi indexing.  Finally, I don't think we should worry too much about initializing the full (padded) array to zero and things like that at this point.  I guess my recommendation is to fix loop indexes where they need to be or are obviously wrong.  But I don't think we need to go thru the code with a fine tooth comb to review and potentially rewrite every loop at this point.  But I'm not set one way or another.

eclare108213:
I agree.  There are a lot of problems or inconsistencies in the code that we are gradually fixing, as we come across them.  Let's definitely fix the loops that are causing the crashes in this case, and initialize stuff better if that's what's needed.  Otherwise we can keep our eyes open for similar issues as new code modifications go in.

As an aside, it would be interesting to know how much of the code has 'aged' versus getting recoded or modified somehow.  Probably a lot of it turns over, so speak, every 10 years or so.  Although the 'blame' button gives some info, it doesn't necessarily capture white space changes...  I'm not proposing we figure this out!  But I've also wondered how many lines of code we actually have (as opposed to comments and blank lines).  Just curious.
e

phil-blain:
[I thought I had hit the "Comment" button on Friday, but is seems not!]

Yeah, after thinking about this since I posted  my response above I also think that fixing the loop indices where they need to be fixed is a better approach. This way, we should easily find other places in the code that need to be fixed, whereas initializing the padded cells will "hide" possible loop indices "errors" in those other places.

I will also have a closer look at the halo update code to understand what's going on with the padded cells.

apcraig:
I just ran the following test on cheyenne with 3 compiilers,

smoke_gx3_8x2x8x10x20_debug_diag1_droundrobin_run2day

and all pass.  There is a long standing issue to update the debug flags to check more things and initialize data to Nans.  @phil-blain, can you tell me what compiler flags you're using for your testing.  This is with an intel compiler, correct?  

It might be worthwhile to expand this PR to extend the debug flags, add a debug padded test into the base suite, test with multiple compilers, and fix loops.  I think that requires some comprehensive efforts, and I'm willing to take that on.  Maybe we can include it by the end of the month release?  I'm also happy to just merge this PR as a starting point to fix the issues that @phil-blain has run into.  We can then continue addressing the issue moving forward as I've suggested.

phil-blain:
[@apcraig I was writing this as I received your answer, so I'll read it and answer separately]


OK, some updates now:

I ran the `decomp_suite` with arrays initialized to NaNs (without changing the loop indices and without initializing tarea) and I can reproduce the failure that I get on our grid:
~~~ 
$ cat results.log | \grep FAIL
FAIL daley_intel_restart_gx3_6x2x4x29x18_debug_dspacecurve run
FAIL daley_intel_restart_gx3_6x2x4x29x18_debug_dspacecurve test 
FAIL daley_intel_restart_gx3_8x2x8x10x20_debug_droundrobin run
FAIL daley_intel_restart_gx3_8x2x8x10x20_debug_droundrobin test 
FAIL daley_intel_restart_gx3_4x2x19x19x10_debug_droundrobin run
FAIL daley_intel_restart_gx3_4x2x19x19x10_debug_droundrobin test 
FAIL daley_intel_restart_gx3_16x2x5x10x20_debug_drakeX2 run
FAIL daley_intel_restart_gx3_16x2x5x10x20_debug_drakeX2 test 
FAIL daley_intel_restart_gx3_8x2x8x10x20_debug_droundrobin_maskhalo run
FAIL daley_intel_restart_gx3_8x2x8x10x20_debug_droundrobin_maskhalo test 
~~~

so 5 tests are failing. Apart from first one ("dspacecurve" - see below ), they all abort at the same place in `makemask` when they access tarea (so at the same place as on our custom grid).

Next, I tried to change the loop indices for that loop in makemask :
~~~diff
diff --git i/cicecore/cicedynB/infrastructure/ice_grid.F90 w/cicecore/cicedynB/infrastructure/ice_grid.F90
index 5308236..33cae7f 100644
--- i/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ w/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -1684,8 +1684,14 @@ subroutine makemask
          tarean(:,:,iblk) = c0
          tareas(:,:,iblk) = c0
 
-         do j = 1, ny_block
-         do i = 1, nx_block
+         this_block = get_block(blocks_ice(iblk),iblk)
+         ilo = this_block%ilo
+         ihi = this_block%ihi
+         jlo = this_block%jlo
+         jhi = this_block%jhi
+
+         do j = jlo, jhi
+         do i = ilo, ihi
 
             if (ULAT(i,j,iblk) >= -puny) lmask_n(i,j,iblk) = .true. ! N. Hem.
             if (ULAT(i,j,iblk) <  -puny) lmask_s(i,j,iblk) = .true. ! S. Hem.
~~~

Here I used  `[ilo, ihi], [jlo,jhi]` since I don't see any reason why we should use `[ilo - nghost, ihi + nghost], [jlo - nghost, jhi + nghost]`, since we should not be accessing the `tarea[ns]` arrays outside of  `[ilo, ihi], [jlo,jhi]`.

I then reran `daley_intel_restart_gx3_8x2x8x10x20_debug_droundrobin` and it failed with floating invalid (again, accessing a NaN-initialized location in `to_ugrid`, line 1929:
https://github.com/CICE-Consortium/CICE/blob/c136267896615793c42ec959d03698d5483a1ac8/cicecore%2FcicedynB%2Finfrastructure%2Fice_grid.F90#L1922-L1941

so when accessing `tarea(i, j+1, iblk)`. So this really makes me think that something is fishy with the halo updates in the case of padded cells.

--- 

One last thing, about the "dspacecurve" test: it's failing in a different way, so this might be a different bug, but just to document it here (we might want ot opne a separate issue for that):
~~~
 ice_domain work_unit, max_work_unit =          608          10
 ice_domain nocn =            0        6078      261744
 ice_domain work_per_block =            0          11         570
forrtl: severe (408): fort: (3): Subscript #1 of the array FACTORS has value 0 which is less than the lower bound of 1

cice               00000000022AD476  Unknown               Unknown  Unknown
cice               00000000014E2A96  ice_spacecurve_mp        1451  ice_spacecurve.F90
cice               000000000150E9DA  ice_spacecurve_mp        1727  ice_spacecurve.F90
cice               00000000007327C0  ice_distribution_        2190  ice_distribution.F90
cice               00000000006FF72B  ice_distribution_         121  ice_distribution.F90
cice               000000000074A6B9  ice_domain_mp_ini         551  ice_domain.F90
cice               0000000000DC3AC4  ice_grid_mp_init_         298  ice_grid.F90
cice               0000000000402032  cice_initmod_mp_c         112  CICE_InitMod.F90
cice               0000000000401EB3  cice_initmod_mp_c          52  CICE_InitMod.F90
cice               000000000040168B  MAIN__                     43  CICE.F90
cice               00000000004015F2  Unknown               Unknown  Unknown
cice               0000000002392A4F  Unknown               Unknown  Unknown
cice               00000000004014DA  Unknown               Unknown  Unknown

~~~

phil-blain:
@apcraig yes, this is with Intel. The flags that I use are:
https://github.com/CICE-Consortium/CICE/blob/684e3310b9ce3bd81d929bb21c15447273872571/configuration/scripts/machines/Macros.cesium_intel#L19

GCC/GFortran should have equivalent flag that I do not remember off the top of my head.

I think we should fix the loops (or the halo updates if that is the problem - although as @JFLemieux73 and I were discussing offline last Friday, if the decomp suite is bit for bit, I have trouble understanding how they could be wrong... meaning that if we are accessing non-initialized locations and still getting bit-for-bit results even between padded and non-padded decompositions, then maybe it does not influence the outputs in the end somehow...) 

Anyway I think it's better to not do a band-aid `tarea(:,:,:)=c0`patch and really understand what's going on. I can help with that.

apcraig:
I think we're converging on a plan.  Lets try not to duplicate efforts, do you want me to look into the halo update?  Or how can I help best?

The spacecurve thing has always been a little dicey in it's implementation.  It might be better as a separate issue.  I think it depends what tests we plan to add.

I also agree that the fact that we're getting bit-for-bit answers with different decomposition strongly suggests these problems are not influencing model results.  I think that's allowed us to get away with some of this.

TillRasmussen:
spacecurve failed the same way when I used machine freya and debug
freya_intel_logbfb_gx3_6x2x4x29x18_debug_diag1_dspacecurve_long_maskhalo_reprosum

On topic I think that it is dangerous to initialize to 0. If anything then a huge number in order to make the simulation fail in case that someone creates a loop that enters the padding zone or other areas that are not iniitialized. As you have agreed upon I much prefer not to initialize and then fix the problem when issues are found.

apcraig:
I have duplicated @phil-blain's results.  I have also looked at the halo implementation, at least initially.  The halo points are always next to the ilo/ihi/jlo/jhi indices.  One thing is that the fill in the halo seems to operate assuming no padding, 

```
   do j = 1,nghost
      array(1:nx_block,           j,:) = fill
      array(1:nx_block,ny_block-j+1,:) = fill
   enddo
   do i = 1,nghost
      array(i,           1:ny_block,:) = fill
      array(nx_block-i+1,1:ny_block,:) = fill
   enddo
```
should probably be

```
   do j = 1,nghost
      array(1:nx_block, jlo-j,:) = fill
      array(1:nx_block, jhi+j,:) = fill
   enddo
   do i = 1,nghost
      array(ilo-i, 1:ny_block,:) = fill
      array(ihi+i, 1:ny_block,:) = fill
   enddo
```

I'm looking at the error in the to_ugrid subroutine now.

TillRasmussen:
I think that the 1:ny_block should also be changed to jlo-nghost:jhi+nghost and the same in the other dimension

apcraig:
So, I'm pretty sure that the bug in the fill setting in ice_boundary is causing the problems in the to_ugrid.  Unfortunately, I haven't found a way to fix that bug yet because we don't have access to the list of local blocks, so that will take a bit more work.  I can create an issue.  

I have duplicated all of @phil-blain's issues and fixes.  I think we should proceed as follows
- Set tarea=0 before it's initialized as @phil-blain has proposed until we can get a proper fix into ice_boundary.
- Update the loop indices in makemask to operate only in the non-halo non-padded area
- Add at least one test of debug with padding to our suite, maybe more
- Update cheyenne to include snan initialization with intel
- Look into new debug flags for gnu on cheyenne
- Create an issue about the fill value problems in ice_boundary with padded blocks

@phil-blain, I can do a bunch of this and then PR a change onto your branch.  Or if you prefer to delete this PR, I can create a new one.  We could also merge this PR with what you like and then I can create a new PR with some additional stuff for cheyenne and maybe some new tests.  I do want to test across the full test suite on cheyenne and cheyenne is going to be down the next 2 days, so that will slow me down a bit.  But I'll start to expand my testing and see how that goes.

apcraig:
It looks like some of these changes in makemask are changing answers.  I need to dig a bit further.

apcraig:
I've extended this PR in a branch in my fork and added some new tests turning on debug with various decompositions, updated cheyenne debug to add signaling nans, and fixed the spacecurve bug.  I also did a bunch of validation testing for bit-for-bit.  There are several subtle issues.  I will run a full test suite soon and then maybe we can merge my changes to this branch or create a new PR.

I am going to look at #572 again to see if I can get that implemented properly on the branch.  If I can figure it out quickly, I'll include it.  Otherwise, we can defer it a bit.

phil-blain:
Thanks Tony. Sorry for not being very responsive - I'm only working half days and have a lot going on both at and outside work.... Might be simpler in a new PR. Let's see what you come up with and then we can decide which commits we keep in which PR.

If you update the machines files, could you add the signaling NaNs also for `banting` and `daley` ? I tought I had them activated there but turns out they are not.

Nice to have the spacecurve bug also fixed!

apcraig:
@phil-blain, sounds good and will do.

apcraig:
I ran a full test suite with the "full" debug options and there are a bunch of other failing tests now.  I have fixed a couple things, but can't take everything on at the moment.  I will create a new PR and we need to continue to work on this over time.  I am going to back off on the debug compiler options on cheyenne until all the tests can pass.  I will update the issue related to extending the debug compiler options as well.

apcraig:
Closing for #579

PR #568, opened 2021-03-02
Bug fixed 1d evp

TillRasmussen:

## PR checklist
    Removes all bad departure point bugs
Copied from @ssrethmeier
    The issue with departure errors, resulting from default initialization of full land blocks in the gathering part of the EVP 1D implementation. To solve this, and to avoid gathering and scattering constant arrays, the HTE and HTN arrays (including ghost cells) are now, for the EVP 1D implementation, allocated and given their values during initialization
    The issue with tests not being bit-for-bit comparable when compiled with debug/-O0 optimization, resulting from the order of computations for computing tinyarea not being identical between the EVP 2D implementation and the EVP 1D implementation. This was solved by making the EVP 1D computation order identical to the EVP 2D computation order
    The issue with tests not being bit-for-bit comparable when compiled with debug/-O0 optimization, resulting from icetmask not being a subset of iceumask, as initialily assumed in the EVP 1D implementation. This was solved by adding a skiptcell logical array as well for skipping stress calculations in the EVP 1D implementation

Then we did some modifications to the OpenMP parts of the EVP 1D implementation. The core part of the implementation (stress and stepu) was implemented to be NUMA-aware for performance. This was done differently for the interfacing part of the implementation, but is now aligned throughout the EVP 1D implementation for all of it to be NUMA-aware.

As for namelist changes, the only change that has been made, is the renaming of kevp_kernel to evp_algorithm and changing it from integer to string. EVP 2D is now enabled by setting evp_algorithm = 'standard_2d' instead of kevp_kernel = 0 and EVP 1D by setting evp_algorithm = 'shared_mem_1d' instead of kevp_kernel = 102. In connection with this, the option set_nml.evp1d was also added. Documentation has also been updated to reflect this modification.

Finally, we did a full cleanup of ice_dyn_evp_1d.F90 and some cleanup of ice_dyn_evp.F90. This has mainly included:

    Fixing indentations
    Aligning variable names with the rest of the code base
    Removing old, uncommented code blocks, that was use in development, but not needed any longer

This implementation should work for revised evp and "traditional evp". The first should be tested.
- [x] Developer(s): 
    @TillRasmussen @srethmeier (Stefan Rethmeier, DMI) 
- [ ] @eclare108213 @apcraig @phil-blain @mhrib 
    Removes bad departure point bug bug.
failed test
"restart gbox128 4x2. This test runs but fails to restart exactly.
/results.csh | grep FAIL
FAIL freya_intel_restart_gbox128_4x2_kevp102_short test 
FAIL freya_intel_logbfb_gx3_1x1x50x58x4_diag1_droundrobin_kevp102_maskhalo_reprosum_thread bfbcomp freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
FAIL freya_intel_logbfb_gx3_4x1x25x116x1_diag1_dslenderX1_kevp102_maskhalo_reprosum_thread bfbcomp freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
FAIL freya_intel_logbfb_gx3_1x20x5x29x80_diag1_dsectrobin_kevp102_reprosum_short bfbcomp freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
FAIL freya_intel_logbfb_gx3_8x2x8x10x20_diag1_droundrobin_kevp102_reprosum bfbcomp freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
FAIL freya_intel_logbfb_gx3_6x2x50x58x1_diag1_droundrobin_kevp102_reprosum bfbcomp freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
FAIL freya_intel_logbfb_gx3_6x2x4x29x18_diag1_dspacecurve_kevp102_maskhalo_reprosum bfbcomp freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
 FAIL - test failed
7 of 390 tests FAILED
    - [ ] bit for bit
    - [x ] more substantial 
    - [ ] Yes
    - [ x] No
    - [ ] Yes
    - [ x] No
    - [ ] Yes
    - [x ] No, does the documentation need to be updated at a later time?
        - [x ] Yes
        - [ ] No 
This is a bug fix. The 1d evp solver speeds up the evp solver substantially but is as expected not bit for bit reproducible. If it is chosen the solver will use OMP only. The bug was in the gather method when masked blocks where neigbour of an ocean point. When derivatives were found it got the special value which ny default is a very big number. velocities and stresses are set to zero, whereas HTE and HTN stores a global 2d Array. 
There are alternative choices to fix the bug.
a/ ffil the halo zones in the gather routine with values of neighbouring blocks.
b/ merge the conversion from 2d to 1d and the gather routine.
A bit of cleaning should be done
There should be test including 1devp


apcraig:
I can run a test suite on this before we merge with the 1d evp turned on.  Let me know when you think it's ready.  It seems there are at least a couple things still to do,

- decide if we are happy with the approach
- clean up
- update documentation
- add test case

TillRasmussen:
I updated with the output from the full test suite. It shows (as expected that binary identical results should not be expected).
Jacob and Mads wrote a report which should become an article about this. This showed that bfb should only be expected when low optimization flags were chosen.

TillRasmussen:
note that all test fails when compared to 
freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
Not sure what different-data means

cicecore/cicedynB/dynamics/ice_dyn_evp_1d.F90
```diff
@@ -1186,51 +1185,49 @@ subroutine evp_copyin_v2(nx,ny,nblk,nx_glob,ny_glob,
...
-    call gather_global_ext(G_stressp_4, I_stressp_4, master_task, distrb_info)
-    call gather_global_ext(G_stressm_1, I_stressm_1, master_task, distrb_info)
-    call gather_global_ext(G_stressm_2, I_stressm_2, master_task, distrb_info)
-    call gather_global_ext(G_stressm_3, I_stressm_3, master_task, distrb_info)
-    call gather_global_ext(G_stressm_4, I_stressm_4, master_task, distrb_info)
-    call gather_global_ext(G_stress12_1, I_stress12_1, master_task, distrb_info)
-    call gather_global_ext(G_stress12_2, I_stress12_2, master_task, distrb_info)
-    call gather_global_ext(G_stress12_3, I_stress12_3, master_task, distrb_info)
-    call gather_global_ext(G_stress12_4, I_stress12_4, master_task, distrb_info)
+    call gather_global_ext(G_icetmask,   I_icetmask,   master_task, distrb_info, 0      )
```
eclare108213, line 1188: It's curious that icetmask has 0 at the end of the argument list and iceumask has .false.!  I see that is how they are defined.  No need to fix it here...

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -314,6 +315,7 @@ subroutine input_data
       ndtd = 1           ! dynamic time steps per thermodynamic time step
       ndte = 120         ! subcycles per dynamics timestep:  ndte=dt_dyn/dte
       kevp_kernel = 0    ! EVP kernel (0 = 2D, >0: 1D. Only ver. 2 is implemented yet)
+      pgl_global_ext = .false. ! if true, init primary grid lebgths (global ext.)
```
eclare108213, line 318: spell lengths

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -314,6 +315,7 @@ subroutine input_data
       ndtd = 1           ! dynamic time steps per thermodynamic time step
       ndte = 120         ! subcycles per dynamics timestep:  ndte=dt_dyn/dte
       kevp_kernel = 0    ! EVP kernel (0 = 2D, >0: 1D. Only ver. 2 is implemented yet)
```
eclare108213, line 317: The comment here is kind of confusing.  Does kevp_kernel=102 mean 1D, version 02?  Do we have a version 1?  Should we simplify the kevp_kernel choices?  I remember that we chose 102 because it wasn't really ready, so this was a way to 'hide' it from users, or at least make it less likely they'd set it up.

eclare108213 (changes requested):
This looks pretty straight-forward.  Are there also documentation changes needed?  Suggestions for code modifications below.  I'm a bit confused about the state of the tests.  Are you saying that they should all not be BFB, and there's only one test that doesn't complete and still needs a closer look?  It would be a good idea to run the QC tests comparing standard 2D evp with the 1D version (or did we already do that?).  Thanks for figuring out the issues here.

cicecore/cicedynB/general/ice_init.F90
TillRasmussen: Originally several versions of 1d evp was created. Version 1 included most of the refactoring. Version 2 moved derived grid parameters of HTE and HTN to the 1d solver as a scalar.
Version 3 changed some of the variables from real*4 to real*8. The influence of this was limited change especially when compared to the  accuracy of the iteration. In the end only v2 was implemented. I think that kevp_kernel could be included in kdyn as option 4. I have not thought this thrue

mhrib: Right.
102="Our version 2"
Version 1 and version 2 gave identical results (maybe expect for really aggressive flags, I do not recall exactly). But v2 only takes about half of memory and is a bit faster. For conservative flags we were also able to produce BFB results.
That's not the case for v3, where many internal variables was calculated as real*4. But it was again faster and also takes up less memory. As I recall, the uncertainties calculated using uvel,vvel was less or comparable to the uncertainties obtained across different computational architecture.
v3 is an interesting exercise I think and worth to considering for special cases to reduce memory load (and gain of speed as well).

TillRasmussen:
Sorry about the confusion. I may have confused myself a bit as well.
Main conclusion:
We have removed all bad departure point bugs.
If compiler flags are set to standard (see machine freya) then all test run to the end but there are nunerical differences when testing kevp=102
If compile flag “-no-vec” is added all test are successful. It was not expected that numerical differences was observed when testing with kevp=102 but this means that the order of calculations is important to the result.
We will test the significance of these numerical differences by runing a 1 year simulation and compare concentration and thickness between the test.

This test has was used.

./cice.setup --suite first_suite,base_suite,travis_suite,decomp_suite,reprosum_suite,quick_suite --mach freya --env intel --testid test8-kevp102-0 --bgen baseline8

The following results has been achieved.

EVP-2d
379 measured results of 379 total results
377 of 379 tests PASSED
2 of 379 tests PENDING
 0 of 379 tests MISSING data
0 of 379 tests FAILED
The two pending test ask for processors/node  and our hpc system only has 36, thus they never start

 EVP-1d

Here we have 7 not successful test. They all complete their runs but fail the comparison with other

FAIL freya_intel_restart_gbox128_4x2_short test
FAIL freya_intel_restart_gx3_1x1x50x58x4_droundrobin_thread bfbcomp freya_intel_restart_gx3_4x2x25x29x4_dslenderX2 different-data 
FAIL freya_intel_restart_gx3_4x1x25x116x1_dslenderX1_thread bfbcomp freya_intel_restart_gx3_4x2x25x29x4_dslenderX2 different-data
FAIL freya_intel_restart_gx3_1x20x5x29x80_dsectrobin_short bfbcomp freya_intel_restart_gx3_4x2x25x29x4_dslenderX2 different-data
FAIL freya_intel_restart_gx3_1x4x25x29x16_droundrobin bfbcomp freya_intel_restart_gx3_4x2x25x29x4_dslenderX2 different-data
FAIL freya_intel_logbfb_gx3_1x20x5x29x80_diag1_dsectrobin_reprosum_short bfbcomp freya_intel_logbfb_gx3_4x2x25x29x4_diag1_dslenderX2_reprosum different-data
FAIL freya_intel_smoke_gx3_2x1_run2day_thread bfbcomp freya_intel_smoke_gx3_1x2_run2day different-data

All 7 are gone if compile flag “-no-vec” is turned on.

**Debug**
We found two bugs when adding debug flag to the compilation

1/ spacecurves. This was found in #560 as well

2/ The second is a call to tt = mod(ftime/secday,dayyr). This is in the initialization, thus I expect dayyr to be 0 (I have not checked). This would result in a division by 0.
 
orrtl: error (65): floating invalid
Image              PC                Routine            Line          
cice               0000000000B5BAB1  ice_forcing_mp_in        1242  ice_forcing.F90
cice               0000000000C159D8  ice_forcing_bgc_m         184  ice_forcing_bgc.F90
cice               00000000011AC027  ice_init_column_m         876  ice_init_column.F90
cice               000000000041118C  cice_initmod_mp_i         439  CICE_InitMod.F90
cice               000000000040BC9C  cice_initmod_mp_c         174  CICE_InitMod.F90
cice               000000000040ABAB  cice_initmod_mp_c          52  CICE_InitMod.F90
cice               000000000040A391  MAIN__                     43

eclare108213:
See comments in #575 re kevp_kernel namelist value

apcraig:
Thanks for the recent updates.  I think the next things to do are to get rid of "102" and switch kevp_kernal=1 for the 1d implementation (0 = standard 2d).  This is as suggested in #575.  Actually, what I propose is that we change kevp_kernal to a string called evp_algorithm (or something like it) and have the valid values be "standard_2d" and "shared_memory_1d" (or something like that).  We want to move away from numbers for namelist entries when we can.  kevp_kernal is not really in use yet, so I think it's safe to do that.

Then, we'll want to add a test to turn on the 1d solver.

We can do those things as a separate PR and I can take it on if you like.  I think we want to do it before the release though.

TillRasmussen:
I can implement the evp_algorithm part with the suggested changes. One question not directly related but while I am at it. Is the revised evp included in the eap solver?

apcraig:
I think it depends whether revised evp is it's own standalone flag relative to other flags.  Could we have evp with/without revised evp with/without 1d solver?  If so, it might be nice to keep them as separate flags, otherwise we have to deal with multiple combinations of things being set by 1 flag.

It might be nice to create a tree of the possible dynamics solver options in the documentation, something like (this is proto code only, not necessarily complete/correct).  This assumes it adds to the understanding of what things go together in terms of setting up the dynamics.  Just an idea.

- kdyn=0 (none)
   -   ...
- kdyn=1 (evp)
   - revised_evp=.true.
     - evp_algorithm="standard_2d"
     - evp_algorithm="shared_memory_1d"
   - revised_evp=.false.
     - evp_algorithm="standard_2d"
     - evp_algorithm="shared_memory_1d"
- kdyn=2 (eap)
   -    ...
- kdyn=3 (implicit)
   - algo_nonlin='picard'
     - precond='pgmres'
   -   ...

eclare108213:
If I remember correctly, revp was set up so that it could be used with eap, but I don't think that was ever actually implemented and tested.  So maybe it wouldn't work.  At any rate, revp seems like a fundamentally different approach than the 1D kernel - it's more about how the subcycling iteration is handled, while kevp is more about the vectorization, right?  I'd keep them separate.  I like @apcraig 's outline of dynamics options, for the documentation.  This outline should be essentially what's coded in the verbose diagnostics.

TillRasmussen:
I updated and changed namelist from kevp... to evp_algorithm and changed it from integers to strings with options standard_2d and shared_mem_1d. When updating the documentation to reflect the suggestion by @apcraig  we should also consider to revise the  diagnostic output from the initialization to reflect this

apcraig:
Thanks @TillRasmussen, looks like the documentation needs to be updated (evp_algorithm).

How about also adding a test to the base_suite that turns on the 1d solver.

I guess I should also probably try to run a full test suite on multiple compilers with all tests turning on the 1d solver like I did before and compare to a baseline.  I'll try to get to that.

apcraig:
@TillRasmussen, outside of the documentation and test case, is this ready to merge?  If so, I will try to run a full test suite on my end.  If not, could you let me know what still needs to be done.  thanks.

TillRasmussen:
@srethmeier is doing some cleanup but this should not influence the result. I will update the documentation

eclare108213:
It looks like this is ready to merge, except for some documentation?

TillRasmussen:
just updated documentation. There will be a final update from Stefan soon

TillRasmussen:
ready to merge

apcraig:
I will run a full regression suite on cheyenne with the 1d evp on just to make sure it's behaving as we expect.

apcraig:
Could we rename set_nml.evp_algorithm to set_nml.evp1d?  That maintains the precedent of no special characters in the options name.  It also makes it clearer what the namelist is being set to.

I think we should also add the following to the base test suite, configuration/scripts/tests/base_suite.ts,

smoke          gx3     1x8        diag1,run5day,evp1d       smoke_gx3_8x2_diag1_run5day

I recommend this be added just after 

smoke          gx3     4x1        diag1,run5day,thread      smoke_gx3_8x2_diag1_run5day

This will test that a 1x8 configuration running with evp1d is bit-for-bit with the 8x2 running with evp2d.

TillRasmussen:
cha

> Could we rename set_nml.evp_algorithm to set_nml.evp1d? That maintains the precedent of no special characters in the options name. It also makes it clearer what the namelist is being set to.
> 
> I think we should also add the following to the base test suite, configuration/scripts/tests/base_suite.ts,
> 
> smoke gx3 1x8 diag1,run5day,evp1d smoke_gx3_8x2_diag1_run5day
> 
> I recommend this be added just after
> 
> smoke gx3 4x1 diag1,run5day,thread smoke_gx3_8x2_diag1_run5day
> 
> This will test that a 1x8 configuration running with evp1d is bit-for-bit with the 8x2 running with evp2d.

Done

apcraig:
Testing results seem reasonable, maybe.

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#896fff6559501fecd415fa2c4862898adf3540bc

The "all green" results are just running this PR against the current master.  Everything passes and is bit-for-bit.  That means nothing has broken or changed.  The next test results with 100+ compare failures is the same test suite but forcing the evp 1d solver on for everything.

- The tx1 test fails for every compiler, as expected.  There is an abort if using the evp 1d with tripole.
- The regression tests mostly fail.  This is a little surprising.  At least the debug tests should pass.  This seems to contradict prior results and recent results.  Could some of the clean up have broken something.  Has @TillRasmussen checked the current version of the model with the clean up?
- The cheyenne_intel_restart_gbox128_4x2_short fails to restart exactly.  This is what @TillRasmussen found as well.  It was the only test that failed with evp 1d.  Do we have any idea why?  The case runs to completion fine, it just does not restart exactly.  All other intel restart tests as well as all pgi and gnu restart tests do restart exactly.  This is an odd outlier.

I also tried to find a case where the 1d evp was bit-for-bit with the 2d results.  Even with the debug on, I could not find a case with the gx3.  The only cases that are bit-for-bit are ones that use eap or dynamnics off or that have evp1d on by default.  Again, I am concerned about these results as they are not quite as expected.

apcraig:
@TillRasmussen @srethmeier Could you setup a case or two and verify you get bit-for-bit with the -no-vec option for the 1d and 2d solvers.  I am not seeing that in my tests on cheyenne when I test with the debug (-O0) build options.  I'm wondering if this is a machine issue or if something has changed in the recent cleanup of the code.  One of the basis for accepting this implementation is that is produces bit-for-bit results at reduced compiler optimization.  If that's no longer the case, we may need to do a bit more validation effort.

TillRasmussen:
@apcraig We are currently rerunning test and adding complexity to these in order to verify status.

eclare108213:
@TillRasmussen 
We are very close to the release now - how are things going with this ?  Would love to get it in there too...  Anything others of us can do to help?   We can also hold off if there are suspicious weirdnesses that need to be sorted out. Thx

TillRasmussen:
Status without access to the computer where I have run the test (personal memory leaks may be present). After @apcraig comments I updated the code to include the changes that are committed recently to the CICE master. When the ocean forcing from NCAR is read I have set the mixlayer depth to 20 to bypass the issue discovered by the end of last week.
In addition to this I realized that there is a ambiguity in the definition of indexes. ne, se, ee,  nw, sw, sse (ice_dyn_evp1d.F90). These are both global module values to the evp_1d solver and local to subroutines like stepu_iter, stepu_last, stress_iter and stress_last.
The current call to step_u (last and iter) converts the name of sse (global in call) to se (local in routine). This means that locally I was not sure whether it used se or sse. I changed the name of the local variable to sse. The test I ran was with this setting, extended walltime  and forced debug. I ran both Gnu and Intel full test suite according to settings of the machine Freya. For Intel I add -no-vec.

I assume that the test that returns FAIL and compare are the ones that compare 2 test suites. 
The results were as follow.

2d of CICE master and 2d of merge request:
Approximately 14test failed with bit to bit comparisons between test within the same testsuite.  
This were the same test for both compilers (7 each - there may be one additional test in either GNU or intel that fails.) The result is the same for the current CICE master and the merge request. 

When comparing bit for bit between  1d and 2d test I had approximately ¼ (close to 40 test) that passed and ¾ that failed. I have not diagnosed the reason but there are several test that were not run as 1d or without  dynamics in the standard test. In addition I am not sure whether it is expected that all test are bit for bit.

After I changed the indexes I got 1 FAILED during runtime somewhere in the ice_remap file. I cant remember where but if I remember correctly this was not a bad departure point. It is a new bug and it was only captured by the intel compiler.

Short conclusions
I am unsure why the compiler accept both a global and a local variable with the same name. I need to remove these double definitions and redo the test.
1 run time bug was discovered when running 1d.
A limited number of test fails when running 2d (only bit for bit test).
There are test that pass as bit for bit when comparing 1d and 2d but more test fail.

I will be able  to access the hpcsystem on Monday. I will update results then.

TillRasmussen:
Status update.
With the current merge request 40 test pass the comparison between 1d and 2d and 191 fails. I have not been able to find a pattern.
@srethmeier and I have discussed this a bit and we think that they should be identical
1 test fails at run time. I have tracked this back to the initial conditions of the boundary of the domain. I still need to do a bit of diagnostics to eliminate this. The run that fails at runtime is freya_gnu_smoke_gx3_16x2x1x1x800_debug_droundrobin_run2day. It is special in the sense that the block size is 1x1.
In terms of bringing it into the release 6.2 we are working on it and have some ideas but debugging is not always strait forward.  We have verified that it is not the cleanup that has caused the problems.

eclare108213:
@TillRasmussen @srethmeier 
What is the status here?  It sounds like there's one 'real' bug to find, which causes a runtime failure, and a bunch of non-BFB effects that might or might not be bugs.  Does optimization play a role in those?  Anything others of us could do to help out?  If the non-BFB problems persist, maybe run the QC and see if they matter?  Sometimes we just have to accept ones that are unavoidably due to rearrangement of the calculation order.

TillRasmussen:
I have been testing with all debug flags on for the machine "freya" with compilers intel, cray and gnu. This compiler choise did not make any difference.

There is one runtime failure however this is with block size 1x1. I think that it was mentioned that there is an issue.
freya_gnu_smoke_gx3_16x2x1x1x800_debug_droundrobin_run2day

Almost all gbox cases are bit for bit identical
except freya_gnu_restart_gbox80_1x1_box2001
Some gx 3 cases are bit for bit (20 pass and 177 fall
All gx1 cases are not bit for bit
I have been debugging  freya_gnu_restart_gbox80_1x1_box2001 
and freya_gnu_smoke_gx3_1x1_debug_run1day
This were chosen as they only have one block
In order to compare I have made restart output every hour (=timestep). For gx3 it is binary identical until time step 5. From then and onwards they are not bit for bit..
Same pattern was seen with the gbox caae. Here it ran for 7-8 days with standard ndte=240 . In this case I reduced the number of iterations in the evp solver and then it was bit for bit for the full duration of the simulation.

I need to look a bit more into this, however it seems as if the issue occurs near the ice edge.
Suggestions are welcome on what to try.

eclare108213:
-O0 ?

TillRasmussen:
cray compiler
FFLAGS     += -O0 -g -Ofp0 -K trap=fp -Rbcdps
Intel
FFLAGS     += -O0 -g -check -fpe0 -ftrapuv -fp-model except -check noarg_temp_created
gnu
FFLAGS     += -O0 -g -fcheck=bounds -finit-real=nan -fimplicit-none -ffpe-trap=invalid,zero,overflow

Gnu and intel are available in the main repository as
CICE/configuration/scripts/machines/Macros.freya_intel and CICE/configuration/scripts/machines/Macros.freya_gnu

apcraig:
I have been doing some debugging on the 1d evp and have found an odd situation.  A case I'm testing diverges (between the 2d and 1d solvers) several days into the simulation.  I understand why, but am not sure how to interpret the difference.

At a single point on that timestep, the "icetmask" is false but the "iceumask" is true.  So far as I can tell, that means that on that gridcell, at that timestep, in the 2d solver, the stress part of the solver will NOT be called for that gridcell but the stepu part of the solver will be called.  Is that ever supposed to happen?

In the 1d solver, all the masking is based on the icetmask, so at that gridcell, at that timestep, the 1d solver is NOT doing the "stress" part and it's also NOT doing the "stepu" part.  So the 2d and 1d solvers diverge. 

This case where icetmask is false and iceumask is true at a gridpoint is apparently not very common which is why I'm wondering if it should happen at all.  If it shouldn't, we need to look into the logic that's creating that situation and it would be a bug in the dynamics that we need to fix.  If it can happen, then that means the 1d solver needs to be extended to take into account different masking for the stress and stepu parts of solver.  By the way, this is an interior point and has nothing to do with the north and east boundary.

Let me know if anyone has any thoughts about this situation.

TillRasmussen:
Good carch. It is correct that 1d asumes that ucells are a subgroup of t cells. Stefan and I have discussed this but we did not find an example where this is not true , however we did not debug all test. Changing this wiould not be too difficult.

apcraig:
Thanks @TillRasmussen.  I'll assume you'll take care of the recoding of the masking then to account for the T vs U masks in the 1d implementation.  I'm happy to do some more testing when that's implemented.  Hopefully, that'll fix another handful of the differences.

eclare108213:
> This case where icetmask is false and iceumask is true at a gridpoint is apparently not very common which is why I'm wondering if it should happen at all. If it shouldn't, we need to look into the logic that's creating that situation and it would be a bug in the dynamics that we need to fix. 

This is interesting.  If there's ice in a grid cell (icetmask (i,j)=T), then all of the corners of that grid cell should be free to move (iceumask=T for i-1,j; i,j-1; i-1,j-1) as long as they aren't on a coastline.  Since it's a B-grid,  that means that the upper right-hand corner of some grid cells without ice might be moving because there's ice in a neighboring cell.  I think this is correct.  What would not be correct is icetmask=F because it's a land cell, with iceumask=T for that cell.

apcraig:
I can confirm that at this gridcell, most of the time, both T and U masks are either both false or both true.  In this case, both were false for a day or two at the start of the run, then both became true for a day or two.  Then for a single timestep, the T mask switched to false while the U mask remained true.  Then both masks went back to true.  Sounds like it's not unreasonable for this to happen, but that it's relatively rare.  Thanks.

mhrib:
Jacob and I also had the same considerations when the 1d solver was designed for some years ago. We came to the conclusion, that the u-cells will always be a subgroup of the t-cells. As I recall - not sure - Elisabeth also confirmed that.

This makes the logics for the 1d solver much easier, as all variables ranges 1:NA with the exception, that uvel/vvel and stress 1-8 has exactly the same points, but extended with extra points (NAVEL-NA+1) to take care of finite differences. Therefore the logics forming contiguous vectors was relative easy without running through both ucells and tcells.

Perhaps something has changed since outside the 1d block? My suggestion would be to start looking at the 2D structure, especially at the boundaries between blocks including halos.




apcraig:
@mhrib, I don't think this is a halo issue, just a masking issue.  This seems to come down to the naming/numbering convention we use for T and U points.  Actually, it seems like anytime there is ice in one gridcell and but no ice in the west or southwest neighbor, this condition would be triggered on one of those ice-free neighbors.  I'm surprised this condition is all that rare.  My question is, why aren't there dozens of gridcells every timestep in this state at the ice edge?

mhrib:
UPDATE: Wrote my comments while Elisabeth answered above. If I understand correctly, when there is only one potential candidate for icetmask=0 and iceumask=T and that is on the upper right hand corner.

Nevertheless, I think a quick - and correct - fix would be to search for (icetmask==0 .and. iceumask==T)-pairs and set icetmask=1 for these specific points within the evp_copyin_v2 subroutine. Then the 1D solver would take care of the logics within subroutine calc_2d_indices

mhrib:
@apcraig Agree, that it is surprisingly, that we have not seen it before.
If I understand correctly, it can only happen within the upper right corner within the full domain. For GX3 domain this is a point within the Arctic domain, where ice is almost always present. And the test-domain that Jacob and I used has closed boundaries, ie. the situation will never happen here.... Just my quick thoughts...

apcraig:
@mhrib, the point that I debugged was in the middle of the domain.  It was point 64, 94 on the gx3 grid, in the Gulf of Alaska.  This also seems to happen primarily with certain forcings.  We don't have any cases with this happening with the JRA55 forcing at 2005.  If this "mask error" is creating the 2d vs 1d differences in the gx3ncarbulk, yi2008, and gx1coreii cases, then it seems to depend on the forcing as well somehow.

mhrib:
@apcraig Thanks for pointing out, that it can also happen "in the middle" of the domain. This surprises me, but I'll have to think about why.
Perhaps certain forcing could tricker isolated 1 cell ice patches, but this should absolutely be a valid solution in my eyes although it may not happen very often. The only problem I can see wrt. forcing and advection is a too high timestep, that allows the advection to move more than one grid-cell. However, this should still not affect the EVP part as I can see.

Nevertheless, I think my suggested fix above will solve the differences between 2D and 1D EVP solver due to icetmask=0 and iceumask=T cases

srethmeier:
Thanks for taking a crack at this, @apcraig. Greatly appreciated! As @TillRasmussen mentioned, we actually discussed this at some point, since this is one point where the two implementations differ. I'll have to dig into my notes to see what we actually did test and why we concluded that it was not the problem. Maybe we got a little confused by the fact, that there have been several issues resulting i deviating test results. But good that we now know where it stems from.

So the consensus is that the current 2d implementation is correct?   

I have some other things I need to finish before my vacation next week, but i will dig into this and try to fix it as soon as possible. Keep you posted about it.

apcraig:
Thanks @srethmeier.  For now, we think the 2d implementation is correct.  I think you are best positioned to update the implementation, but if you want me to give it a shot, let me know.  I would probably attempt to do what @mhrib proposed as a start.  Let me know when you have some updates pushed and I can retest.  Thanks!

apcraig:
I  have a couple other suggested modifications for the current implementation.

This will improve the information about evp_algorithm a bit,

```
+++ b/cicecore/cicedynB/general/ice_init.F90
@@ -1298,14 +1298,14 @@ fpfunc_andacc = 1      ! fixed point function for Anderson acceleration: 1: g(x)
                write(nu_diag,1010) ' revised_evp      = ', revised_evp,trim(tmpstr2)
        
                if (evp_algorithm == 'standard_2d') then
-                  tmpstr2 = ' : standard EVP solver'
+                  tmpstr2 = ' : standard 2d EVP solver'
                elseif (evp_algorithm == 'shared_mem_1d') then
-                  tmpstr2 = ' : vectorized EVP solver'
+                  tmpstr2 = ' : vectorized 1d EVP solver'
                   pgl_global_ext = .true.
                else
                   tmpstr2 = ' : unknown value'
                endif
-               write(nu_diag,1030) ' evp_algorithm    = ', trim(tmpstr2)
+               write(nu_diag,1031) ' evp_algorithm    = ', trim(evp_algorithm),trim(tmpstr2)
                write(nu_diag,1020) ' ndtd             = ', ndtd, ' : number of dynamics/advection/ridging/steps per thermo 
```

There are indentation issues in ice_dyn_evp.F90, 

```
@@ -377,10 +387,11 @@ subroutine evp (dt)
           stressp_1 ,stressp_2, stressp_3, stressp_4,                   &
           stressm_1 ,stressm_2, stressm_3, stressm_4,                   &
           stress12_1,stress12_2,stress12_3,stress12_4                   )
-          call ice_timer_start(timer_evp_1d)
-          call ice_dyn_evp_1d_kernel()
-          call ice_timer_stop(timer_evp_1d)
-          call ice_dyn_evp_1d_copyout(                                      &
+        call ice_timer_start(timer_evp_1d)
+        call ice_dyn_evp_1d_kernel()
+        call ice_timer_stop(timer_evp_1d)
+        call ice_dyn_evp_1d_copyout(                                      &
```

Finally, I propose that the "do i = 1, ndte - 1" loop in ice_dyn_evp_1d_kernel be renamed to "do ksub = 1, ndte - 1".  That would better mimic the subcycling loop in the 2d implementation and would help tie them together a little clearer.

Thanks!

mhrib:
@apcraig I do not have time to look into it right now, but my suggestion would be something like this:

Add loop within subroutine evp_copyin_v2, that loops over all iceumask:

About 1237:
...
```fortran
    !-- All calculations has to be done on the master-task --

      !-- Check (icetmask==0 .and. iceumask==T)-pairs
      do j=1,ny_glob
        do i=1,nx_glob
          if (icetmask(i,j)==0 .and. iceumask(i,j)==.true.) then
            icetmask(i,j)=1
          endif
        enddo
      enddo
      !-- Find number of active points and allocate vectors --
      ...
    endif
```

apcraig:
I think I have something working.  I will need to test more.  It turns out we need to add a skiptmask like there is a skipumask.  We need to be able to separate logic for doing stress and stepu computations.  

Given that Till is away for several weeks, would it make sense for me to create a new PR with the latest updates including everything in this PR, to delete this PR, and for us to merge a new PR.  Alternatively (and my preference) would be that I PR to Till's branch, that gets merged into this PR, and that we merge this PR to the Consortium.  @srethmeier, are you able to merge a PR to Till's branch if I create it?  I know you are also busy and about to go on vacation.  Let me know what would work best.  My preference is to get this merged once it's working rather than wait a month when everyone is back.

srethmeier:
@apcraig, really great if you have something working! I was just starting to look a little bit at it, but can see my setup has become a little bit outdated in the mean time. Yes, I should be able to merge PRs to @TillRasmussen's branch. Also think that's preferred so that we have the history.

I'm also in the office Monday and Tuesday and already back the week after next again.

apcraig:
Thanks @srethmeier, I should be able to test today and if all goes well, create the PR to @TillRasmussen branch.   Once you merge my PR, I'll update the PR with some comments and test results and then we can probably merge the PR to the Consortium master.  Fingers crossed.  Thanks.

srethmeier:
> I have a couple other suggested modifications for the current implementation.
> 
> This will improve the information about evp_algorithm a bit,
> 
> ```
> +++ b/cicecore/cicedynB/general/ice_init.F90
> @@ -1298,14 +1298,14 @@ fpfunc_andacc = 1      ! fixed point function for Anderson acceleration: 1: g(x)
>                 write(nu_diag,1010) ' revised_evp      = ', revised_evp,trim(tmpstr2)
>         
>                 if (evp_algorithm == 'standard_2d') then
> -                  tmpstr2 = ' : standard EVP solver'
> +                  tmpstr2 = ' : standard 2d EVP solver'
>                 elseif (evp_algorithm == 'shared_mem_1d') then
> -                  tmpstr2 = ' : vectorized EVP solver'
> +                  tmpstr2 = ' : vectorized 1d EVP solver'
>                    pgl_global_ext = .true.
>                 else
>                    tmpstr2 = ' : unknown value'
>                 endif
> -               write(nu_diag,1030) ' evp_algorithm    = ', trim(tmpstr2)
> +               write(nu_diag,1031) ' evp_algorithm    = ', trim(evp_algorithm),trim(tmpstr2)
>                 write(nu_diag,1020) ' ndtd             = ', ndtd, ' : number of dynamics/advection/ridging/steps per thermo 
> ```
> 
> There are indentation issues in ice_dyn_evp.F90,
> 
> ```
> @@ -377,10 +387,11 @@ subroutine evp (dt)
>            stressp_1 ,stressp_2, stressp_3, stressp_4,                   &
>            stressm_1 ,stressm_2, stressm_3, stressm_4,                   &
>            stress12_1,stress12_2,stress12_3,stress12_4                   )
> -          call ice_timer_start(timer_evp_1d)
> -          call ice_dyn_evp_1d_kernel()
> -          call ice_timer_stop(timer_evp_1d)
> -          call ice_dyn_evp_1d_copyout(                                      &
> +        call ice_timer_start(timer_evp_1d)
> +        call ice_dyn_evp_1d_kernel()
> +        call ice_timer_stop(timer_evp_1d)
> +        call ice_dyn_evp_1d_copyout(                                      &
> ```
> 
> Finally, I propose that the "do i = 1, ndte - 1" loop in ice_dyn_evp_1d_kernel be renamed to "do ksub = 1, ndte - 1". That would better mimic the subcycling loop in the 2d implementation and would help tie them together a little clearer.
> 
> Thanks!

No problem :)

apcraig:
My PR will includes some of my proposed minor clean up things as well.

srethmeier:
> Thanks @srethmeier, I should be able to test today and if all goes well, create the PR to @TillRasmussen branch. Once you merge my PR, I'll update the PR with some comments and test results and then we can probably merge the PR to the Consortium master. Fingers crossed. Thanks.

Great ... well think we're the ones saying thank you here :)

apcraig:
@srethmeier, there is a PR to @TillRasmussen's branch.  I think everything is working as expected now.  All test results can be seen here with the latest PR, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#126245535038576bb68746004754915fe96ee06c.

Full test suites were run on cheyenne with intel, pgi, and gnu.  All of the standard test suite passes with 0 test failures.  This tests the default 2d solver (except for alt04 which uses the 1d solver).  The 19 regression failures for each compiler in this mode consist of alt04, 1x1 and 2x2 blocks, and unit tests.  All are expected to fail regression in these runs.

In addition, a full suite was run with the 1d evp solver vs 2d evp solver with debug turned on for all tests in both cases.  The debug flag reduces optimization which makes it more likely to get bit-for-bit results with 1d vs 2d evp.  These tests also all seem to pass as expected.  10, 3, and 5 tests (out of 212) fail for each compiler due to timing out because the debug flag runs the model slower.   Because the number of timed out tests was small, they were ignored.  Otherwise, all tests pass.  18, 189, and 18 regression tests fail for the same reasons as above, alt04, 1x1 and 2x2 blocks, and unit tests.  In the case of the pgi compiler (189 regression failures), even reduced optimization with the debug flag leads to non bit-for-bit changes due to the compiler.  Had intel and gnu also failed the same way, it would be difficult to confirm that the 2d and 1d results were identical (with reduced optimization).  But in this case, intel and gnu confirm bit-for-bit at reduced optimization.

As far as I'm concerned, the evp 1d solver is now fully validated with no outstanding issues once https://github.com/TillRasmussen/CICE/pull/17 is merged to this PR.

doc/source/science_guide/sg_dynamics.rst
```diff
@@ -734,3 +733,13 @@ Finally, as with the classic EVP approach, the stresses are initialized using th
...
+
+The scalability of geophysical models is in general terms limited by the memory usage. In order to optimize this the 1d EVP solver solves the same equations that are outlined in the section :ref:`stress-evp` but it transforms all matrices to vectors (1d matrices) as this complies better with the computer hardware. The vectorization and the contiguous placement of arrays in the memory makes it easier for the compiler to optimize the code and pass pointers instead of copying the vectors. The 1d solver is not supported for tripole grids and the code will abort if this combination is attempted.
```
apcraig, line 745: I think complies should be compiles?

as this compiles better with the computer hardware

apcraig (changes requested):
Just a few suggestions for minor revisions in the documentation at this point.

srethmeier:
@apcraig Everything looks good to me. I will do some tests in our end as well, but have merge your changes "as is".

I agree with your minor doc revisions. They will follow in a separate commit.

Anything else we need to take a look at? @TillRasmussen, @mhrib?

mhrib (approved):

mhrib:
@apcraig Implementation looks good using skiptmask similar to skiptmask. Readable and "cycle" should keep the computational performance. Nice error catch by the way. Thanks for your efforts.

apcraig (approved):

apcraig:
It looks like we're close to merging this.  @srethmeier, let me know if you see/found any issues in your testing or review.   @eclare108213 and others, if there are any other comments/reviews, please suggest.  Thanks!

srethmeier:
@apcraig, I haven't found any issues in my testing, so we should be good to go.

@eclare108213 and @apcraig, I believe we have addressed all your change requests, but let me know if that's not the case, or if you should have any additional ones. Then I'll try to get it fixed tomorrow or at the latest when I'm back in the beginning of next week.

Again, thanks for helping out with this issue!

eclare108213:
I've just scanned the whole conversation, and there are a lot of different elements in this PR, including multiple bug fixes.  Could one of you go through and summarize the changes, all in one place?  Include bug fixes, changes to namelist, optimization effects, etc?

apcraig:
I will do a QC test of 1d vs 2d with standard optimization and report back.

Looking back at the PR, these modifications primarily fix issues in the 1d evp that prevented identical results with 2d (at reduced optimization).  That includes at least a fix to the tiny area computation, addition of grid_lengths_global_ext, and addition of skiptcell logic.  In addition, the namelist was changed so we now have strings instead of integers to set evp_algorithm, and a bunch of other formatting (i.e. indentation) was cleaned up.  There may have been a few other changes, maybe @srethmeier, @mhrib, or @TillRasmussen can comment.

The issues related to 1x1 and 2x2 blocks, calendar, and spacecurve were addressed elsewhere. 

I agree there is still an outstanding issue about what parts of the dynamics work with each other (i.e. revp, eap plans, etc).  I think it should also include overall discussion of control, namelist, and implementation of the dynamics.  I have created a new issue, #619 that can serve as a general place holder for those issues.  This is something that was raised recently in our monthly meetings as well.

apcraig:
The qc test with the 1d evp fails at the end of the 4th year with zap snow temperature errors, 

```
  (zap_snow_temperature)Tmin:  -100.000000000000
  (zap_snow_temperature)Tmax:  4.486434370863951E-006
  (zap_snow_temperature)zqsn:  -182365602.580354
  (zap_snow_temperature)zap_snow_temperature: temperature out of bounds!
  (zap_snow_temperature)k:           1
  (zap_snow_temperature)zTsn:  -103.894744835945
  (zap_snow_temperature)Tmin:  -100.000000000000
  (zap_snow_temperature)Tmax:  5.419524611890995E-006
  (zap_snow_temperature)zqsn:  -182424769.766085
```

The baseline 2d runs fine.  This suggests there still may be some issues with the 1d implementation.  I actually ran it twice, once with 9x4 pes (threaded) and another time with 36x1 (not threaded) and both failed at the same time in the same way.  I guess that means it's a robust problem.

I still suggest we merge this but then create a new issue noting the new problem discovered.

srethmeier:
> I will do a QC test of 1d vs 2d with standard optimization and report back.
> 
> Looking back at the PR, these modifications primarily fix issues in the 1d evp that prevented identical results with 2d (at reduced optimization). That includes at least a fix to the tiny area computation, addition of grid_lengths_global_ext, and addition of skiptcell logic. In addition, the namelist was changed so we now have strings instead of integers to set evp_algorithm, and a bunch of other formatting (i.e. indentation) was cleaned up. There may have been a few other changes, maybe @srethmeier, @mhrib, or @TillRasmussen can comment.
> 
> The issues related to 1x1 and 2x2 blocks, calendar, and spacecurve were addressed elsewhere.
> 
> I agree there is still an outstanding issue about what parts of the dynamics work with each other (i.e. revp, eap plans, etc). I think it should also include overall discussion of control, namelist, and implementation of the dynamics. I have created a new issue, #619 that can serve as a general place holder for those issues. This is something that was raised recently in our monthly meetings as well.

@apcraig, I believe that actually summarizes all major things covered in this PR. We mainly fixed three issues:
1. The issue with departure errors, resulting from default initialization of full land blocks in the gathering part of the EVP 1D implementation. To solve this, and to avoid gathering and scattering constant arrays, the `HTE` and `HTN` arrays (including ghost cells) are now, for the EVP 1D implementation, allocated and given their values during initialization
2. The issue with tests not being bit-for-bit comparable when compiled with `debug`/`-O0` optimization, resulting from the order of computations for computing `tinyarea` not being identical between the EVP 2D implementation and the EVP 1D implementation. This was solved by making the EVP 1D computation order identical to the EVP 2D computation order
3. The issue with tests not being bit-for-bit comparable when compiled with `debug`/`-O0` optimization, resulting from `icetmask` not being a subset of `iceumask`, as initialily assumed in the EVP 1D implementation. This was solved by adding a `skiptcell` logical array as well for skipping stress calculations in the EVP 1D implementation

Then we did some modifications to the OpenMP parts of the EVP 1D implementation. The core part of the implementation (stress and stepu) was implemented to be NUMA-aware for performance. This was done differently for the interfacing part of the implementation, but is now aligned throughout the EVP 1D implementation for all of it to be NUMA-aware. 

As for namelist changes, the only change that has been made, is the renaming of `kevp_kernel` to `evp_algorithm` and changing it from integer to string. EVP 2D is now enabled by setting `evp_algorithm = 'shared_mem_1d'` instead of `kevp_kernel = 0` and EVP 1D by setting `evp_algorithm = 'shared_mem_1d'` instead of `kevp_kernel = 102`. In connection with this, the option `set_nml.evp1d` was also added. Documentation has also been updated to reflect this modification.

Finally, we did a full cleanup of `ice_dyn_evp_1d.F90` and some cleanup of `ice_dyn_evp.F90`. This has mainly included:
- Fixing indentations
- Aligning variable names with the rest of the code base
- Removing old, uncommented code blocks, that was use in development, but not needed any longer

Maybe @TillRasmussen can update the PR checklist with this, so that it is at the top of the PR? I can't modify it.

srethmeier:
> The qc test with the 1d evp fails at the end of the 4th year with zap snow temperature errors,
> 
> ```
>   (zap_snow_temperature)Tmin:  -100.000000000000
>   (zap_snow_temperature)Tmax:  4.486434370863951E-006
>   (zap_snow_temperature)zqsn:  -182365602.580354
>   (zap_snow_temperature)zap_snow_temperature: temperature out of bounds!
>   (zap_snow_temperature)k:           1
>   (zap_snow_temperature)zTsn:  -103.894744835945
>   (zap_snow_temperature)Tmin:  -100.000000000000
>   (zap_snow_temperature)Tmax:  5.419524611890995E-006
>   (zap_snow_temperature)zqsn:  -182424769.766085
> ```
> 
> The baseline 2d runs fine. This suggests there still may be some issues with the 1d implementation. I actually ran it twice, once with 9x4 pes (threaded) and another time with 36x1 (not threaded) and both failed at the same time in the same way. I guess that means it's a robust problem.

Okay, looks like there still is something we need to look at here. I'll try and dig into it next week.

> I still suggest we merge this but then create a new issue noting the new problem discovered.

Sounds good to me. Whatever you prefer and find to be the best approach.

srethmeier:
> The rEVP question needs a github issue for follow up -- can this vectorization technique be run for that case, and if so, does it work in the current code? I think there's already an issue for "kernelizing" EAP.

Not complete sure what this entails? Is it that the combination of Revised EVP and EVP 1D has not been tested?

TillRasmussen:
Revised evp should work

TillRasmussen:
Revised evp should work

eclare108213:
> > The rEVP question needs a github issue for follow up -- can this vectorization technique be run for that case, and if so, does it work in the current code? I think there's already an issue for "kernelizing" EAP.
> 
> Not completely sure what this entails? Is it that the combination of Revised EVP and EVP 1D has not been tested?

Yes, if rEVP and 1D can be run together, then test to make sure it works.  Otherwise put an abort in the code in case they're both turned on.  

Thanks for summarizing all the changes in the PR.  I'd like to better understand which, if any, change the answers when 1D EVP is turned off.  @apcraig 's testing showed regression failures for alt04, 1x1 and 2x2 blocks, and unit tests, plus some that timed out due to the low optimization level.  Please confirm that you're confident that all of these failures are expected, and that none of them affect the 2D EVP results.  My preference would be to find and fix the problem causing snow temperatures to explode, but I'm fine with merging this PR and then fixing that problem.  Thanks everyone -- 1D EVP has been a huge effort!

apcraig:
Regarding the testing and failed tests.  All of the 2d tests are bit-for-bit with this PR.  This PR has no impact on the standard 2d results which are most of our test suite.  

The failed regression tests in the 1d (with this PR) vs 2d (with master) with debug on are also all expected.  It's also only the regression (comparison with baseline) part of the tests that fail, the tests themselves pass. The alt04 test fails because it has the 1d evp on by default and we have changed answers.  The 1x1 and 2x2 tests fail because my sandboxes did not include the recent fixes.  The unit tests fail because there continue to be a couple of nagging (but still bit-for-bit) issues in the regression testing on unit tests.  The tests that timed out could be rerun with longer submission times and I will try to do some of that today to see.

The aborted QC test case is unfortunate.  It could be that the roundoff errors introduced by the 1d evp have changed the state of the solution such that an abort is created.  I don't think this is the most likely case because then we'd be encountering this as well with 2d cases.  It could be that there are still some differences in 1d and 2d that would appear if we ran both cases with debug on for 5 years.  We really only tested bit-for-bit with debug on for days or months.  However, if all we are doing is introducing a roundoff difference, why is the model aborting.  It should be more robust than that.   It's surprising to me with all the testing that we've done that the QC evp 1d run aborted.

apcraig:
I submitted most of the 1d tests that timed out for intel and gnu with more time and they all pass including regression vs 2d.  So I think all the 1d vs 2d debug results are accounted for now with no outstanding issues.  The only known issue is the QC failure after 4 years (out of a 5 year run).

TillRasmussen:
> I think I have something working. I will need to test more. It turns out we need to add a skiptmask like there is a skipumask. We need to be able to separate logic for doing stress and stepu computations.
> 
> Given that Till is away for several weeks, would it make sense for me to create a new PR with the latest updates including everything in this PR, to delete this PR, and for us to merge a new PR. Alternatively (and my preference) would be that I PR to Till's branch, that gets merged into this PR, and that we merge this PR to the Consortium. @srethmeier, are you able to merge a PR to Till's branch if I create it? I know you are also busy and about to go on vacation. Let me know what would work best. My preference is to get this merged once it's working rather than wait a month when everyone is back.

@apcraig Thank you for implementing this. I agree that a skiptmask is needed.

TillRasmussen:
> I will do a QC test of 1d vs 2d with standard optimization and report back.
> 
> Looking back at the PR, these modifications primarily fix issues in the 1d evp that prevented identical results with 2d (at reduced optimization). That includes at least a fix to the tiny area computation, addition of grid_lengths_global_ext, and addition of skiptcell logic. In addition, the namelist was changed so we now have strings instead of integers to set evp_algorithm, and a bunch of other formatting (i.e. indentation) was cleaned up. There may have been a few other changes, maybe @srethmeier, @mhrib, or @TillRasmussen can comment.
> 
> The issues related to 1x1 and 2x2 blocks, calendar, and spacecurve were addressed elsewhere.
> 
> I agree there is still an outstanding issue about what parts of the dynamics work with each other (i.e. revp, eap plans, etc). I think it should also include overall discussion of control, namelist, and implementation of the dynamics. I have created a new issue, #619 that can serve as a general place holder for those issues. This is something that was raised recently in our monthly meetings as well.

Done

TillRasmussen:
I can confirm that some test fail due to wall time on the DMI system as well. All were fixed by increasing wall time.
I would also recommend that this merge is accepted. The only issue is the QC test that we will look  at.  We should open an issues that explains this. I think that this would make it more clear
- This should not change the 2d solution and it did not do so before the changes by @apcraig. I will rerun the test with the new changes.
-  The difference between the 1d and the 2d soiution is only in the subcycling loop. The point is to avoid memory usage when calling the two functions multiple times and to avoid mpi communications (multiple), which both scale poorly. The conversion between 1d and 2d may have some overhead, therefore the benefit may not be that clear in a test domain like gx3 which is relatively small as it may not be memory bounded the same way as the larger domains. This conversion is only needed as long as CICE is a 2d model. Currently this is only the dynamical part of CICE where this is required.
- EAP: is not currently implemented. An implementation of this would reuse the conversion from 2d to 1d. The functions within would need to be rewritten to 1d.
- ESMF now enable separation of different modules (ocean, sea ice, atmosphere.... ) onto different different nodes with different hybrid versions of OMP+MPI. This should enable running CICE on its own node and thereby enable this to run only CICE on OMP only. I have not tested this. Further modulization of CICE could increase the usage of this.,

apcraig:
Does anyone object to merging this at this point?  We should create a new issue as part of the merge to track outstanding issues.

eclare108213 (approved):

eclare108213:
I'm okay with merging this, although the QC failure is a nagging issue.  Does it mean that the code may abort for any user trying to run with 1Devp, or is that just a failure in the QC software?

apcraig:
The QC failure is an abort in the ice model, nothing related to the QC test as far as I can tell.  It took 4 years to hit it.  If we believe the 2d implementation is robust, then this suggests there may still be a bug in the 1d implementation that's possibly triggered only on timescales of months to years.  All our bit-for-bit testing with debug flags has been shorter than a year.  We could run QC with debug flags on with 1d and 2d and see what happens.  My concern is the time to completion in that case with reduced optimization, running 1 degree for 5 years.  But it's probably worth doing.  I propose we add that idea to the issue that we'll create.

eclare108213:
I guess you can't restart the QC from shortly before the crash with debug flags on, and expect it to crash again.  Would that be worth a try?

eclare108213:
That's quite a debugging project.  It needs to be done, but I don't think it needs to be done before this particular PR is merged.

TillRasmussen:
I will try to rerun the qc test and output more diagnostic.

eclare108213:
@dabail10 Is this the type of error you get for CFL violations in CESM? 

> ```
>   (zap_snow_temperature)Tmin:  -100.000000000000
>   (zap_snow_temperature)Tmax:  4.486434370863951E-006
>   (zap_snow_temperature)zqsn:  -182365602.580354
>   (zap_snow_temperature)zap_snow_temperature: temperature out of bounds!
>   (zap_snow_temperature)k:           1
>   (zap_snow_temperature)zTsn:  -103.894744835945
>   (zap_snow_temperature)Tmin:  -100.000000000000
>   (zap_snow_temperature)Tmax:  5.419524611890995E-006
>   (zap_snow_temperature)zqsn:  -182424769.766085
> ```

 I'm wondering if the 1D evp QC test just happens to be hitting one of these, and the 2D case barely misses it.  (The "incremental" part of incremental remap assumes that ice moves no farther than 1 grid cell in 1 time step.)  If restarting the 1D case with a reduced timestep runs through this point, CFL could be the culprit -- there might not be anything wrong with the 1D evp implementation at all, just unlucky.

@apcraig Here's a suggestion for the warning system.  Is is possible to have it call print_state for the failing grid cell, just for the aborts that are from solutions becoming physically unreasonable, like this one?

apcraig:
It could be a CFL issue, but then I think if it were a matter of being "unlucky", then we'd be seeing it more often in the 2d.  I think it's unlikely that the 2d is quite robust for lots of configurations, but that the 1d gets unlucky on the first long run we do.

@eclare108213, I like your idea of trying to leverage print_state more.  I have created an issue, #622.

apcraig:
See #623 for followup discussion.

Issue #544, opened 2020-12-17
Step therm1 using vertex velocity instead of cell velocity


mattdturner:
There was a recent [bugfix PR in MPAS-seaice](https://github.com/MPAS-Dev/MPAS-Model/pull/773) that addressed an issue in step therm1 where the vertex velocities were being used instead of cell center velocities.  Per that PR:

> This bug involves Step therm 1 using the vertex ice velocity rather than the interpolated cell ice velocity. This has been there since the code was written and follows what CICE does:
https://github.com/CICE-Consortium/CICE/blob/master/cicecore/cicedynB/general/ice_step_mod.F90#L327
This is much less of a problem with CICE since the vertex velocity of the same indices is always close (one of the cell vertices) to the cell velocity of the same indices. That is not true for MPAS-Seaice where a vertex velocity of a given index might be on the other side of the partition from the cell velocity of the same index. A consequence of this for CICE is that CICE is not invariant for a transformation of the grid that leaves the physical cell positions unchanged but alters the indices used to label those cells.

eclare108213:
Thanks for pointing that out @mattdturner 
This does not need to be fixed in Icepack itself, only the fields that are passed to it from CICE.

proteanplanet:
@eclare108213 Please can you create a pull request to fix this - I'm up to my arms in E3SM work at present.

eclare108213:
@mattdturner @proteanplanet
Do I understand correctly that the assumption here is that the wind components arrive in the cell centers rather than on the vertices, since the ice velocities are only used to compute the wind stress components?  Do we need to be more careful here that the ice and atmo velocities are consistent, located in the same place in the cell?

Related question for @rallard77 @daveh150:  are the JRA vector quantities colocated with the temperatures etc?

proteanplanet:
> @mattdturner @proteanplanet
> Do I understand correctly that the assumption here is that the wind components arrive in the cell centers rather than on the vertices, since the ice velocities are only used to compute the wind stress components? Do we need to be more careful here that the ice and atmo velocities are consistent, located in the same place in the cell?
> 
> Related question for @rallard77 @daveh150: are the JRA vector quantities colocated with the temperatures etc?

[...]

PR #576, opened 2021-03-14
use cell-centered ice velocity for high-freq wind stress calculation

eclare108213:
Use interpolated, T-grid cell-centered ice velocity for high-frequency wind stress calculation instead of U-grid velocity
- [x] Developer(s): 
E. Hunke, based on modification to MPAS-seaice
     https://github.com/CICE-Consortium/Test-Results/wiki/5a0a559aaf.badger.intel.21-03-13.231354.0
    40-pe runs always fail on badger.
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No but the cell-centered assumption could be noted there
Addresses https://github.com/CICE-Consortium/CICE/issues/544.
This PR should change the answers for -s alt01 tests, because it alters the ice velocity used to calculate the wind stress when highfreq=T.  However the ice dynamics is turned off for alt01 (kdyn=0), so the ice velocity is identically zero and the results are BFB.  We need to fix the alt01 test.


eclare108213:
We need to fix the test combinations so that kdyn is on when highfreq is on.  alt01 has kdyn=0 (off), as does boxdyn.  I think we could just change alt01 to have kdyn=1.

apcraig:
@eclare108213, would you like to update the alt01 test as part of this PR or do you prefer that be a separate PR?

eclare108213:
There's potentially a bigger problem with the alt01 and boxdyn tests which I need to look at more closely (why is kdyn =0 off for a 'dyn' box test?).  If the other PR reviewers approve this PR as it is, then I think we should go ahead and merge it and keep moving forward.  If I manage to look at the tests before then and see how to change them, then I might add that to this PR, but it's not necessary in my opinion.

proteanplanet (approved):
Approved upon visual inspection. One point to note is the problem of diffusive dynamic coupling through scalar grid points.  That cannot be addressed in this PR which is a fix to an existing problem.

mattdturner (approved):
Approved via visual inspection

eclare108213:
I changed the alt01 test to turn dynamics on, in order to test the effects of changes in the highfreq calculation.  As expected, they aren't BFB.  Results of just the revised alt01 comparisons, with and without this code modification are here:
https://github.com/CICE-Consortium/Test-Results/wiki/7ba26347b2.badger.intel.21-03-15.174352.0
I don't think we need to do QC on this, but we can if anyone thinks we should.
(The boxdyn test, which also has kdyn=0, is poorly named, something we can change in a different PR.  I'll note this in an issue somewhere...)

apcraig (approved):

Issue #572, opened 2021-03-09
Halo Update Fill Implemention Bug with Padded Blocks


apcraig:
In ice_boundary.F90 (both mpi and serial), the fill implementation in the halo update is carried out assuming blocks are NOT padded.  This is a bug.  The current implementation looks like

```
    do j = 1,nghost
       array(1:nx_block,           j,:) = fill
       array(1:nx_block,ny_block-j+1,:) = fill
    enddo
    do i = 1,nghost
       array(i,           1:ny_block,:) = fill
       array(nx_block-i+1,1:ny_block,:) = fill
    enddo
```
but it should look more like

```
       do j = 1,nghost
          array(1:nx_block, jlo-j,iblk) = fill
          array(1:nx_block, jhi+j,iblk) = fill
       enddo
       do i = 1,nghost
          array(ilo-i, 1:ny_block,iblk) = fill
          array(ihi+i, 1:ny_block,iblk) = fill
       enddo
```

One particular challenge is that ilo/ihi/jlo/jhi do not seem to be accessible in ice_boundary due to circular logic, so some data may need to be moved around in the code.

The good thing is that this doesn't seem to affect any results.  This was detected by initializing all arrays with signalling nans and then trapping.  Without that, you seem to fortuitously get a 0 multiplied by the uninitialized values in all cases.  The fact that padded and non-padded decomps produce bit-for-bit results suggests this has not affected answers in actual model runs.

In the one place where the signalling nans was trapped (to_ugrid in ice_grid.F90), we have worked around it by initializing tarea to zero everywhere.  When the halo fill is fixed, that fix should be removed, and we should confirm that the original issue is resolved.

See also #560.

phil-blain:
For easy reference, one such loop is at:
https://github.com/CICE-Consortium/CICE/blob/d529df8cf473cec234a4f5c1bb8d4cef3bf1f5a1/cicecore%2FcicedynB%2Finfrastructure%2Fcomm%2Fmpi%2Fice_boundary.F90#L1288-L1295

Issue #247, opened 2018-11-20
extend CICE technical testing


apcraig:
There are a few things I'd like to do,
- qc test different compilers/machines against each other
- add signalling nans to debug compiler options
  - cray, "-ei"
  - intel, "-init=snan,arrays" (may need intel 18+)
- set unused tracer indices to -1 and code around their use if needed

eclare108213:
What do the signalling-nan errors mean?

We need to make a list of other tests that we know are missing, such as revp, maybe not in this issue.

phil-blain:
The flags means that all arrays and scalars are initialized with signaling nans. Signaling nans make the code abort if any computations uses them. It is a way to check that the code does not use uninitialized variables

phil-blain:
Revp is included in the base_suite through test using the options "alt02" and "boxrestore"

phil-blain:
Two thoughts : 
1. I think that we should consider trying to run CICE in Valgrind to check memory usage (using memcheck)

2. On intel, compiling with -check arg_temp_created gives a lot of output. Has this been discussed before ?

apcraig:
I have run a number of qc tests against each other with CICE #5f97e45e2362518d which is CICE6.1.1.  All qc comparisons pass.  These include

  conrad_intel
  conrad_pgi
  conrad_gnu
  conrad_cray
  onyx_intel
  cheyenne_intel

all with 64x1 pe counts.  I also tried a 16x4 case, but it was running too slow, so I abandoned that test.  I think for the time being, we can consider this task as being checked and OK.

PR #579, opened 2021-03-16
Fixes for uninitialized variables (snan testing)

apcraig:
## PR checklist
    This fixes a handful of issues trapped by snan debug testing.  See #572, #560
- [X] Developer(s): 
    phil-blain, apcraig
    bit-for-bit full test suite on cheyenne 3 compilers, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#eec256efb85e5206cd6e7745188e2c44a093c028
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

This replaces #560.

See also #572, #247 

This fixes several issues found by turning on snan debug tests with the intel/gnu compilers.  There are still outstanding issues, so snan is not out by default.  This fixes

- adds several tests including decomp tests with debug on, padded decomp with 1 active gridcell width in i and j directions, a 1x1 decomp test, and some serial decomp tests.
- fixes fill of halo in haloupdate.  was not working corrected for padded blocks.  required storing some additional information in the halo datatype at initialization.
- fixes issues with uvm and tarea, fixed by haloupdate fill.
- pulls in fixes from #560 (sst initiallization, Lsub initialization)
- fixes a bug in the size 1 padded grid cell decomp (ice_blocks)
- fixes a spacecurve issue
- adds snan debug options for several machines, but many still commented out due to outstanding issues.

### DIFF
--- a/cicecore/cicedynB/infrastructure/ice_blocks.F90
+++ b/cicecore/cicedynB/infrastructure/ice_blocks.F90
@@ -252,8 +252,8 @@ contains
             !*** set last physical point if padded domain
 
             else if (j_global(j,n) == ny_global .and. &
-                     j > all_blocks(n)%jlo      .and. &
-                     j < all_blocks(n)%jhi) then
+                     j >= all_blocks(n)%jlo      .and. &
+                     j <  all_blocks(n)%jhi) then
                all_blocks(n)%jhi = j   ! last physical point in padded domain
             endif
          end do
@@ -300,8 +300,8 @@ contains
             !*** last physical point in padded domain
 
             else if (i_global(i,n) == nx_global .and. &
-                     i > all_blocks(n)%ilo      .and. &
-                     i < all_blocks(n)%ihi) then
+                     i >= all_blocks(n)%ilo      .and. &
+                     i <  all_blocks(n)%ihi) then
                all_blocks(n)%ihi = i
             endif
          end do
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -384,7 +384,7 @@
       ! T-grid cell and U-grid cell quantities
       !-----------------------------------------------------------------
 
-!     tarea(:,:,:) = c0
+      tarea(:,:,:) = c0
 
       !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block)
       do iblk = 1, nblocks
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -1636,7 +1636,7 @@
       !-----------------------------------------------------------------
 
       bm = c0
-!     uvm = c0
+      uvm = c0
 
       !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block)
       do iblk = 1, nblocks
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -2401,7 +2401,8 @@
          do iblk = 1, nblocks
             do j = 1, ny_block
             do i = 1, nx_block
-               k = min(nint(kmt(i,j,iblk)),nlevel)
+               k = nint(kmt(i,j,iblk))
+               if (k > nlevel) call abort_ice(subname//' kmt gt nlevel error')
                if (k > puny) bathymetry(i,j,iblk) = depth(k)
             enddo
             enddo
@@ -2492,8 +2493,8 @@
       do iblk = 1, nblocks
          do j = 1, ny_block
          do i = 1, nx_block
-            k = kmt(i,j,iblk)
-            if (k > nlevel) call abort_ice(subname//' kmt/nlevel error')
+            k = nint(kmt(i,j,iblk))
+            if (k > nlevel) call abort_ice(subname//' kmt gt nlevel error')
             if (k > 0) bathymetry(i,j,iblk) = depth(k)
          enddo
          enddo
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -384,9 +384,6 @@
       ! T-grid cell and U-grid cell quantities
       !-----------------------------------------------------------------
 
-! tcraig, tcx, this is temporary, see https://github.com/CICE-Consortium/CICE/issues/572
-      tarea(:,:,:) = c0
-
       !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block)
       do iblk = 1, nblocks
          this_block = get_block(blocks_ice(iblk),iblk)
@@ -1637,8 +1634,6 @@
       !-----------------------------------------------------------------
 
       bm = c0
-! tcraig, tcx, this is temporary, see https://github.com/CICE-Consortium/CICE/issues/572
-      uvm = c0
 
       !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block)
       do iblk = 1, nblocks
@@ -1673,11 +1668,11 @@
          jlo = this_block%jlo
          jhi = this_block%jhi
 
-         ! needs to cover halo OR need a halo update
-         do j = 1, ny_block
-         do i = 1, nx_block
-            tmask(i,j,iblk) = .false.
-            umask(i,j,iblk) = .false.
+         ! needs to cover halo (no halo update for logicals)
+         tmask(:,:,iblk) = .false.
+         umask(:,:,iblk) = .false.
+         do j = jlo-nghost, jhi+nghost
+         do i = ilo-nghost, ihi+nghost
             if ( hm(i,j,iblk) > p5) tmask(i,j,iblk) = .true.
             if (uvm(i,j,iblk) > p5) umask(i,j,iblk) = .true.
          enddo
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -1196,15 +1196,9 @@
       if (icepack_warnings_aborted()) call abort_ice(error_message=subname, &
          file=__FILE__, line=__LINE__)
 
-      !$OMP PARALLEL DO PRIVATE(iblk,i,j)
-      do iblk = 1, nblocks
-         do j = 1, ny_block
-         do i = 1, nx_block
-            ANGLE(i,j,iblk) = c0              ! "square with the world"
-         enddo
-         enddo
-      enddo
-      !$OMP END PARALLEL DO
+      hm (:,:,:) = c0
+      kmt(:,:,:) = c0
+      angle(:,:,:) = c0   ! "square with the world"
 
       allocate(work_g1(nx_global,ny_global))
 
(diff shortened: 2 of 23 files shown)


cicecore/cicedynB/infrastructure/ice_blocks.F90
```diff
@@ -252,8 +252,8 @@ subroutine create_blocks(nx_global, ny_global, ew_boundary_type, &
             !*** set last physical point if padded domain
 
             else if (j_global(j,n) == ny_global .and. &
-                     j > all_blocks(n)%jlo      .and. &
-                     j < all_blocks(n)%jhi) then
+                     j >= all_blocks(n)%jlo      .and. &
```
eclare108213, line 255: If I understand correctly, this change allows jlo=jhi, i.e. 1x1 domains, which weren't allowed before?

cicecore/cicedynB/infrastructure/ice_grid.F90
```diff
@@ -1684,11 +1682,14 @@ subroutine makemask
          tarean(:,:,iblk) = c0
          tareas(:,:,iblk) = c0
 
-         do j = 1, ny_block
-         do i = 1, nx_block
+         do j = jlo,jhi
```
eclare108213, line 1685: Are the ghost cells for these masks filled later?

cicecore/cicedynB/infrastructure/ice_blocks.F90
apcraig: Correct.  This is where the 1x1 blocks failed.  The 1x1 and 2x2 are still producing different results from the other block sizes and decomps, but they are running and part of our test suite now.

cicecore/cicedynB/infrastructure/ice_grid.F90
apcraig: They are not, but the halos are never accessed, so there is no error when we initialize with snan.  There are probably many arrays that have uninitialized memory, but that's fine.  In many ways, that's what we want.  We want to only initialize memory that we think is used.  That way we can trap code (or learn more about our codes) when an snan actually triggers an error.  If we initialized everything explicitly to zero just for the sake of doing so, the snan testing wouldn't tell us anything.

cicecore/cicedynB/infrastructure/ice_blocks.F90
apcraig: Just to elaborate a bit further.  Before this fix, jhi was not set properly if the active block size was 1.  It remained the default (ny_block - nghost), see line 194.    This section of code is where the lo/hi indices are set for padded blocks.  It worked fine as long as the block size was at least 2.

eclare108213 (approved):

cicecore/cicedynB/infrastructure/ice_grid.F90
```diff
@@ -1198,15 +1196,9 @@ subroutine rectgrid
...
-      !$OMP PARALLEL DO PRIVATE(iblk,i,j)
-      do iblk = 1, nblocks
-         do j = 1, ny_block
-         do i = 1, nx_block
-            ANGLE(i,j,iblk) = c0              ! "square with the world"
-         enddo
-         enddo
-      enddo
-      !$OMP END PARALLEL DO
+      hm (:,:,:) = c0
```
TillRasmussen, line 1199: Is it necessary to initialize to zero in all of the block (also valid for kmt and angle?

apcraig: There are a few different grid initialiizations (popgrid, popgrid_nc, latlon, rectangular, etc), and they don't all initialize the variables in the same way.  I believe zero-ing these out here is required for the rectangular grid option that is used for the box setups.  It is not needed for the popgrid option that we use for our more standard gx3+ testing.  It could be that a better solution is to make all the grid options behave in the same way, but it wasn't clear to me what those requirements should be, so I went this direction.

cicecore/cicedynB/infrastructure/ice_grid.F90
```diff
@@ -2407,7 +2401,7 @@ subroutine get_bathymetry
          do iblk = 1, nblocks
             do j = 1, ny_block
             do i = 1, nx_block
-               k = kmt(i,j,iblk)
+               k = min(nint(kmt(i,j,iblk)),nlevel)
```
phil-blain, line 2404: Is this also something that was found with signalling NaNs ? not sure what this is changing ...

phil-blain (commented):
OK so I took a look at each commit. If I understand correctly, the issues with `tarea` and `uvm`  highlighted in #560 are fixed, and so #572 would be closed as well. Is that right ? 

However, I'm not sure I understand completely how the fix to the `fill` values in the halo updates routine fix the problem  for uvm, since `fillValue` is not used for uvm... unless there is something I'm missing....

--- 

On the Git side of things, I would prefer that we not keep my temporary fixes of initializing tarea and uvm to zero, only to remove them  in latter commits. Even if we use a squash merge workflow, the commit messages of all commits on the branch are still kept (if not manuallt edited-out) in the merge commit message, so it can be very confusing to read when you are actually reading the commit log to try to understand the evolution of the codebase...

apcraig:
@phil-blain, thanks for having a look.  I was hoping you'd find a few minutes.  uvm=0 and tarea=0 are no longer being set.  I think the halo fill fixed this.  The halo fill operates even if no fill value is passed.  It's set internally to zero by default, so it does matter in those cases.  So yes, this addresses #560 and #572.

The ice_spacecurve mod does fix the spacecurve error.  The spacecurve implementation is not particularly robust in general, so my fix isn't wrong, but may not be how a robust implementation might deal with the issue.

The k = "...nint(kmt)...", is needed in cases where kmt can be greater than nlevel.  In that case, you'll end up referencing a element in the bathy arrays that is out of bounds.   This whole implementation is a little dicey as nlevel is hardwired to 40 with fixed depths and that may have no relationship to the kmt file used by the CICE/ocean model or the depths defined there.  But the idea is that you would normally read the depth file separately if you really had that information and were coupling to an ocean model.  That's what we do in RASM when we use the bathymetry/landfast ice.  For landfast ice, it also can't really matter what the depth is once it's greater than a few 10s of meters?  Anyway, a better fix would be to get rid of this hardwired 40 levels and require that a depth profile be input when landfast ice is on.  Also, kmt in CICE is a real, so we need to nint it to use the min function with nlevel.

Some of the errors that this PR fixes appeared when I turned on the signalling nans and then ran the full test suite.  There are still several outstanding issues that I couldn't fix quickly, so those are deferred and the snan is NOT on by default on the machines I test on.  There is still work to do in this regard.  #247 is still open.

Finally, I understand the concerns about the git commit history, but I don't think it's worth making an effort to create a new branch or to try to clean up the history.  I can edit some of it when I squash merge.  In some ways, the squash merge does what we want and is why I advocate for it's use, it largely eliminates all that potentially chaotic history within the master branch commit series.  Although I understand it does appear within an individual commit unless it's reviewed and cleaned up when we squash.  

Let me know if there are any other questions.  thanks!

TillRasmussen:
I think that you could argue that use_bathymetry should be true as default. This reads in a depth from a bathymetry file. use_bathymetry=false is for a specific ocean setup used by Environment Canada (correct  me if I am wrong @phil-blain) .

eclare108213:
I agree with @TillRasmussen that an abort is appropriate if kmt > bathymetry file levels, when bathymetry is being read in.  My general philosophy is that all namelist parameters relative to a given 'master namelist flag' should be set by default to the values desired when the master flag is turned on, but it sounds like the bathymetry file should be off by default, except for our tests, since coupled models read it in outside of the sea ice component.  Is that correct?

apcraig:
I have updated the bathymetry implementation to abort if kmt > nlevel.  That seems like a good implementation.

phil-blain:
> The k = "...nint(kmt)...", is needed in cases where kmt can be greater than nlevel. In that case, you'll end up referencing a element in the bathy arrays that is out of bounds. This whole implementation is a little dicey as nlevel is hardwired to 40 with fixed depths and that may have no relationship to the kmt file used by the CICE/ocean model or the depths defined there. But the idea is that you would normally read the depth file separately if you really had that information and were coupling to an ocean model. That's what we do in RASM when we use the bathymetry/landfast ice. For landfast ice, it also can't really matter what the depth is once it's greater than a few 10s of meters? Anyway, a better fix would be to get rid of this hardwired 40 levels and require that a depth profile be input when landfast ice is on. Also, kmt in CICE is a real, so we need to nint it to use the min function with nlevel.


OK, I have read that part of the code more carefully and understand more what it's doing and what you are changing. If I understand correctly, the conversion to integer was already implicit since `k` is declared as int, but it's more readable if it's explicit.  I also agree with Till and Elizabeth that it makes more sense to just abort if `kmt > nlevel`. While we're there, I think we could change `k > puny` to `k >  0` since  `k` is an integer but `puny` is a double. It would be less confusing, I think (and it's already written that way in `get_bathymetry_popfile`). As an aside, the whole last part of these two subroutines is in fact duplicated, so a further clean-up could be to refactor this; but that can come latter.

---

> I think that you could argue that use_bathymetry should be true as default. This reads in a depth from a bathymetry file. use_bathymetry=false is for a specific ocean setup used by Environment Canada (correct me if I am wrong @phil-blain) . 

> [...] My general philosophy is that all namelist parameters relative to a given 'master namelist flag' should be set by default to the values desired when the master flag is turned on, but it sounds like the bathymetry file should be off by default, except for our tests, since coupled models read it in outside of the sea ice component. Is that correct?

To clarify, in our in-house CICE4 implementation we pass the dynamic water height (bathy + SSH) fro NEMO to CICE, so no we do not read the bathymetry field from CICE. I've not yet gotten around to revisit that in my work on coupling our NEMO with CICE 6, but the same approach will be used. I was planning to do that soon.

With regards to what should be the default for `use_bathymetry`  in our namelist, I think both approaches are valid. I'd say it depends if we want the namelist to have good defaults for coupled setups or standalone setups: for coupled, `use_bathymetry = false ` (as it is now) is OK, because usually the ocean model is in charge of reading the bathymetry, but for standalone runs  `use_bathymetry = true` would make more sense, since we do provide bathymetry files for our 3 grids, and that should give a more detailed field than the harcoded levels...

apcraig:
Thanks @phil-blain, there are a few things about the bathymetry implementation that I haven't understood well.  I really don't understand why we have a 40 level default at all.  I guess it's for testing.  The get_bathymetry_popfile was added by RASM because pop has a well defined bathymetry file different from whatever was implemented before.  I thought it was easier to create a new "read" method rather than duplicate the file in another format.  Hence, it's largely a duplicate of the default implementation.

k = kmt() is not the same as k = nint(kmt()).  The former is doing an "int(kmt())".  We really do want an explicit nint in this case, although in practice, they seem to be producing the same result for our test cases (which may not have any landfast ice unfortunately).

I will modify the k>puny vs k>0.  I agree that's a more robust implementation.  kmt is normally thought of as an integer but it's a real in CICE.  I think that's where the k/kmt type problems are showing up in the bathymetry, confusion about the types.  I suspect we have (kmt()>puny) elsewhere, but should k=nint(kmt()), k>0 when converted to integer as opposed to k=kmt(), k>puny which is what seems to have been done.

I also agree we may want to rethink use of the 40 level hardcoded default, but as a separate PR.

apcraig:
I have rerun the full test suite on cheyenne after rebasing the code to the current master.  Results are as expected and here, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#52e8102274c8a948f43e3e4e6b0428250706df58

apcraig:
I updated the k>puny line and ran a quick suite to make sure it didn't create any problems.  all seems fine with that change.

phil-blain (approved):

PR #602, opened 2021-05-24
Update testing, in particular for diagnostics and decompositions

apcraig:

## PR checklist
    Update testing, in particular for diagnostics and decompositions
- [X] Developer(s): 
    apcraig
    Full test suite run on cheyenne with 3 compilers.  Tests are bit-for-bit except boxslotcyl which had a change to the namelist settings (tr_lvl=tr_pond_lvl=.false.) and the calchk unit test which was set to 1000 years of testing by default, but should have been 100,000 years.  Several new tests were added.
    - [X] bit for bit (except as noted above)
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

List of changes are below.  Of particular note
 - a couple bugs were fixed for sections of the code that we were generally not using/testing but are now
     - error in size of allocation in 1d rake
     - out of bounds issue for fcontopn_d and fsurfn_d when ncat > 6
 - the dbug namelist was renamed to forcing_diag which could impact users
 - two new namelist were added, debug_model (logical) and debug_model_step (integer) and these trigger the debug_ice diagnostics after the timestep associated with debug_model_step.  
 - there is a new file for testing the wghtfile decomposition option, CICE_data/grid/gx1/cice62_gx1_wghtmask.nc.  This file is on cheyenne and I will move it to a few other testing platforms, but maybe it should be added to the gx1 input files.  This file represents the number of months that ice existed at any grid point during the 2005-2009 gx1 production test run and is used to allocate blocks to procs at gx1 with the wghtfile test decomposition.  The maximum value is 120 and these values are read and normalized during the decomposition step in CICE.

These changes address many issues in #437 as well as lines 7-17 and maybe 81-82 in https://docs.google.com/spreadsheets/d/12tIfm5OzvzH_LF_ie9WbOn_w5aY1hv4SrqTQQ8bqM-s.  Code coverage should go up after this update and that will be evaluated separately.

Otherwise, a summary of all changes follows
    
- Add alt06 test, ncat=7, kcatbound=3, nslyr=3
- Add bigdiag test with lots of diagnostic output, debug_model=true, debug_blocks=true, etc
- Add spiralcenter decomp test
- Add gx1 weightfile decomp test
- Add gbox180 with spacecurve decomp test to trigger Peano and Cinco decompositions
- Update sectrobin decomp test to check land block elimination
- Rename boxdyn to boxnodyn
- Update calchk to check 100,000 years instead of 1000 years (which was set for quick debugging, but was not supposed to be checked in).
- Modify boxslotcyl, turn off tr_lvl and tr_pond_lvl, changes answers
- Clarify boxadv, boxnodyn, boxrestore, explicitly set shortwave=ccsm3, albedo_type=constant
- Cleanup spacecurve implementation including debug flagging, recode print to write, module private, add printcurve for gridsize of 30, remove qsort and partition subroutines.
- Deprecate IsLoadBalanced in ice_spacecurve
- Remove CICE_RunMod.F90_debug and migrate high level debug checks to CICE_RunMod.F90 with new debug_model namelist.  Also add debug_model_step to specify at what timestep to start writing debug_model output.  debug_model writes data for the first point specified by lonpnt, latpnt.
- Move debug_ice routine into ice_diagnostics.F90 and out of CICE.F90
- Remove writeout_finished_file from CICE_Finalize in several drivers
- Deprecate print_points_state via an ifdef
- Rename l_stop in ice_transport_driver.F90 to ckflag.  This looked like l_stop logic because of the naming but it isn't.
- Fix issue of size of fcontopn_d and fsurfn_d when ncat greater than 6 in init_coupler_flux.  Would go out of bounds, now initializes fcontopn and fsurfn to 6th value for any category greater than 6.
- Rename dbug variable name in ice_forcing.F90 to forcing_diag and change namelist variable name.
- Update ice_distributionGet output and implementation to support more general arguments
- Add ice_distributionGet test in init_domain_distribution when debug_blocks is true.
- Fix bug in create_distrb_rake for 1d rake of allocation size.  Add a test that triggers the 1d rake.
- Update debug flag in ice_distribution.F90 leverage debug_blocks.
- Update complog part of baseline script to report missing data as MISS instead of FAIL
- Update documentation.


configuration/scripts/ice_in
```diff
@@ -29,14 +29,16 @@
     diagfreq       = 24
     diag_type      = 'stdout'
     diag_file      = 'ice_diag.d'
+    debug_model    = .false.
+    debug_model_step = 999999999
```
eclare108213, line 33: Would it be better to set debug_model_step to 0 so that it always starts printing immediately, as the default?  Assuming it's ignored unless debug_model=T.

configuration/scripts/options/set_nml.alt03
```diff
@@ -23,3 +23,4 @@ Ktens          = 0.
 e_ratio        = 2.
 seabed_stress    = .true.
 use_bathymetry = .true.
+l_mpond_fresh = .true.
```
eclare108213, line 26: If particular changes are addressing lines in the spreadsheet, could you add a note on the line that says which test script was changed, and how?

configuration/scripts/tests/baseline.script
```diff
@@ -69,21 +69,26 @@ if (${ICE_BASECOM} != ${ICE_SPVAL}) then
...
-      set logstatus = $status
-
-      if ( ${logstatus} == 0 ) then
-        echo "PASS ${ICE_TESTNAME} complog ${ICE_BASECOM}" >> ${ICE_CASEDIR}/test_output
-        echo "Regression baseline and test dataset may be the same"
-      else if ( ${logstatus} == 1 ) then
-        echo "FAIL ${ICE_TESTNAME} complog ${ICE_BASECOM} different-data" >> ${ICE_CASEDIR}/test_output
-        echo "Regression baseline and test dataset are not the same"
-      else if ( ${logstatus} == 2 ) then
+      if ("${base_file}" == "" || "${test_file}" == "" ) then
```
eclare108213, line 72: are you getting rid of comparelog.csh?

doc/source/user_guide/ug_implementation.rst
```diff
@@ -270,8 +269,11 @@ routines, is adopted from POP. The boundary routines perform boundary
 communications among processors when MPI is in use and among blocks
 whenever there is more than one block per processor.
 
-Open/cyclic boundary conditions are the default in CICE; the physical
-domain can still be closed using the land mask. In our bipolar,
+Open/cyclic boundary conditions are the default in CICE.  Closed boundary
```
eclare108213, line 272: thank you for clarifying!

doc/source/user_guide/ug_troubleshooting.rst
```diff
@@ -119,17 +119,24 @@ Several utilities are available that can be helpful when debugging the
 code. Not all of these will work everywhere in the code, due to possible
 conflicts in module dependencies.
 
-*debug\_ice* (**CICE.F90**)
+*debug\_ice* (**ice\_diagnostics.F90**)
```
eclare108213, line 122: I think this was originally in CICE.F90 due to circular dependencies.  If that's not longer a problem, great!

doc/source/user_guide/ug_troubleshooting.rst
```diff
@@ -138,11 +145,11 @@ conflicts in module dependencies.
 
 `print\_points` (**ice\_in**)
     If true, print numerous diagnostic quantities for two grid cells,
-    one near the north pole and one in the Weddell Sea. This utility
+    defined by `lonpnt` and `latpnt` in the namelist file.
```
eclare108213, line 148: I understand that this will be generalized so that the user alternatively can insert the indices directly, perhaps in a later PR.

eclare108213 (approved):
Looks great, thanks @apcraig.  I have some questions but nothing that keeps me from approving this as it is.

Does anyone (or any modeling group) use the space-filling curves distribution?

In the diff of ug_case_setting.rst, it looks like the first many (200?) lines of the file are italized but later ones are not.  Is that a problem in the file itself, or just with this diff?

configuration/scripts/ice_in
apcraig: It was 99999999 by default before, so I kept it that way.  I'm happy to make the default 0, let me know if you prefer that.

configuration/scripts/options/set_nml.alt03
apcraig: This recommendation came from #437.  https://github.com/CICE-Consortium/CICE/issues/437#issuecomment-734004147.  But it doesn't change answers which is a little concerning and something that we need to look into more probably.

configuration/scripts/tests/baseline.script
apcraig: It's still there, just below this line in the else and indented.  This if just provides a better error message if the either base or test file are not defined.  Up to now, it would produce a FAIL, now it produces a MISS which is more accurate.

doc/source/user_guide/ug_troubleshooting.rst
apcraig: It didn't seem to be a problem to move.

configuration/scripts/ice_in
eclare108213: I think 0 would be better, since you have a flag to turn it off.

configuration/scripts/options/set_nml.alt03
eclare108213: Here's what to look for:  l_mpond_fresh might only be used for coupling, and the mixed layer model only calculates heat fluxes, not water/salt, so  l_mpond_fresh might not be testable in our standalone configurations.

configuration/scripts/tests/baseline.script
eclare108213: ok, I missed that part

configuration/scripts/ice_in
apcraig: Sounds good, I'll make this change in the next test coverage PR.

eclare108213: why not this PR?

configuration/scripts/options/set_nml.alt03
apcraig: So far as I can tell, l_mpond_fresh shows up in CICE_RunMod.F90 in the standalone model as

```
            if (l_mpond_fresh) then
               fpond(i,j,iblk) = fpond(i,j,iblk) * rhofresh/dt
               fresh(i,j,iblk) = fresh(i,j,iblk) - fpond(i,j,iblk)
            endif
```

I'm not sure what happens to fpond or fresh after that, but it doesn't seem to impact the solution.  I guess I think that's OK.  When we turn on options, it would be nice to see an impact, but if that's not possible, I still think it worthwhile to have a test that turns the flag on and at least tests the code in a technical sense.

configuration/scripts/ice_in
apcraig: I can do it in this PR too.  I sort of trying to avoid doing some extra testing.  Given this change is minor, nobody is going to use this feature at the moment, and this will be updated on the trunk in the next week (hopefully), I thought I'd just wait.  I have three options
- make the change on this PR and rerun all or a subset of tests
- make the change on the PR and not test
- make the change in the next PR and carry out a full test suite (which I'd do anyway)

I was going for the 3rd option.  I can do the 1st.  I try to avoid doing the 2nd when changes occur in the source code or scripts as it's bad practice and risky.

eclare108213: That's fine, whatever works best for you.

dabail10 (approved):
These are pretty extensive changes, but look to be completely diagnostic. I think these are nice features and am fine with the naming and structure.

phil-blain:
Hi @apcraig,

> - there is a new file for testing the wghtfile decomposition option, CICE_data/grid/gx1/cice62_gx1_wghtmask.nc. This file is on cheyenne and I will move it to a few other testing platforms, but maybe it should be added to the gx1 input files. This file represents the number of months that ice existed at any grid point during the 2005-2009 gx1 production test run and is used to allocate blocks to procs at gx1 with the wghtfile test decomposition. The maximum value is 120 and these values are read and normalized during the decomposition step in CICE.

This should indeed be added to the gx1 input files, since it is used in the `decomp_suite` . I've just ran that suite and the  `daley_intel_restart_gx1_64x1x16x16x10_dwghtfile` test failed because it's missing that file.

Issue #567, opened 2021-02-27
Outstanding Time Manager Issues


apcraig:
This is an attempt at aggregating outstanding time manager issues.  Several have been fixed as a result of the new JRA55 implementation and the new time manager. But there are still a few outstanding questions/issues.

- [ ] We do not read the time axis in the forcing datasets.  The file record is based on the yday value and it assumes the input files are yearly and have a particular input frequency.  There are few checks that the forcing date and model date actually match.  This could be improved (a bunch).

- [ ] Still open question about whether we should advance at the beginning of the timestep in the standalone model.  The coupled models advance at the beginning, but the standalone advances at the end of the timestep (including at the end of the initialization).
- [ ] Some question about the usage of use_restart_time, ice_ic, and runtype interaction.  For now, we're leaving as is.  Partly this is #527.
- [ ] Documentation/naming of dt and dt_dyn, #162.

Feel free to edit (add/remove items) this list here while also commenting below.

phil-blain:
We could add:
- [ ] Remove `istep0` from the namelist (and maybe from the code), https://github.com/CICE-Consortium/CICE/pull/566#discussion_r605058944

phil-blain:
> Still open question about whether we should advance at the beginning of the timestep in the standalone model. The coupled models advance at the beginning, but the standalone advances at the end of the timestep (including at the end of the initialization).

Tangentially related: this leads to the initial condition being written *after* the first time step...

Another minor thing I noticed: when using `hist_avg = .true.`, the initial condition is also written as being "averaged" from "time = 0" to "time = after the first time step", which is kind of weird...

apcraig:
A couple comments.  I don't think we should remove istep0.  I see a use case where someone runs 100 years and then wants to start a new run from the restart at year 50 and run 50 years with a parameter change.  It could be helpful to set istep0 to a value that allows direct comparison of the step numbers of the two runs.

The initial condition is written before the model state is advanced in standalone mode.  The problem is that the timestep is advanced as the model is initialized.  So the real problem is that the initial condition filename/date is incorrect even though the data is the initial data.  I don't think there is a clean way around this.  We do not want to subtract a timestep from the date/filename internally as that could create problems for models that don't advance their timestep at initialization.  Ultimately, this comes back to the sequencing of time advancement in the model relative to other operations.

I am doing some testing/updates and I have confirmed that history files are written at step=1 with the updated time manager, so have checked that off the list above.

PR #610, opened 2021-06-19
Fix history and restart frequency, new features to scripts

### LINKED ISSUE #527: use_restart_time and runtype = 'continue'
This is not working as it should in the CESM driver for CICE6. I'm not sure where there is a conditional to make sure that the restart time from the file should be used on a continue run?

### LINKED ISSUE #541: Remove/fix close_boundaries and update box2001 and boxslotcyc tests
The close_boundaries option does not work.  This option needs to be fixed or removed.  Currently, box2001 and boxslotcyc use close_boundaries and turning that option off changes answers, so it will continue to live in the code until some decisions are taken.  See also #540.

### LINKED ISSUE #569: Running a test suite with additional options (-s / --sets) breaks when tests are BFB-compared with other tests
I just ran into this issue again and realized I already knew about that but had not opened an issue to document it.

In the context of #560 I ran the `decomp_suite` in a debug configuration, i.e. I added `-s debug` to the cice.setup command line.

However, it seems our testing framework is unprepared for that use case if any test in the test suite have an entry under the `BFB-compare` column in the `<test_suite>.ts` definition file. For example, for the decomp suite: 

https://github.com/CICE-Consortium/CICE/blob/0ac4cd9cc2934742d04187527d9ba258d3360cfe/configuration/scripts/tests/decomp_suite.ts#L1-L6

The name of the test in the BFB-compare column is written to the `ICE_BFBCOMP` variable in cice.settings verbatim, i.e. it does not take into account additional options that were added on the command line. So the BFB compare part of the test fails with "missing-data" because they are not looking at the right place, i.e. the test on line 5 will look for `restart_gx3_4x2x25x29x4_dslenderX2` but should look for `restart_gx3_4x2x25x29x4_debug_dslenderX2`.

So cice.setup should ideally add additional options from the command line before writing  `ICE_BFBCOMP` in cice.settings for each test.

### LINKED ISSUE #589: Hourly output broken by recent time manager update
I've just noticed that hourly outputs does not work anymore since the update to the time manager. 

I'm comparing b720380 (Update Time Manager (#566), 2021-03-16) and 5a0a559 (Modified k1 value and some of the doc (#574), 2021-03-11) (the commit just before).

With these modification to the namelist:
~~~diff
diff --git 1/../before-time-manager-update2/ice_in 2/./ice_in
index 1dfb0b6..684c8ed 100644
--- 1/../before-time-manager-update2/ice_in
+++ 2/./ice_in
@@ -34,7 +33,7 @@ &setup_nml
     latpnt(2)      = -65.
     lonpnt(2)      = -45.
     dbug           = .false.
-    histfreq       = 'm','x','x','x','x'
+    histfreq       = 'h','x','x','x','x'
     histfreq_n     =  1 , 1 , 1 , 1 , 1
     hist_avg       = .true.
     history_dir    = './history/'
@@ -413,9 +412,9 @@ &icefields_nml
     f_VGRDb        = .false.
     f_VGRDa        = .true.
     f_bounds       = .false.
-    f_aice         = 'm' 
+    f_aice         = 'h' 
     f_hi           = 'm'
     f_hs           = 'm' 
~~~

With 5a0a559:
~~~
$ ll -rt $(cice_rundir)/history
total 12M
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_ic.2005-01-01-03600.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-32400.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-28800.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-25200.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-21600.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-18000.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-14400.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-10800.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-07200.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-75600.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-72000.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-68400.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-64800.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-61200.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-57600.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-54000.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-50400.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-46800.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-43200.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-39600.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-36000.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-02-00000.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-82800.nc
-rw-r--r-- 1 me mygroup 503K Apr  1 12:40 iceh_01h.2005-01-01-79200.nc
~~~

with b720380:
 ~~~
$ ll -rt $(cice_rundir)/history
total 504K
-rw-r--r-- 1 me mygroup 503K Apr  1 12:41 iceh_ic.2005-01-01-03600.nc
~~~

/cc @apcraig

### LINKED ISSUE #601: 'decomp' test is not re-runnable
I've noticed the 'decomp' test in the base suite does not work well if the suite is submitted more than once, i.e. 
~~~bash
./cice.setup --suite base_suite ...
# wait for all tests to finish
cd $suite_dir
./suite.submit
~~~

I haven't investigated further but I suspect it's something to do with the shell globbing in the test script. When the test is re-run I get a weird arboresence in the run directory where the `restart` folder is re-created inside each already existing `restart.$decomp` folder...

It's not too important but I thought I'd mention it anyway...

@apcraig

### LINKED ISSUE #608: Smoke test should abort early if the run fails
The `smoke` test does not exit with failure when the CICE run fails, it simply write the `FAIL` result to the `test_output` file:
https://github.com/CICE-Consortium/CICE/blob/1c6c225ec6bd0635025d9aef969a1c25f5c74204/configuration/scripts/tests/test_smoke.script#L6-L7
[...]
https://github.com/CICE-Consortium/CICE/blob/1c6c225ec6bd0635025d9aef969a1c25f5c74204/configuration/scripts/tests/test_smoke.script#L23-L29

In contrast, the `restart` test does `exit 99`:
https://github.com/CICE-Consortium/CICE/blob/1c6c225ec6bd0635025d9aef969a1c25f5c74204/configuration/scripts/tests/test_restart.script#L10-L22

This makes more sense in my opinion. If the initial run fails it makes no sense to go on and try to do the baseline generation, baseline comparing and BFB compare step, no ?

This can lead to misleading "missing data" results when in fact it is the data for the current test that is missing (because the run failed) and not the data for the run which we are comparing against.

apcraig:
## PR checklist
    Fix history and restart frequency, update scripts, update unit tests
- [X] Developer(s): 
    apcraig
    All tests are bit-for-bit on cheyenne with full test suite three compilers, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#0cd4402af2528228ea3aa33dc34e426e5a353474
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

    Fix history/restart frequency bugs
    
    - Fix bugs in history/restart frequency associated with new calendar, closes #589
    - Add histfreq_base and dumpfreq_base namelist ('init' or 'zero') to specify reference date for history and restart output.  defaults are 'zero' and 'init' respectively for hist and dump.  Setting histfreq_base to 'zero' allows for consistent output across multiple runs.  Setting dumpfreq_base to 'init' allows the standard testing which requires restarts be written, for example, 5 days after the start of the run 
    - Create compute_relative_elapsed method in ice_calendar to improve code reuse
    - Update set_nml.histall to include hourly output

    Update initial/restart logic

    - "restart" namelist is deprecated, now computed internally
    - Modify initial/continue init checks and set restart and use_restart_time as needed, closes #527

   New features in scripts

    - Update test scripts to cleanly abort if run fails where possible, closes #608
    - Update decomp test so it's rerunable, remove restart at start of run, closes #601
    - Add ability to do bfbcomp tests with additional options set on command line, closes #569

    Update documentation

    - Update documentation of calendar frequency computation, calendar types, and closed boundaries, closes #541
    - Update documentation with regard to initial/continue modes

    Update unit tests

    - Add optional doabort flag to abort_ice to control whether the method aborts.  This is useful for testing and code coverage statistics, although doabort=.false. will not call the actual abort method, but we can test the interfaces and rest of the code.   
    - Clean up some of the unit tests to better support regression testing.  Remove extra abort calls in bcstchk and sumchk on runs that complete fine but don't pass checks.  These aborts should never have been there.


DeniseWorthen:
@apcraig I've tested this in ufs-weather;  I can now obtain 6hrly history files as before. Everything is b4b with our current baselines.

apcraig:
Thanks @DeniseWorthen!

eclare108213 (commented):
This generally looks okay to me.   This could be done in a later PR, but I think the documentation needs a bit more information added about how the various calendar and restart options interact with each other.  E.g. the table showing ice_ic, runtype and restart values could be expanded to indicate how the *_init dates play in, depending on use_restart_time (or show some examples).  Are use_restart_time and dumpfreq_base completely independent?

cicecore/shared/ice_calendar.F90
```diff
@@ -126,10 +126,9 @@ module ice_calendar
          dumpfreq               ! restart frequency, 'y','m','d'
 
       character (len=char_len), public :: &
-         dumpfreq_base,    & ! restart frequency basetime ('zero', 'init')
-         histfreq_base,    & ! history frequency basetime ('zero', 'init')
-         calendar_type       ! differentiates Gregorian from other calendars
-                             ! default = ' '
+         dumpfreq_base = 'zero', & ! restart frequency basetime ('zero', 'init')
+         histfreq_base = 'init', & ! history frequency basetime ('zero', 'init')
+         calendar_type             ! define calendar type
```
phil-blain: why are the default values set here in **`ice_calendar`** instead of in **`ice_init`** as for other namelist flags ?

cice.setup
```diff
@@ -786,6 +786,11 @@ EOF
       if ($bfbcomp != ${spval}) then
         set fbfbcomp = ${machcomp}_${bfbcomp_test_grid_pes}${bfbcomp_soptions}
       endif
+    else
```
phil-blain: we are far below the ` if (${docase} == 0) then` line here, so for documentation purpose maybe a comment after the `else` would help to clarify where we are in the logic?

doc/source/cice_index.rst
```diff
@@ -28,25 +28,25 @@ either Celsius or Kelvin units).
    "a4Ds", "history field accumulations, 4D categories, vertical snow", ""
    "a4Df", "history field accumulations, 4D categories, fsd", ""
    "a_min", "minimum area concentration for computing velocity", "0.001"
-   "a_rapid_mode", ":math:`{\bullet}` brine channel diameter", ""
-   "add_mpi_barriers", ":math:`\bullet` turns on MPI barriers for communication throttling", ""
-   "advection", ":math:`\bullet` type of advection algorithm used (‘remap’ or ‘upwind’)", "remap"
+   "a_rapid_mode", "brine channel diameter", ""
+   "add_mpi_barriers", "turns on MPI barriers for communication throttling", ""
+   "advection", "type of advection algorithm used (‘remap’ or ‘upwind’)", "remap"
```
phil-blain, line 33: Here and below, all the bullets are removed. From what I understand, these were used to denote namelist settings, and now that we have a separate table of namelist options, the bullets are redundant. Is that the reasoning for this change ? Maybe a mention of that rationale in the commit message would be good ?

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -1116,6 +1121,16 @@ subroutine input_data
          abort_list = trim(abort_list)//":22"
       endif
 
+      if(histfreq_base /= 'init' .and. histfreq_base /= 'zero') then
```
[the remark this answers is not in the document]
apcraig: We don't, but I think it's reasonable to trap them during the initialization for efficiency.  I think it's also proper to have the if loop in the calendar operate only on valid values.

cicecore/shared/ice_calendar.F90
```diff
@@ -402,9 +452,31 @@ subroutine calendar()
...
+         elapsed_hours = elapsed_days * hours_per_day + hour
+      elseif (dumpfreq_base == 'init') then
+         elapsed_years = myear - year_init
+         elapsed_months =  elapsed_years * months_per_year + (mmonth - month_init)
+         elapsed_days = compute_days_between(year_init,month_init,day_init,myear,mmonth,mday)
+         elapsed_hours = elapsed_days * hours_per_day + hour
+      else
+         write(nu_diag,*) trim(subname),' ERROR dumpfreq_base not recognized, ',trim(dumpfreq_base)
+         call abort_ice(subname//'ERROR: dumpfreq_base value invalid')
+      endif
```
phil-blain, line 468: I think it would be cleaner to have a separate subroutine that computes `elapsed_{years,months,days,hours}` depending on the 'init' and 'zero' setting, and have separate variables for the restart and history parts of the subroutine. This way the computation themselves are not repeated in the code.

phil-blain (commented):
Thanks for these fixes @apcraig. I left a few notes but is mostly looks good. It's nice to see that it was not too hard to add support for both 'zero' and 'init' modes.

cice.setup
phil-blain: oups, I l left this comment on an earlier commit of the series, the most recent does it differently so it's ok.

configuration/scripts/ice_in
```diff
@@ -24,6 +24,7 @@
     pointer_file   = './ice.restart_file'
     dumpfreq       = 'd'
     dumpfreq_n     = 1
+    dumpfreq_base  = 'init'
```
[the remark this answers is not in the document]
apcraig: Will do.

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -162,6 +162,8 @@ setup_nml
    "", "``m``", "write restart every ``dumpfreq_n`` months", ""
    "", "``y``", "write restart every ``dumpfreq_n`` years", ""
    "", "``1``", "write restart every ``dumpfreq_n`` time step", ""
+   "``dumpfreq_base``", "init", "restart output frequency is relative to the model initial date", "init"
```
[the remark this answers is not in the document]
apcraig: This is a very good question.  'init' means relative to year_init, month_init, day_init.  So while those generally do not play a role after the initial run, they now do if freq_base is 'init'.  Is there a reason to support a mode where we create a freq_base relative to the start of the current run.  It probably would not be hard to do that.  It could create chaos in the sense that if you write output every 4 days and run 10 days at a time, you'd have output on day 4, 8, 14, 18 in that mode.  In zero mode you'd have it every 4 days with a possible phase shift (i.e. 1, 5, 9, 13, 17).  And in init mode you'd have it on days 4, 8, 12, 16.  As I alluded to, there are so many things we could do.

Let me know if we need another mode here.

cicecore/shared/ice_calendar.F90
apcraig: They are also set in ice_init in the standard method.  The reason that I set them here as well is to help the unit testing work better.  calchk doesn't go thru any CICE initialization, so having defaults set here means I don't have to explicitly set them when I run the calchk unit test.  One could argue whether that is the best way to do it.  I also made a similar change to ndtd in ice_calendar.

doc/source/cice_index.rst
apcraig: Interesting, I couldn't figure out why those bullets were there.  It just looked like some had them and others didn't which seemed inconsistent to me.  So I just removed them all.   If we need to restore and review all the current variables, we can do that.  Sorry about that.

cicecore/shared/ice_calendar.F90
apcraig: Good point.  I don't know if we need different variables but creating a new subroutine is a good idea.

cice.setup
apcraig: This has actually changed in a more recent commit.  It's now outside the docase if test and establishes the standard.  It's only with suites that we invoke the more complex implementation.

doc/source/user_guide/ug_implementation.rst
```diff
@@ -571,7 +578,62 @@ to set the initial model time to the namelist values defined by
...
+The history (:ref:`history`) and restart (:ref:`restartfiles`) 
+outputs and frequencies are specified in namelist and
+are computed relative to a reference date defined by the namelist
+``histfreq_base`` and ``dumpfreq_base``.  Valid values for each are 
+``zero`` and ``init``.  If set to ``zero``, all output will be relative 
+to the absolute reference date, 0000-01-01-00000.  This is the default
+value for ``histfreq_base``, so runs with different initial
+dates will have identical output.  If set to ``init``, all frequencies
+will be relative to the model init time specified by ``year_init``,
+``month_init``, and ``day_init``.  This is the default for
```
phil-blain, line 615: Maybe we would want to emphasize specifically that `sec_init` is _not_ taken into account ?

apcraig:
I have updated the implementation.  I fixed several issues and also updated the initial/restart implementation after discussions with @eclare108213.  The documentation was also update to reflect these changes.

apcraig:
Also, I am running a full test suite on cheyenne now.

cicecore/shared/ice_calendar.F90
```diff
@@ -122,14 +126,18 @@ module ice_calendar
...
+         histfreq_base = 'init', & ! history frequency basetime ('zero', 'init')
+         calendar_type             ! define calendar type
 
       ! PRIVATE
 
       integer (kind=int_kind) :: &
          hour         ! hour of the day
 
+      integer (kind=int_kind), parameter :: &
+         myear_max = 200000           ! maximum year, limited by integer overflow in elapsed_horus
```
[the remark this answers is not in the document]
apcraig: Fixed.

configuration/scripts/ice_in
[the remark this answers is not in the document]
apcraig: Done.

doc/source/user_guide/ug_case_settings.rst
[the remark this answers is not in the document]
apcraig: @eclare108213 and I decided we didn't need another mode at this time.  Documentation has been updated to clarify.

cicecore/shared/ice_calendar.F90
apcraig: I implemented a new subroutine to improve code reuse.

doc/source/user_guide/ug_implementation.rst
```diff
@@ -684,18 +687,65 @@ layers and the ice thickness distribution defined by ``kcatbound`` = 0.
...
+   +================+==========================+======================================+========================================+
+   | `initial`      | `none`                   | not used                             | no ice, namelist defines start date    |
+   +----------------+--------------------------+--------------------------------------+----------------------------------------+
+   | `initial`      | `default`                | not used                             | latitude dependent internal ic,        |
+   |                |                          |                                      | namelist defines start date            |
+   +----------------+--------------------------+--------------------------------------+----------------------------------------+
+   | `initial`      | *filename*               | false                                | read file, namelist defines start date |
+   +----------------+--------------------------+--------------------------------------+----------------------------------------+
+   | `initial`      | *filename*               | true                                 | read file, file defines start date     |
```
phil-blain, line 744: Maybe be a little more verbose ?

```suggestion
   | `initial`      | *filename*               | false                                | read ice state from file, namelist defines start date |
   +----------------+--------------------------+--------------------------------------+----------------------------------------+
   | `initial`      | *filename*               | true                                 | read file state from file, file defines start date     |
```

doc/source/user_guide/ug_implementation.rst
```diff
@@ -684,18 +687,65 @@ layers and the ice thickness distribution defined by ``kcatbound`` = 0.
...
+is forced to true and ``ice_ic`` plays no role.  When ``runtype`` =
+`initial`, ``ice_ic`` has three options, ``none``, ``default``,
+or *filename*.  These initial states are no-ice, latitudinal dependent
+ice, and ice defined by a file respectively.  In `initial` mode,
+``use_restart_time`` should generally be set to false and the initial
+time is then defined by ``year_init``, ``month_init``, ``day_init``, 
+and ``sec_init``.  These combinations options are summarized in 
+:ref:`tab-ic`. 
+
+Restart files and initial condition files are generally the same file.
```
phil-blain, line 703: I'm not sure about the formulation here. Maybe "are generally in the same format" ?

doc/source/user_guide/ug_implementation.rst
```diff
@@ -684,18 +687,65 @@ layers and the ice thickness distribution defined by ``kcatbound`` = 0.
...
-“./restart/[filename]". The first option is convenient when repeatedly
-starting from a given file when subsequent restart files have been
-written. With this arrangement, the tracer restart flags can be set to
-true or false, depending on whether the tracer restart data exist. With
-the second option, tracer restart flags are set to ‘continue’ for all
-active tracers.
+Three namelist variables generally control model initialization, ``runtype``,
+``ice_ic``, and ``use_restart_time``.  The valid values for ``runtype``
+are ``initial`` or ``continue``.  When ``runtype`` = `continue`, the
+initial condition is defined by the rpointer file, ``use_restart_time``
```
phil-blain, line 693: I'm not sure I like the name "rpointer". I would prefer something like "ice pointer file" or something of sorts... and maybe add a "(see below)" because it's the first time the name is introduced.

doc/source/user_guide/ug_implementation.rst
```diff
@@ -684,18 +687,65 @@ layers and the ice thickness distribution defined by ``kcatbound`` = 0.
...
+In this mode, the CICE model writes out a small text (rpointer) file
+to the run directory that names the most recent restart file.   On
+restart, the model reads the rpointer file which defines the
+name of the restart file.  The model then reads that restart file.
+By having this feature, the ice namelist does not need to be constantly
+updated with the latest
+restart filename, and the model can be automatically resubmitted.
+Manually editing the rpointer file in the middle of a run will reset
+the restart filename, and this makes it possible to relatively easily
+backup a run to rerun.
```
phil-blain, line 723: I'm not sure exactly what you mean here... also I guess you meant to write "back up" and not "backup" ?

phil-blain (commented):
Thanks for these additional clean-ups Tony, this was really long overdue I think. Especially the "Ice initialization" table in the doc, the new version is really clearer. I think it makes sense to start deprecating the `restart` namelist flag,  since it was reset already internally under several conditions.

doc/source/user_guide/ug_implementation.rst
apcraig: Done.

doc/source/user_guide/ug_implementation.rst
apcraig: Done.

doc/source/user_guide/ug_implementation.rst
apcraig: I have simplified that sentence.

doc/source/user_guide/ug_implementation.rst
apcraig: I updated this.

doc/source/user_guide/ug_implementation.rst
apcraig: rpointer is a term in CESM.  I have changed those to "pointer".  thanks for catching that.

doc/source/cice_index.rst
phil-blain: I think it's ok to delete them, and remove the mention about them at the top of the index, like you do in your latest version. Closing this conversation.

phil-blain: ... turns out I can't do that :p It turns out I instead hid my first comment above.

phil-blain (approved):
Thanks for these fixes @apcraig. I left a few notes but is mostly looks good. It's nice to see that it was not too hard to add support for both 'zero' and 'init' modes.

phil-blain:
@apcraig @eclare108213 @dabail10 :

I remembered this issue: https://github.com/CICE-Consortium/CICE/issues/527. I just skimmed it and from what I understand it would be closed by this here PR with the latest changes to the restart flags, right ? If so we could also check it in the list at https://github.com/CICE-Consortium/CICE/issues/567.

apcraig:
I think this does address #527.  In general, we could support use_restart_time=.false. with runtype='continue' and were up to now.  But after some discussion, it was decided to simplify the options a bit, so we're eliminating that feature (for now).  The issue in #527 could (should) have also been solved by modifications in the nuopc cap or changes to the namelist generation in CESM, and maybe that was done or is still worth doing.

Unless other disagree, I think this can close 527 and will add that comment in the PR documentation.

apcraig:
I tested the latest changes and everything still works.  This could be merged unless there are other comments to address.

apcraig:
Unless I hear any concerns or someone gets to it before me, I'll merge this PR in the next couple days.  Thanks.

phil-blain:
Thanks a lot for these fixes @apcraig ! The history frequency bug was the last missing bit for me to wrap up the first phase of our project for updating to CICE6 in our forecasting systems!

Issue #595, opened 2021-04-08
bug in gather_global_ext in ice_gather_scatter.F90


TillRasmussen:
@phil-blain @apcraig @dabail10 
In ice_gather_scatter.F90 there is an interface gather_global_ext,.
This gathers data from all blocks and fills the interior first. After this it fills out the ghost cells to the east, west, south or north.
This is done when of of the below conditions are fulfilled
this_block%iblock == 1, this_block%iblock == 1, this_block%jblock == nblocks_y, this_block%iblock == nblocks_x
The corners are filled when two of the above conditions are fulfilled. E.g 
this_block%iblock == 1 and this_block%iblock == 1 will fill the south western corner of the ghost cells.
If this block is masked out nothing will be done and the value will be whatever the compiler chooses. 
This happens in test freya_gnu_smoke_gx3_16x2x1x1x800_debug_droundrobin_run2day where each block has a size of 1x1

The 1d evp solver requires a global gather in order to run parts on the master task.
With the gnu compiler I forced an initialization to NAN, which for the velocities caused an initialization to NAN. This spreads out to the rest of the array during the iterations and fails when it is used to calculate indexes in the transport routine.
I found two solutions. I have tested number 1

1/  initialize the global array to the special value.
I replaced 
   if (present(spc_val)) then
      special_value = spc_val
   else
      special_value = spval_dbl
   endif

with

  if (present(spc_val)) then
      special_value = spc_val
   else
      special_value = spval_dbl
   endif
ARRAY_G=special_value._
 
2/ This is not tested but in the same function there is a cpp flag CICE_IN_NEMO
I think that the code in here will fix the issue as well as it handles masked cells the same way as unmasked cells.

Questions
Do you prefer solution 1 or 2 (if 2 works)
@phil-blain Are you using this code in nemo_cice. I cannot see any references to it in the rest of the code nor in the NEMO4 code.What I want to is to delete this if we aim for solution 1. If it is solution 2 I would remove the cpp flag

phil-blain:
Apart from the multiple calls in **`ice_dyn_evp_1d`**, all calls to `gather_global_ext` are in subroutines that are only called if `restart_ext = .true.` in the namelist. I think none of our configurations use this setting, but I'll confirm with people that know more about this.

daveh150:
[...]

PR #618, opened 2021-07-16
Fix 1x1 and 2x2 blocksize error

apcraig:
## PR checklist
    Fix 1x1 and 2x2 blocksize error, update initialization of ARRAY_G in global_gather_ext
- [X] Developer(s): 
    apcraig
    Full test suite run on cheyenne with gnu.  All results are bit-for-bit except 1x1 and 2x2 results which don't change answers but do change restart files slightly due to the fix in missing blocks.  See cheyenne gnu results https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#85531cf8f8ae93f628344df6ecfa4be12e9629d8
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

- Fix 1x1 and 2x2 non-bit-for-bit results.  It turns out the 1x1 and 2x2 blocks were
  bit-for-bit physically, just not technically.  The problem was that a set 
  of blocks were "land block eliminated"
  near the equator incorrectly.  These blocks never have sea ice, so it's not a problem
  scientifically, but it changes the masking and restart files.  The fix is a one line change
  in ice_domain.F90 which that sets the minimum "flat" value to 1.  At the equator, flat 
  was set to zero and this eliminated those gridcells incorrectly.  This error could only
  happen in cases where the gridcells were within 0.5 degrees of the equator, and a set of
  conditions were required where an active gridcell was set to inactive which then resulted
  in the entire block being eliminated.  This could really only happen with the smallest blocks
  which is why it only showed up with 1x1 and 2x2.  
- Update the initialization of ARRAY_G in the gather_global_ext interfaces.  Get rid of
  older, more complex approach which left some blocks uninitialized.
- Set psalt(n) to zero by default.  Was uninitialized in cases where vice=0.
- Add a test to check bit-for-bit in log files for 1x1 blocks with bfbflag set to reprosum.

### DIFF
--- a/cicecore/cicedynB/analysis/ice_diagnostics.F90
+++ b/cicecore/cicedynB/analysis/ice_diagnostics.F90
@@ -825,6 +825,7 @@
                      enddo
                   endif
                endif
+               psalt(n) = c0
                if (vice(i,j,iblk) /= c0) psalt(n) = work2(i,j,iblk)/vice(i,j,iblk)
                pTsfc(n) = trcr(i,j,nt_Tsfc,iblk)   ! ice/snow sfc temperature
                pevap(n) = evap(i,j,iblk)*dt/rhoi   ! sublimation/condensation
--- a/cicecore/cicedynB/infrastructure/comm/mpi/ice_gather_scatter.F90
+++ b/cicecore/cicedynB/infrastructure/comm/mpi/ice_gather_scatter.F90
@@ -636,6 +636,7 @@
    else
       special_value = spval_dbl
    endif
+   ARRAY_G = special_value
 
    nx = nx_global + 2*nghost
    ny = ny_global + 2*nghost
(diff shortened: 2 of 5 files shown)


apcraig:
I think this can close #595.  This addresses the 1x1 and 2x2 block issue as well as the ARRAY_G issue.  The evp1d solver is being addressed in #568.

It would be nice to merge this by Friday afternoon if we can so it can be included in weekend testing.

eclare108213 (approved):
cool

phil-blain (approved):
I'm not that familiar with this part of the code, especially the MPI gather-scatter stuff, but removing a bunch of lines and replacing it with a single line is always welcome!

Thanks.

PR #633, opened 2021-09-14
sensible+latent heatfluxes using linear bulk formula

mhrib:
See also:
https://github.com/CICE-Consortium/Icepack/pull/371

## PR checklist
    For stability boundary layer (atmbndy='default'): Use linear bulk formula for sensible+latent heat fluxes but preserve stress formulation.
- [x] Developer(s): 
    @mhrib 
   @apcraig @eclare108213 
  19 measured results of 19 total results
  19 of 19 tests PASSED
  0 of 19 tests PENDING
  0 of 19 tests MISSING data
  0 of 19 tests FAILED
    - [x] bit for bit  (_for default settings_)
    - [x] more substantial 
- Does this PR create or have dependencies on CICE or any other models?
    - [x] Yes
    - [ ] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

By introducing the "simpler" linear bulk formula for sensible+latent heat fluxes, but preserve the temperature/humidity formulation for stress, the DMI-HYCOM-CICE (ERA5) code got more stable for long runs (several of years) without significant changes in the overall sea-ice cover/thickness. Without these modifications, the "Flux Error" (ferr) - rarely - became very high in special circumstances with low wind speeds and Tair>Tocn. In these situations the Jordan et al (1999) formulation might have it's limitations. These high Flux Errors seems to dissapear with the simpler formulations for sensible and latent heatfluxes for our setup at least.

The results are not BFB with the new setting turned on. However it is implemented turned off, which results in BFB results compared to the unmodified code.

NOTE:
* Both ICEPACK and CICE have to be updated, as the "atmbndy" namelist flag goes into both codes with default renamed to 'similarity'
* Icepack PR: https://github.com/CICE-Consortium/Icepack/pull/371


eclare108213 (changes requested):
I don't see any major problems but have a couple of picky suggestions. Please remove the word "traditional" everywhere and use "linear bulk formula" for the PR title.  Also, the heatflux_linear flag does't apply to all heat fluxes, so let's find a more specific name for it.  I usually refer to sensible+latent as the turbulent heat fluxes.  Could use turbheat_linear or turbheatflux_linear or something like that?

eclare108213:
Actually, looking at the Icepack PR, I'm thinking that the namelist flag should instead be an additional option for atmbndy, e.g. instead of 'default' and 'constant, use 'similarity' (referring to Monin-Obukhov similarity theory), 'mixed' and 'constant'.

apcraig:
If it makes sense to implement the new feature as a new option of an existing namelist and icepack argument, that would be good.

mhrib:
Good point using an existing namelist argument. More, I am not fan of my own invented namelist parameter name "heatflux_linear". I'll look at it using 'atmbndy' namelist parameter instead as suggested by @eclare108213

mhrib:
Rewritten using 'atmbndy' namelist variable as suggested.
@eclare108213  I can't rename the PR removing 'traditional' but feel free to change it. Alternative I can create a new PR.

mhrib:
Update code using 'atmnbdy' which now allows three options: 'similarity' (default), 'constant', or 'mixed'
Also update documentation section 3.4.3.10.forcing_nml and section 5 (index of variables) with the above info

eclare108213 (approved):
This is good.  Icepack will need to be updated here after that PR is merged.

apcraig:
I will add one other thing.  I think we should continue to support "default", even if it's not documented, for backwards compatibility.  On the CICE side, if "default" is set in namelist, on input it should be reset to "similarity" and a brief message written.  The same can be done in the Icepack driver.  In Icepack, if "default" is passed in, it should treat it the same as "similarity" either by renaming it as it's passed in or by making "default" and "similarity" appear together in all the logic.  Actually Icepack probably does this now because....

One other thing that seems to be the case in Icepack, there is no check anywhere that the atmbndy value is reasonable.  The only place it's used is in icepack_atmo.F90.  The logic there allows any value to be valid.  Is that something we can/should improve?  That could be either by checking atmbndy value on input to icepack_parameters (which we don't typically do) or by fixing the if/else logic explicitly adding an abort if the atmbndy is not valid.

mhrib:
Updated:
1) Update “Atmosphere” section 2.1.1 with almost the suggested text. (in icepack, not here)
2) Abort if 'atmbndy' setting is unknown (done via. abort in icepack)
3) For backward compability: 'atmbndy'=default is renamed to similarity. And a WARNING message is given in that case to the log.

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -1214,6 +1214,14 @@ subroutine input_data
          endif
       endif
 
+      if (trim(atmbndy) == 'default') then
+         if (my_task == master_task) then
+            write(nu_diag,*) subname//' WARNING: atmbndy = default'
+            write(nu_diag,*) subname//' WARNING:   For backward compability, set atmbndy = similarity'
```
phil-blain, line 1220: Small wording suggestion:
```suggestion
            write(nu_diag,*) subname//' WARNING: "atmbndy = default" is deprecated, use "atmbndy = similarity"'
            write(nu_diag,*) subname//' WARNING:   setting atmbndy = similarity'
```

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -1641,13 +1649,18 @@ subroutine input_data
          write(nu_diag,1010) ' rotate_wind      = ', rotate_wind,' : rotate wind/stress to computational grid'
          write(nu_diag,1010) ' formdrag         = ', formdrag,' : use form drag parameterization'
          write(nu_diag,1000) ' iceruf           = ', iceruf, ' : ice surface roughness at atmosphere interface (m)'
-         if (trim(atmbndy) == 'default') then
-            tmpstr2 = ' : stability-based boundary layer'
+         if (trim(atmbndy) == 'constant') then
+            tmpstr2 = ' : boundary layer uses bulk transfer coefficients'
```
eclare108213, line 1653: I would change 'bulk' to 'constant' in this line.  If I remember correctly, the overall parameterization we use is considered a 'bulk' formulation, and we're just choosing whether the coefficients are constant or not.  Anyone please correct me if that's wrong.  (And yes I used 'bulk' in the original comment!)

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -1641,13 +1649,18 @@ subroutine input_data
...
+                 trim(atmbndy) == 'mixed') then
             write(nu_diag,1010) ' highfreq         = ', highfreq,' : high-frequency atmospheric coupling'
             write(nu_diag,1020) ' natmiter         = ', natmiter,' : number of atmo boundary layer iterations'
             write(nu_diag,1002) ' atmiter_conv     = ', atmiter_conv,' : convergence criterion for ustar'
-         elseif (trim(atmbndy) == 'constant') then
-            tmpstr2 = ' : boundary layer uses bulk transfer coefficients'
+            if (trim(atmbndy) == 'similarity') then
+               tmpstr2 = ' : stability-based boundary layer'
+            else
+               tmpstr2 = ' : stability-based boundary layer, but constant for sensible+latent heatfluxes'
```
eclare108213, line 1662: How about 
' : stability-based wind stress, constant-coefficient sensible+latent heat fluxes'

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -528,8 +528,10 @@ forcing_nml
    :widths: 15, 15, 30, 15 
 
    "", "", "", ""
-   "``atmbndy``", "``constant``", "bulk transfer coefficients", "``default``"
-   "", "``default``", "stability-based boundary layer", ""
+   "``atmbndy``", "string", "bulk transfer coefficients", "``similarity``"
+   "", "``similarity``", "stability-based boundary layer", ""
+   "", "``constant``", "constant-based boundary layer", ""
+   "", "``mixed``", "stability-based, but constant for sensible+latent heatfluxes", ""
```
eclare108213, line 534: Likewise,
'stability-based wind stress, constant-coefficient sensible+latent heat fluxes'

eclare108213 (changes requested):
I have some picky requests re wording in the comments, nothing significant.

cicecore/cicedynB/general/ice_init.F90
mhrib: rewritten

cicecore/cicedynB/general/ice_init.F90
mhrib: Changed / rewritten

cicecore/cicedynB/general/ice_init.F90
mhrib: Changed / rewritten

doc/source/user_guide/ug_case_settings.rst
mhrib: Changed / rewritten

cicecore/cicedynB/general/ice_init.F90
```diff
@@ -1216,8 +1216,8 @@ subroutine input_data
 
       if (trim(atmbndy) == 'default') then
          if (my_task == master_task) then
-            write(nu_diag,*) subname//' WARNING: atmbndy = default'
-            write(nu_diag,*) subname//' WARNING:   For backward compability, set atmbndy = similarity'
+            write(nu_diag,*) subname//' WARNING: "atmbndy = default" is decrecated'
```
eclare108213, line 1219: spelling error: deprecated
Are the double quotes needed in these write statements?

eclare108213 (changes requested):
Getting close

eclare108213 (approved):

cicecore/cicedynB/general/ice_init.F90
eclare108213: Thank you!

eclare108213:
I've just merged https://github.com/CICE-Consortium/Icepack/pull/371, so now I think you should update Icepack here and then we'll merge this PR.

mhrib:
@eclare108213 I think I have updated Icepack with all the suggested changes, but maybe I have missed one (two, three, ..) comments?

apcraig (approved):
Do we need to add a new test case or two?  Otherwise, looks good to me.

mhrib:
@apcraig I don't think it is necessary with new testcase, as the default=similarity is exactly the same as previous.
However if necessary, I think there should be a new testcase for both atmbndy = constant (same as before) AND atmbndy = mixed (new).

apcraig:
Can you add those @mhrib?

mhrib:
@apcraig I am a bit uncertain what to test. All three settings of atmbndy will give different results, but that is on purpose...

apcraig:
@mhrib, I think you should just add a couple new restart tests with the other atmbndy settings.  You should add a set_nml.atmbndyconst and set_nml.atmbndymixed in configuration/scripts/options/.  Then you should add two new tests to the base_suite (configurations/scripts/tests/base_suite.ts) that would simply be something like

```
restart        gx3     12x2        atmbndyconst
restart        gx3      8x2        atmbndymixed
```

Then just make sure those tests pass.  I think it would be fine to just run each test manually to confirm.  

The other way to do this is to modify a couple of "alt" tests to add the constant and mixed options to those.  That would be more efficient and I would support that assuming it's clear which alt tests would be reasonable to modify.  In that case, you'd just modify the set_nml.alt* files and not add the new set_nml files.  Either way seems reasonable.  Thanks.

mhrib:
@apcraig Thanks for explanation for creating relevant tests.
Now adding test-cases following your first approach adding 2x restart tests using atmbndy={constant,mixed}

All test passes with "SUCCESSFULLY" on the DMI computer (freya), except for three tests, that tries to allocate more OpenMP processors than cores/node, which is obviously not valid. Ie. 118 good and 3 not started.

So I think the PR is about ready to be merged...

apcraig (approved):
I have a slight preference for not having special characters in the set_nml filenames (set_nml.atmbndyconst vs set_nml.atmbndy_const) as I worry about parsing a little and for consistency.  But I think this works.  Will approve.

mhrib:
@apcraig I have changed to "non-underscore".
Not sure I fully understand the problem wrt. underscore "_" in text unless we're talking LaTeX interpretter, text writing in plotting routines etc.. Nevertheless, the functionality remains unchanged.

eclare108213:
@mhrib Thanks for your patience and diligence, working through all the various requests!  This looks good - thanks also for adding this flexibility to the code.

mhrib:
@eclare108213 I really appreciate the various requests.
It has ended up in a much simpler solution without introducing an extra namelist parameter as original purposed. Thereby much easier to understand especially in combination with the requested documentation. We at DMI appreciate the inclusion of this new feature, which makes future updates of CICE easier for us. So a win-win for all :-)

PR #644, opened 2021-10-14
Update test features: add create_fails script, QC process test, and wait on baseline results

### LINKED ISSUE #609: Re-running selected cases in suites
Is there a script option to re-submit only the failed tests in a suite, without having to redo the entire suite?  That would be super handy.  E.g. base_suite currently has 328 tests and takes a long time to get through the queue; 74 of the tests are failing in my current development sandbox and I'm wondering how many of them will start working when I fix something.

apcraig:
## PR checklist
    Add create_fails script, QC process test, and batch wait on baseline results when using --diff in test suites
- [X] Developer(s): 
    apcraig
    All tests are bit-for-bit with testing on cheyenne, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#adab3d33996089d66c9a69a26815409c3edc584c.  Tested scripts several ways/times.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [X] Yes, qc process test
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

- Add create_fails script.  This looks at the results of a test suite and creates a new test suite file (fails.ts) with only the failed tests.  That can be edited and passed into cice.setup (--suite fails) to continue code validation.  Hopefully this addresses and closes #609.
- Add qcchk and qcchkf tests which use the QC validation process to compare to other results.  qcchk verifies the QC test passes and qcchkf verifies that the QC test fails.  Add gx3 qc test to prod_suite.ts and verify it works on cheyenne.  Had to add the appropriate python env setup to the cheyenne env files.  The current tests just validate the QC process in bfbcomp (--diff) mode.  There is no ability yet to run QC comparisons against prior versions of the code as this would require a gx1 qc run with appropriate qc flags (let me see if I can make that happen).
- Add batch checks when bfbflag (--diff) checking is done to make sure the baseline is complete before the comparison is done.  Up to now, there has been an issue if the baseline test is NOT completed before a test that is comparing to that baseline.  With this implementation, if the baseline job is still in the queue, then the comparison waits for it to complete in similar style to poll_queue.csh.


eclare108213 (approved):
Nice.  Approving based only on visual inspection.

configuration/scripts/tests/create_fails.csh
```diff
@@ -0,0 +1,24 @@
...
+  set test = `echo $line | cut -d "_" -f 3`
+  set grid = `echo $line | cut -d "_" -f 4`
+  set pes  = `echo $line | cut -d "_" -f 5`
+  set opts = `echo $line | cut -d "_" -f 6- | sed 's/_/,/g'`
+  echo "$test  $grid  $pes  $opts" >> fails.ts
+end
+
+rm $tmpfile
+echo "$0 done"
+echo "Not passed tests written to file...... $outfile"
```
phil-blain, line 22: Maybe "Failed tests" instead ?

configuration/scripts/tests/create_fails.csh
```diff
@@ -0,0 +1,24 @@
...
+echo "To setup a new test suite, try something like"
+echo "  ./cice.setup --suite fails.ts ..."
```
phil-blain: does this work ? I guess you first have to move `fails.ts` to `configuration/scripts/tests/` in the sandbox ? it this what you mean by "something like" ?

configuration/scripts/tests/create_fails.csh
apcraig: Failed tests isn't quite accurate.   Failed != Not passed.  But I think you're right on reflection that failed is probably a better word here.  I'll change it.

configuration/scripts/tests/create_fails.csh
apcraig: You're right that It wouldn't work exactly as described.  You can move fails.ts to the cice.setup directory too.  It works from there.  But you're right, maybe I'll clarify that a bit.  I don't want to get too deep in the weeds though.

apcraig: Updated.

PR #649, opened 2021-10-28
Implementation of plastic potential for VP and EVP.

JFLemieux73:

## PR checklist
    Implementation of plastic potential approach of Ringeisen et al. 2021
- [x] Developer(s): 
    @JFLemieux73 
    @phil-blain 
337 measured results of 337 total results
337 of 337 tests PASSED
0 of 337 tests PENDING
0 of 337 tests MISSING data
0 of 337 tests FAILED
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 
@TillRasmussen  FYI: I have not modified the 1d-evp. This is why I kept ecci=1/e_yieldcurve**2.

### DIFF
--- a/doc/source/science_guide/sg_dynamics.rst
+++ b/doc/source/science_guide/sg_dynamics.rst
@@ -351,7 +351,7 @@ Following again the LKD method, the seabed stress coefficients are finally expre
 .. _internal-stress:
 
 ********************************
-Internal stress and strain rates
+Internal stress
 ********************************
 
 For convenience we formulate the stress tensor :math:`\bf \sigma` in
@@ -413,7 +413,7 @@ described in the `Icepack Documentation <https://cice-consortium-icepack.readthe
 Two modifications to the standard VP rheology of :cite:`Hibler79` are available.
 First, following the approach of :cite:`Konig10` (see also :cite:`Lemieux16`), the
 elliptical yield curve can be modified such that the ice has isotropic tensile strength.
-The tensile strength :math:`T_p` is expressed as a fraction of :math:`P`, that is :math:`T_p=k_t P`
+The tensile strength is expressed as a fraction of :math:`P`, that is :math:`k_t P`
 where :math:`k_t` should be set to a value between 0 and 1 (this can
 be changed at runtime with the namelist parameter ``Ktens``).
 
@@ -423,6 +423,8 @@ yield curve (:cite:`Ringeisen21`). :math:`e_f` and :math:`e_g` are respectively 
 can be set in the namelist. The plastic potential can lead to more realistic fracture angles between linear kinematic features. :cite:`Ringeisen21` suggest to set :math:`e_f` to a value larger than 1 and to have :math:`e_g < e_f`.
 
 By default, the namelist parameters are set to :math:`e_f=e_g=2` and :math:`k_t=0` which correspond to the standard VP rheology.
+
+There are four options in the code for solving the sea ice momentum equation with a VP formulation: the standard EVP approach, a 1d EVP solver, the revised EVP approach and an implicit Picard solver. The modifications to the yield curve and to the flow rule described above are available for these four different solution methods.  
 
 .. _stress-evp:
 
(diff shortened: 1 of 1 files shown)


eclare108213 (commented):
The documentation changes were made only in the VP section.  Is this new yield curve only available for VP, or could it be used for the other dynamics options too?

JFLemieux73:
The VP section is an introduction for the implicit and EVP approaches. It is in this section that we introduce the viscous coefficients and Delta. The new plastic potential affects the definition of eta (viscous coeff) and Delta. It is therefore available for the implicit solver (VP), EVP and rEVP. I think the first sentence in the EVP section suggests this:

"In the EVP model the internal stress tensor is determined from a regularized version of the VP constitutive law"

Do we need more info?

eclare108213:
It wasn't obvious to me, reading all of those sections, that those options were available for all of the (E)VP variants.  Please make that explicit.  When we rework the dynamics drivers etc we should take a look at the documentation too (or maybe vice-versa, as guidance).

doc/source/science_guide/sg_dynamics.rst
```diff
@@ -395,28 +386,43 @@ Viscous-Plastic
...
+
+The ice strength :math:`P` is a function of the ice thickness distribution as
+described in the `Icepack Documentation <https://cice-consortium-icepack.readthedocs.io/en/master/science_guide/index.html>`_.
+   
+Two modifications to the standard VP rheology of :cite:`Hibler79` are available.
+First, following the approach of :cite:`Konig10` (see also :cite:`Lemieux16`), the
+elliptical yield curve can be modified such that the ice has isotropic tensile strength.
+The tensile strength :math:`T_p` is expressed as a fraction of :math:`P`, that is :math:`T_p=k_t P`
+where :math:`k_t` should be set to a value between 0 and 1 (this can
+be changed at runtime with the namelist parameter ``Ktens``).
```
phil-blain, line 418: It would be clearer if `T_p` appeared in the equations below, or at least mention how its effect is incorporated in the viscous coefficients (since `k_t` is removed from the equations).

cicecore/cicedynB/dynamics/ice_dyn_shared.F90
```diff
@@ -230,10 +232,10 @@ subroutine set_evp_parameters (dt)
       !dtei = c1/dte              ! 1/s
       dtei = real(ndte,kind=dbl_kind)/dt 
 
-      ! major/minor axis length ratio, squared
-      !ecc  = e_ratio**2
-      !ecci = c1/ecc               ! 1/ecc
-      ecci = c1/e_ratio**2               ! 1/ecc
+      ! variables for elliptical yield curve and plastic potential
+      epp2i = c1/e_plasticpot**2
```
phil-blain, line 236: not sure I like that name.. but I don't have anything much better to propose :P

phil-blain (commented):
Thanks @JFLemieux73. It's nice how this can be implemented with so little changes to the code. I left a few comments below.

doc/source/science_guide/sg_dynamics.rst
```diff
@@ -395,28 +386,43 @@ Viscous-Plastic
 The VP constitutive law is given by
 
 .. math::
-   \sigma_{ij} = 2 \eta \dot{\epsilon}_{ij} + (\zeta - \eta) D_D - P_R(1 - k_t)\frac{\delta_{ij}}{2}
+   \sigma_{ij} = 2 \eta \dot{\epsilon}_{ij} + (\zeta - \eta) D_D - P_R\frac{\delta_{ij}}{2}
```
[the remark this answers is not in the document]
JFLemieux73: The replacement pressure now includes the (1-k_t) term. This is consistent with Ringeisen et al. 2021.

cicecore/cicedynB/dynamics/ice_dyn_shared.F90
JFLemieux73: At least it looks like the previous variable ecci...

doc/source/science_guide/sg_dynamics.rst
JFLemieux73: I have removed T_p in the doc.

JFLemieux73:
@eclare108213 I agree we should have a look and probably modify some of the doc for the dynamics. Note that I have reordered some subsections (EAP  is now at the end as it is really different than the other ones). I also think we should rework the revised EVP subsection. Some eqs should go in the momentum section and not in the internal stress section.

JFLemieux73:
@eclare108213 I think we should go ahead with this PR. I plan to go back to the documentation soon. I think we can make it simpler. For example we have two sections EVP and two sections VP...these can be combined.

doc/source/science_guide/sg_dynamics.rst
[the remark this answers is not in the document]
eclare108213: I'm a little confused about this too.  I think I understand what's happening in the code.  What's confusing is that P and replacement pressure aren't quite the same, and in our docs both are referenced to old publications, which don't include the new tensile strength.  So when we clean up the dynamics documentation as part of the overall C-grid effort, let's fix that too.

eclare108213 (approved):
I'm willing to approve and merge these changes, under the assumption that the documentation will be fixed fairly soon.

phil-blain:
One small point, maybe we want to add a new test with a non-default value of `e_plasticpot` ?

JFLemieux73:
I added a note in issue #651 for the new test. I will do that later. Can we please merge this in the main code? I would like to have the plastic potential for next week for the C-grid implementation of rheology.

phil-blain:
@JFLemieux73  I should have noticed that sooner, but maybe it would have been a good idea to keep `e_ratio` in the namelist, for backward compatibility ? meaning, if `e_plasticpot` and `e_yieldcurve` are not in the namelist, but `e_ratio` is, then set `e_plasticpot = e_yieldcruve = e_ratio` ...

I just had a failure with the new code and a namelist from a few weeks ago, that's why I noticed...

JFLemieux73:
I think this complicates things. With time backward compatibility is less and less important so I would not worry about this. Moreover I find that it is a good idea to get rid completely of e_ratio. e_yieldcurve is a better name.

PR #664, opened 2021-11-18
Fix multi-pe advection=none bug

apcraig:
## PR checklist
    Fix multi-pe advection=none bug
- [X] Developer(s): 
    apcraig
    Full test suite run on cheyenne, gnu, everything is bit-for-bit.  Also tested 4x1 box2001 test and that now runs.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#dc1269bcc1970bcc53bd1f95cd88cdac6e06fb68
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Also updated the parser to check for invalid cice.settings env variables and removed an obsolete env setting in set_env.box2001.

Improved script performance in parse_namelist and parse_settings as well as ability to reuse ice_in file in test suites.

Updated test default status from FAIL to PEND.

### DIFF
--- a/cice.setup
+++ b/cice.setup
@@ -1070,6 +1070,10 @@ EOF2
     end
 
     ${casescr}/parse_settings.sh cice.settings ${fsmods}
+    if ($status != 0) then
+      echo "${0}: ERROR, parse_namelist.sh aborted"
+      exit -1
+    endif
     ${casescr}/parse_namelist.sh ice_in ${fimods}
     if ($status != 0) then
       echo "${0}: ERROR, parse_namelist.sh aborted"
--- a/cice.setup
+++ b/cice.setup
@@ -937,7 +937,7 @@ EOF
     if (-e ${fsmods}) rm ${fsmods}
 
     # Use an existing ice_in file from the suite if it exists
-    # to reduce time spend in parse_namelist
+    # to reduce time spent in parse_namelist
     set skip_parse_namelist = spval
     if (${dosuite} == 1) then
       set iceinfn = ../ice_in_save_${grid}${soptions}
@@ -1116,7 +1116,7 @@ EOF2
 #@ pdd = ( $pd1 - $pd0 ) / 1000000
 #echo "tcxp after parse $pdd"
 
-    # Save ice_in in the suite to reduce time spend in parse_namelist
+    # Save ice_in in the suite to reduce time spent in parse_namelist
     if (${dosuite} == 1) then
       if !(-e ${iceinfn}) then
         cp ice_in ${iceinfn}
--- a/cice.setup
+++ b/cice.setup
@@ -1,5 +1,7 @@
 #!/bin/csh -f
 
+#set pd0 = `date -u "+%s%N"`
+
 set ICE_SANDBOX = `pwd`
 set ICE_VERSION = unknown
 if (-e cicecore/version.txt) then
@@ -824,8 +826,8 @@ EOF
     # set default test output as failure
     if (${docase} == 0) then
       echo "#---" >! test_output
-      echo "FAIL ${testname_noid} build" >> test_output
-      echo "FAIL ${testname_noid} run" >> test_output
+      echo "PEND ${testname_noid} build" >> test_output
+      echo "PEND ${testname_noid} run" >> test_output
     endif
 
     # from basic script dir to case
@@ -934,9 +936,21 @@ EOF
     if (-e ${fimods}) rm ${fimods}
     if (-e ${fsmods}) rm ${fsmods}
 
+    # Use an existing ice_in file from the suite if it exists
+    # to reduce time spend in parse_namelist
+    set iceinfn = ../ice_in_save_${grid}${soptions}
+    set skip_parse_namelist = spval
+    if (${dosuite} == 1) then
+      if (-e ${iceinfn}) then
+        echo "use ${iceinfn}"
+        cp ${iceinfn} ice_in
+        set skip_parse_namelist = true
+      endif
+    endif
+
+    # Set decomp info in namelist
     cat >! ${fimods} << EOF1
 # cice.setup settings
-
 nprocs            = ${task}
 nx_global         = ${ICE_DECOMP_NXGLOB}
 ny_global         = ${ICE_DECOMP_NYGLOB}
@@ -965,7 +979,6 @@ EOF1
 
     cat >! ${fsmods} << EOF1
 # cice.setup settings
-
 setenv ICE_SANDBOX  ${ICE_SANDBOX}
 setenv ICE_SCRIPTS  ${ICE_SCRIPTS}
 setenv ICE_CASENAME ${casename}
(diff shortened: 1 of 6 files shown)


cice.setup
```diff
@@ -1176,6 +1224,7 @@ if ( ${dosuite} == 1 ) then
   cat >> ${tsdir}/suite.submit << EOF0
 
 set nonomatch && rm -f ciceexe.* && unset nonomatch
+set nonomatch && rm -f ice_in_save* && unset nonomatch
```
phil-blain, line 1227: why not simply
~~~diff
-set nonomatch && rm -f ciceexe.* && unset nonomatch
+set nonomatch && rm -f ciceexe.* ice_in_save* && unset nonomatch
~~~

configuration/scripts/parse_namelist_from_env.sh
```diff
@@ -5,16 +5,13 @@ if [[ "$#" -ne 1 ]]; then
...
+echo "running $scriptname"
 
-sed -i.sedbak -e 's|ICE_SANDBOX|'"${ICE_SANDBOX}"'|g' $filename
-sed -i.sedbak -e 's|ICE_MACHINE_INPUTDATA|'"${ICE_MACHINE_INPUTDATA}"'|g' $filename
-
-if [[ -e "${filename}.sedbak" ]]; then
-  rm ${filename}.sedbak
-fi
+sed -i -e 's|ICE_SANDBOX|'"${ICE_SANDBOX}"'|g' $filename
+sed -i -e 's|ICE_MACHINE_INPUTDATA|'"${ICE_MACHINE_INPUTDATA}"'|g' $filename
```
phil-blain, line 15: this `.sedbak` extension to the `-i` argument  is needed on macOS to avoid warnings, I'd prefer we keep them (that's why they were added).

configuration/scripts/parse_settings.sh
```diff
@@ -43,18 +46,19 @@ do
...
-    #sed -i 's|\(^\s*set.* '"$vname"' \)[^#]*\(#*.*$\)|\1 '"$value"'  \2|g' $filename
-    sed -i.sedbak -e 's|\(^[[:space:]]*set.* '"$vname"' \)[^#]*\(#*.*$\)|\1 '"$value"'  \2|g' $filename
-
-    if [[ "${found}" == "${foundstring}" ]]; then
-      vnamearray+=($vname)
-      valuearray+=($value)
+    grep -q "^[[:space:]]*set.* ${vname}[[:space:]]*" $filename
+    grepout=$?
+    if [ ${grepout} -eq 0 ]; then
+      sed -i -e 's|\(^[[:space:]]*set.* '"$vname"' \)[^#]*\(#*.*$\)|\1 '"$value"'  \2|g' $filename
```
phil-blain, line 52: same thing here

phil-blain (changes requested):
HI Tony, I had a quick look. I did not look in details at all the script changes.

Regarding the fix in ice_init, I'm not sure I understand the bug (and how it relates to "multi-pe" in the title). From what I could understand, we were only setting `advection='none'` inside an `if (kdyn >=1)` block, so if dynamics were disabled, and `ktransport=-1` in the namelist, advection would still be running since it was not deactivated. Is that the bug you are fixing ?

configuration/scripts/parse_namelist_from_env.sh
apcraig: OK, I'll test on Mac.  The sedbak slows things down a bit which is why I removed it after timing the scripts.  I was trying to speed things up.  If it breaks Mac, then we need to find a work around or use sedbak.

apcraig:
> Regarding the fix in ice_init, I'm not sure I understand the bug (and how it relates to "multi-pe" in the title). From what I could understand, we were only setting `advection='none'` inside an `if (kdyn >=1)` block, so if dynamics were disabled, and `ktransport=-1` in the namelist, advection would still be running since it was not deactivated. Is that the bug you are fixing ?

The advection='none' was set inside an "if (mastertask)" so the logic was only carried out on the root task.  That's why we couldn't run the advection='none' tests on more than 1 task, it was not set consistently across tasks.  This change moves it out of the "if (mastertask)" block and allows that option to work in parallel now.  I've been curious, when the no advection test was added was it setup to run 1x1 because it didn't work in parallel or because there's some usefulness to testing it on 1x1.  I don't have that answer :), but this fixes that issue.  This fix came out of the implementation of more box tests for the CD development.

configuration/scripts/parse_namelist_from_env.sh
apcraig: I played around with sed on cheyenne as well as three flavors of sed on two Macs.  gnu-sed works fine on the Macs if it's installed, which it is on my machines.  The default sed is different.  I tried various things, again, and have decided to just revert the .sedbak implementation.  There are other things we could do, but none are ideal.

apcraig:
Retested everything again and I think it's working fine.  Ready to merge again if others agree.

cice.setup
apcraig: We can go either way.  This happens once at the end of the test suite, so cost is not an issue.  By splitting the lines up, it's easier to comment out one or the other for testing/debugging of the ciceexe or ice_in_save feature.  Two lines seems a little cleaner, but it doesn't make much difference.  I guess I prefer to leave two lines.

eclare108213 (commented):
This looks fine to me but I'm not a scripting expert, so deferring to @phil-blain to approve.

phil-blain:
> > Regarding the fix in ice_init, I'm not sure I understand the bug (and how it relates to "multi-pe" in the title). From what I could understand, we were only setting `advection='none'` inside an `if (kdyn >=1)` block, so if dynamics were disabled, and `ktransport=-1` in the namelist, advection would still be running since it was not deactivated. Is that the bug you are fixing ?
> 
> The advection='none' was set inside an "if (mastertask)" so the logic was only carried out on the root task. 

Ah! Indeed; I did not scroll up enough in my review :P

phil-blain:
@eclare108213 I'm not a C-shell expert either :P but I'll take a second look.

cice.setup
```diff
@@ -934,9 +936,21 @@ EOF
     if (-e ${fimods}) rm ${fimods}
     if (-e ${fsmods}) rm ${fsmods}
 
+    # Use an existing ice_in file from the suite if it exists
+    # to reduce time spend in parse_namelist
```
phil-blain, line 940: s/spend/spent/

cice.setup
```diff
@@ -934,9 +936,21 @@ EOF
...
+    set skip_parse_namelist = spval
+    if (${dosuite} == 1) then
+      set iceinfn = ../ice_in_save_${grid}${soptions}
```
phil-blain, line 943: I think I'd maybe prefer we put the saved namelists in a separate directory, so  as to not clutter the suite's top level directory, but I don't have that strong of an opinion about it.

phil-blain (commented):
OK, I had another look and I guess if it works and is faster, then we're good.

I'm not exactly sure what the changes in `parse_settings.sh` are doing, though.. is that also an optimization ?

cice.setup
apcraig: These sit in the test suite directory and are deleted at the end of the test suite.  You're right that it may be cleaner to put them in a separate directory, but I think we're OK for now.

cice.setup
apcraig: Fixing.

apcraig:
> OK, I had another look and I guess if it works and is faster, then we're good.
> 
> I'm not exactly sure what the changes in `parse_settings.sh` are doing, though.. is that also an optimization ?

The changes reduce the cost of the parsing.  It turns out, these sed commands are taking most of the time setting up the cases.  We are using sed to update cice.settings and ice_in for the case based on the grid and options specified.  This refactoring improves the sed cost a bit and removes an extra sed that was happening in favor of grep to check whether the namelist/env variable exists.  We continue to have some issues with our options setup (including dealing with conflicting or redundant options as well as performance) and I'd love to hear any other ideas how to do this better or different, even moving away from csh :)

apcraig:
Just to expand a bit, the ice_ic_saved files allow a suite to use a pre-parsed ice_in file if it's available.  It still has to change a few things, but many things are reusable.  By copying in an existing pre-parsed ice_in file, we save a lot of the sed calls.  It only helps so much.  For a full test suite on cheyenne, we have about 220 separate cases.  Of those, I think there are about 160 unique ice_in files.  The benefit really comes if you are running suites on multiple envs (compilers) at the same time because those cases are fully redundant, so after creating the 160 unique ice_in files, you can reuse them all when you run the 2nd compiler.

Overall though, the speed up will be useful.  The time for generating a full suite of cases on cheyenne went from about 20 minutes to 11.  The second compiler then only takes 4 minutes if it leverages existing ice_in_saved files.  So running full suites on cheyenne with 3 compilers previously took about an hour (3 x 20) to setup all the cases, now it take more like 20 minutes (11 + 4 + 4).  The compilation and run still takes the same amount of time.  The compilation already had this feature where executables are saved and reused when they can be.

I plan to merge this to cgridDEV hopefully fairly soon.

phil-blain:
Thanks for the details, this is indeed a nice speedup!

apcraig:
@phil-blain, could you give a quick approve if you agree, then I'll merge today for weekend testing.  thanks.

phil-blain (approved):

PR #667, opened 2021-11-25
Makefile: make Fortran object files depend on their dependency files

phil-blain:
When 'make' is invoked on the CICE Makefile, the first thing it does is
to try to make the included dependency files (*.d) (which are in fact
Makefiles themselves) [1], in alphabetical order.

The rule to make the dep files have the dependency generator, 'makdep',
as a prerequisite, so when processing the first dep file, make notices
'makdep' does not exist and proceeds to build it. If for whatever reason
this compilation fails, make will then proceed to the second dep file,
notice that it recently tried and failed to build its dependency
'makdep', give up on the second dep file, proceed to the third, and so
on.

In the end, no dep file is produced. Make then restarts itself and
proceeds to build the code, which of course fails catastrophically
because the Fortran source files are not compiled in the right order
because the dependency files are missing.

To avoid that, add a dependency on the dep file to the rules that make
the object file out of the Fortran source files. Since old-fashioned
suffix rules cannot have their own prerequisites [2], migrate the rules
for the Fortran source files to use pattern rules [3] instead. While at
it, also migrate the rule for the C source files.

With this new dependency, the builds abort early, before trying to
compile the Fortran sources, making it easier to understand what
has gone wrong.

Since we do not use suffix rules anymore, remove the '.SUFFIXES' line
that indicates which extension to use suffix rules for (but keep the
line that eliminates all default suffix rules).

[1] https://www.gnu.org/software/make/manual/html_node/Remaking-Makefiles.html
[2] https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html
[3] https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html#Pattern-Rules

## PR checklist
    title
- [x] Developer(s): 
    me
 tested that everything compiles correctly (full base_suite) and that it fails early if `makdep` can't be compiled.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 

I stumbled on that while working on  the test system for our next HPC migration. This system does not have all locales installed, I was connecting from a system configured in French, and somehow the missing locale made `icc` completely fail to compile `makdep`, and catastrophy ensued.

### DIFF
--- a/configuration/scripts/Makefile
+++ b/configuration/scripts/Makefile
@@ -73,7 +73,6 @@ RM    := rm
 AR    := ar
 
 .SUFFIXES:
-.SUFFIXES: .F90 .F .c .o
 
 .PHONY: all cice libcice targets target db_files db_flags clean realclean helloworld calchk sumchk bcstchk
 all: $(EXEC)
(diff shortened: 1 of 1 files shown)


apcraig (approved):

apcraig:
Thanks @phil-blain, looks like a nice improvement.

PR #670, opened 2021-12-02
NUOPC/CMEPS cap updates

dabail10:

## PR checklist
Some recent updates to the NUOPC/CMEPS cap for CESM.
- [X] Developer(s): 
dabail10 (D.Bailey)
Passes the aux_cice6 testsuite in CESM.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
This has several updates. There is a fix in the PIO module for cases when land block distributions might change across a restart or from an initial file. This basically sets missing values to zero. There is also a fix for the CAM single column mode (SCAM) in CESM. Also, a number of additions for the advanced snow physics on top of what Denise added. I have also removed dragio from the cap initializations. We use the same default in CESM as CICE. I would eventually like to add dragio to the CICE namelist. This is currently in  the Icepack namelist only.


cicecore/drivers/nuopc/cmeps/CICE_InitMod.F90
```diff
@@ -165,6 +162,17 @@ subroutine cice_init2()
...
+    if (tr_snow) then         ! advanced snow physics
+       call icepack_init_snow()
+       call icepack_warnings_flush(nu_diag)
+       if (icepack_warnings_aborted()) call abort_ice(error_message=subname, &
+          file=__FILE__, line=__LINE__)
+       if (snw_aging_table(1:4) /= 'test') then
+          call init_snowtable()
+       endif
+    endif
+
```
DeniseWorthen, line 175: Duplicate block (see LNs 180:189)

cicecore/drivers/nuopc/cmeps/CICE_RunMod.F90
```diff
@@ -201,6 +201,17 @@ subroutine ice_step
...
+         ! snow redistribution and metamorphosis
+         !-----------------------------------------------------------------
+
+         if (tr_snow) then         ! advanced snow physics
+            do iblk = 1, nblocks
+               call step_snow (dt, iblk)
+            enddo
+            call update_state (dt) ! clean up
+         endif
+
```
DeniseWorthen, line 214: This block also is present at LNS 324:329. When should it actually be called? I was following the standalone and dmi placement.

dabail10: Good catch. The merge seemed to cause this. I will look to see where these should be or if it matters.

dabail10:
I think that is all the redundant code in the cap.

cicecore/drivers/nuopc/cmeps/CICE_RunMod.F90
```diff
@@ -90,9 +90,9 @@ subroutine CICE_Run
 
       call ice_step
 
-   !--------------------------------------------------------------------
-   ! end of timestep loop
-   !--------------------------------------------------------------------
+     !--------------------------------------------------------------------
+     ! end of timestep loop
+     !--------------------------------------------------------------------
```
DeniseWorthen, line 95: We could remove the extra spaces, otherwise looks good.

DeniseWorthen (approved):

dabail10:
I ended up cleaning up the indents here, so it should be consistent.

apcraig (commented):
Some of the restart reads zero out filled values?  Is this done consistently across all the different fields?  Why or why not?  Is this done consistently across io_pio2 and io_netcdf?  Should it be?

dabail10:
I don't think it is a problem with netCDF. The problem is that pio writes missing values to the restart where there were missing blocks. There is no way to override this with zeroes. So, the only way to fix it is to check for missing values on read and then zero them out.

apcraig (approved):

apcraig:
@dabail10, this change causes pio1 to fail.  See https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash#4c9024c4b5b78135afe1627d7b9eee12decf471e for failures on cheyenne from weekend testing.  The code doesn't build with pio1 with the following error,

```
             where (work == PIO_FILL_DOUBLE) work = c0
                                          1
Error: Symbol 'pio_fill_double' at (1) has no IMPLICIT type
```

Could you review and propose a new PR to fix this?  You should be able to quickly create a pio1 and pio2 test on cheyenne to verify both are working.  Let me know if you need any help.  If this effort may be delayed, I think we should revert the change in cice io_pio2 and create an issue.  Thanks.

dabail10:
Got it. I had not tested with PIO1.

Dave

dabail10:
Let's revert this change for now.

dabail10:
I am confused. We do not have an io_pio1 directory?

apcraig:
That's true because pio1 and pio2 are nearly identical.  There is a CESM1_PIO CPP that differentiates them.  It may not even be a difference in pio1 and pio2, it's more likely a failure to be backwards compatible in the pio_setframe implementation introduced who-knows-when.  You could probably use the CESM1_PIO CPP for this if you wanted.

apcraig:
If you want to revert, we should create a PR just to undo the few lines in PIO.  We don't want to revert this entire PR.  I think we should see if we can find a proper workaround before using the CESM1_PIO CPP if we're going that way.

dabail10:
I will create a new PR with the CESM1_PIO ifdef.

Dave

PR #675, opened 2021-12-15
Add CESM1_PIO for fill value check

dabail10:

## PR checklist
This accounts for the possibility of PIO1 which does not recognize PIO_FILL_DOUBLE.
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#7a1bd6982e9c8fec985ca3b2cbdd421dffe9492a (PGI had to be run separately)
I also ran aux_cice6 in CESM.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
This fixes an issue with PIO1.

### DIFF
--- a/cicecore/cicedynB/infrastructure/io/io_pio2/ice_restart.F90
+++ b/cicecore/cicedynB/infrastructure/io/io_pio2/ice_restart.F90
@@ -718,7 +718,8 @@
          status = pio_inq_varid(File,trim(vname),vardesc)
 
          if (status /= PIO_noerr) then
-            call abort_ice(subname//"ERROR: CICE restart? Missing variable: "//trim(vname))
+            call abort_ice(subname// &
+               "ERROR: CICE restart? Missing variable: "//trim(vname))
          endif
 
          status = pio_inq_varndims(File, vardesc, ndims)
@@ -728,7 +729,10 @@
 !         if (ndim3 == ncat .and. ncat>1) then
          if (ndim3 == ncat .and. ndims == 3) then
             call pio_read_darray(File, vardesc, iodesc3d_ncat, work, status)
+#ifndef CESM1_PIO
+!           This only works for PIO2
             where (work == PIO_FILL_DOUBLE) work = c0
+#endif
             if (present(field_loc)) then
                do n=1,ndim3
                   call ice_HaloUpdate (work(:,:,n,:), halo_info, &
@@ -738,7 +742,10 @@
 !         elseif (ndim3 == 1) then
          elseif (ndim3 == 1 .and. ndims == 2) then
             call pio_read_darray(File, vardesc, iodesc2d, work, status)
+#ifndef CESM1_PIO
+!           This only works for PIO2
             where (work == PIO_FILL_DOUBLE) work = c0
+#endif
             if (present(field_loc)) then
                call ice_HaloUpdate (work(:,:,1,:), halo_info, &
                                     field_loc, field_type)
--- a/cicecore/cicedynB/infrastructure/io/io_pio2/ice_restart.F90
+++ b/cicecore/cicedynB/infrastructure/io/io_pio2/ice_restart.F90
@@ -729,10 +729,10 @@
 !         if (ndim3 == ncat .and. ncat>1) then
          if (ndim3 == ncat .and. ndims == 3) then
             call pio_read_darray(File, vardesc, iodesc3d_ncat, work, status)
-#ifndef CESM1_PIO
-!           This only works for PIO2
-            where (work == PIO_FILL_DOUBLE) work = c0
-#endif
+!#ifndef CESM1_PIO
+!!           This only works for PIO2
+!            where (work == PIO_FILL_DOUBLE) work = c0
+!#endif
             if (present(field_loc)) then
                do n=1,ndim3
                   call ice_HaloUpdate (work(:,:,n,:), halo_info, &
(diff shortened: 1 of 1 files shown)


apcraig:
Did you test the io options?  The base suite doesn't test pio at all.  You can either do it manually or run the io_suite which tests the netcdf, pio1, pio2, and binary options fairly comprehensively (only works on cheyenne because of pio requirement).  Thanks.

dabail10:
Sorry. I misunderstood. I thought you said the tests on cheyenne were failing and assumed the base_suite covered this. I will rerun with the io_suite.

Dave

apcraig:
Only the pio1 tests were failing and those are carried out weekly with the io_suite.  Thanks!

apcraig:
Any chance we could finish testing and merge before tomorrow (Fri, Dec 17) for weekend testing.  I will also try to find a few minutes to fire off some tests as backup.

dabail10:
I can definitely get this in. Am I using this flag incorrectly? Does CESM1_PIO mean PIO2 or PIO1? I would like this to be more clearly named.

Dave

apcraig:
I think the usage is correct.  When CESM1_PIO is turned on, it's assuming it's compiled with pio from CESM1.

dabail10:
So, why does this code not work?

#ifndef CESM1_PIO
!           This only works for PIO2
            where (work == PIO_FILL_DOUBLE) work = c0
#endif

apcraig:
In what way does it not work?  Are you wondering if the where statement should work in PIO1 or are you asking why the ifdef still turns on the where statement with PIO1?

dabail10:
The code does not compile because it cannot find PIO_FILL_DOUBLE. However, if CESM1_PIO is not defined, the code should be removed? I checked the pio1 cases and CESM1_PIO is not defined anywhere.

apcraig:
Good point.  I think I understand now.  It looks like CESM1_PIO gets turned on by RASM (and should by other CESM1 versions when using this version of CICE and an older version of PIO).  The pio1 we're testing on cheyenne is newer, so it supports the newer implementation of pio than what is needed for RASM.  Sorry about my confusion.  The CESM1_PIO is never turned on in CICE standalone testing on cheyenne, just in RASM testing.   That's why it's not a PIO1 vs PIO2 thing.  It's the slow evolution of PIO in ways that make it difficult to support backwards compatibility.  So, it seems this new PIO_FILL_DOUBLE is also something that is newer than the pio1 we're testing on cheyenne which means it also will need to be turned off by CESM1_PIO.  I guess, unfortunately, we need a new cpp for this in the standalone testing. 

So the options are to continue to introduce CPPs that support various versions
- CESM1_PIO for RASM and older versions of pio1 that use the older version of pio_setframe
- ?? for pio1 testing in standalone on cheyenne to avoid the PIO_FILL_DOUBLE
- no CPP for newest version of pio (which we may not even be testing to be honest).

Or we could just try to implement this without using PIO_FILL_DOUBLE.  For instance, could we have control of the fill value in PIO directly in CICE?  That way CICE could set it to a special value, pass it to PIO, and then check for that special value at read-time.  That might work in all versions of PIO too?

dabail10:
Got it. The other issue that I do not have an answer to is Jim is calling the latest PIO2, and all previous versions are an iteration of PIO1. So, the code that we call PIO2 is actually PIO1. I wondered if we could just replace PIO_FILL_DOUBLE with spval_dbl. Let me think about it overnight and I'll look into a solution for tomorrow.

apcraig:
Thanks @dabail10, let me know if I can help.

dabail10:
So, the only issues with PIO1 are the setframe calls and PIO_FILL_DOUBLE. So, I would like to introduce a PIO1 flag and use this instead of CESM1_PIO. Would this work for RASM?

dabail10:
I am also fine with CESM1_PIO and just making sure that we turn this on for our pio1 tests.

dabail10:
So, this now works for PIO1, but it is breaking PIO2. The problem is when it tries to read iceumask (line 405):

/glade/work/dbailey/CICE_ESCOMP/cicecore/cicedynB/infrastructure/ice_restart_driver.F90

It is aborting in ice_restart.F90 at the where statement, line 747. The only difference for iceumask is that it reads into work1 as a temporary. However, it is not happy with the check for PIO_FILL_DOUBLE. It works for the rest of the 2D fields though. Any thoughts?

apcraig:
I don't see any reason iceumask would obviously fail.  There doesn't seem to be anything special about it.

Regarding use of the CPPs.  The problem is we have 3 cases, so we need at least 2 CPPs.  The three cases are

- RASM (old pio1), use old setframe, no PIO_DOUBLE_FILL
- pio1.10.1 (pio1 in standalone), use new setframe, no PIO_DOUBLE_FILL
- pio2.4.4 (pio2 in standalone), use new setframe, can use PIO_DOUBLE_FILL

I wonder if at this point, it would make sense to remove the PIO_DOUBLE_FILL mods for now and check that in today?  Then we can continue to explore how to deal with this best?  I still sort of like the idea of having CICE set the fill/special value explicitly if possible and not use PIO_DOUBLE_FILL at all?  That may not work in PIO.  It could be that the fill value is not settable or it could be that the fill value for parts of the array that are not passed into PIO are not settable.  We'd have to explore that.

dabail10:
I'm fine with that. Gives me some more time to sort this out.

dabail10:
I found this declaration in PIO2:

double precision, public, parameter :: PIO_FILL_DOUBLE = 9.9692099683868690d+36

Weird that it is not the same as spval_dbl which is 1.0d30. I could do something like:

where (work >= spval_dbl) work = c0

I was worried that checking if work was identically equal to PIO_FILL_DOUBLE that it might cause problems if roundoff different.

Dave

dabail10:
I found the problem with iceumask. It was because work1 was not initialized before it was sent to the read_restart_field subroutine. I assume the where statement choked because the ghost cells had nothing in them. I am rerunning the iosuite.

apcraig:
I have run the quick suite and io suite on cheyenne with 3 compilers, everything is fine and bit-for-bit with the version from 2 weekends ago.  I will merge now.

apcraig (approved):

dabail10:
Fixed in #799

PR #671, opened 2021-12-02
Main namelist debug

daveh150:
For detailed information about submitting Pull Requests (PRs) to the CICE-Consortium,
please refer to: <https://github.com/CICE-Consortium/About-Us/wiki/Resource-Index#information-for-developers>

## PR checklist
- [ ] Short (1 sentence) summary of your PR: 
    Provide detailed output when namelist has wrong variables
- [ ] Developer(s): 
    David Hebert
- [ ] Suggest PR reviewers from list in the column to the right.
- [ ] Please copy the PR test results link or provide a summary of testing completed below.
Quick suite BFB test on GAFFNEY shows:
21 measured results of 21 total results
21 of 21 tests PASSED
0 of 21 tests PENDING
0 of 21 tests MISSING data
0 of 21 tests FAILED

- How much do the PR code changes differ from the unmodified code? 
    - [ x] bit for bit
    - [ ] different at roundoff level
    - [ ] more substantial 
- Does this PR create or have dependencies on Icepack or any other models?
    - [ x] Yes
    - [ ] No
- Does this PR add any new test cases?
    - [ ] Yes
    - [x ] No
- Is the documentation being updated? ("Documentation" includes information on the wiki or in the .rst files from doc/source/, which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/. A test build of the technical docs will be performed as part of the PR testing.)
    - [ ] Yes
    - [x ] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [ x] No 
- [ x] Please provide any additional information or relevant details below:
If you have a namelist variable incorrect (say CICE was updated with different namelist variable name), the error just tells you which namelist. This enhancement adds method to print the actual line in namelist. 

Example: CICE6 changed 'sst_data_type' to 'ocn_data_type'. If you didn't change it new code will now print:

 (abort_ice)ABORTED:
 (abort_ice) called from ice_init.F90
 (abort_ice) line number          579
 (abort_ice) error =
 (input_data)ERROR: reading namelist    sst_data_type   = 'gofs_daily'

PR #668, opened 2021-11-29
nuopc/cmeps driver updates

DeniseWorthen:
For detailed information about submitting Pull Requests (PRs) to the CICE-Consortium,
please refer to: <https://github.com/CICE-Consortium/About-Us/wiki/Resource-Index#information-for-developers>

## PR checklist
- [x] Short (1 sentence) summary of your PR: 
    Updates nuopc/cmeps driver for advanced snow physics and debug_model,debug_model_step options. 
    Adds the ability to use ESMF memory profiling optionally.
    Sets a background level for ``rhoa`` over land (UWM only).

- [x] Developer(s): 
    @DeniseWorthen 
- [X] Suggest PR reviewers from list in the column to the right.
    @apcraig 
    @dabail10 
- [X] Please copy the PR test results link or provide a summary of testing completed below.
I tested this in the aux_cice6 CESM testsuite and everything is bfb.    
- How much do the PR code changes differ from the unmodified code? 
    - [x] bit for bit
    - [ ] different at roundoff level
    - [ ] more substantial 
- Does this PR create or have dependencies on Icepack or any other models?
    - [ ] Yes
    - [x] No
- Does this PR add any new test cases?
    - [ ] Yes
    - [x] No
- Is the documentation being updated? ("Documentation" includes information on the wiki or in the .rst files from doc/source/, which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/. A test build of the technical docs will be performed as part of the PR testing.)
    - [ ] Yes
    - [x] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [x] No 
- [x] Please provide any additional information or relevant details below:

-Both ``debug_model`` and ``debug_model_step`` options were exercised and seen to produce the expected output.
-A test case was run setting ``tr_snow`` to true. The model successfully completed but the results were not examined in detail. 
-Driver updates were committed to ufs-weather-model in PR [#910](https://github.com/ufs-community/ufs-weather-model/pull/910). Baselines changed in that PR due solely to the update of a library required by the UPP (unified post processor) submodule on two platforms.

PR #677, opened 2021-12-21
Update CICE namelist implementation

### LINKED ISSUE #676: Namelist implementation issues
Recent changes to the CICE namelist implementation to trap bad namelist inputs is failing with the NAG compiler.  After some debugging, this comes down to a few issues

- The new feature leverages trapping bad lines with the iostat flag in the read namelist command.  But that iostat flag is also used to cycle thru the namelist file to find the correct group in some cases.
- Most compilers seem to reopen the namelist file in the place where it was most recently closed.  NAG does not do that, it rewinds the file.
- As currently coded, there is an assumption (mostly in ice_init) that namelist groups occur in a particular order.  That is not documented.  In other parts of the code, the namelist group could be in an arbitrary order.  The required order seems to be more-or-less fortuitously followed by users and is implemented in just the perfect order in the standalone code.

The new feature that was added to trap bad namelist lines happens to work on most compilers because

- The namelist groups appear in just the right order in the standalone testing to not require cycling in the file.
- Most compilers do not rewind the namelist file when reopening, so there is no cycling of the file to find the right group.  
- The iostat command happens to only be in use to trap bad lines in this case.

But NAG rewinds the file, so it's trapping the "line" error during normal cycling.  I think there are three possible options to move things forward,

1. Keep the current implementation which has some (but not all) required order of the namelist groups and document that.
2. Generalize the implementation, so there is no required order.  That would mean we'd have to rewind the namelist file anytime we open it and we'd have to cycle thru the groups (as we do now for some of the groups) until we find the right one for all namelist groups.
3. Formalize the order of all namelist groups.  To do this robustly, we'd want to open the namelist file once and leave it open.  The groups would have to always be in a particular order from start to end.  We would document this.

Only option 3 will allow us to trap and write a bad line in the namelist group.  With option 1 or 2, we'd need to use the iostat to cycle thru the groups and we couldn't trap an error on a single line of the group...  unless we check the iostat value and are able to robustly determine that certain values mean certain things and that it's uniform across all compilers.  I don't think the iostat values are consistent across compilers.

Maybe there are other options.  How should we proceed in terms of options 1, 2, or 3 at this point?  We probably want to fix this soon-ish, even if that only means reverting back to what we had before (option 1) since the code is broken on NAG at the moment.

@dabail10, @eclare108213 @daveh150 @phil-blain

apcraig:
## PR checklist
    Update CICE namelist implementation
- [X] Developer(s): 
    apcraig
    Full test suites on izumi and cheyenne, all OK.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#33a3c2bd4107a9d566078bdea3a6a40e8903f3ef
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

See #676 for more details.  This is option 2 on that issue, generalize the implementation so namelist groups can be read in any order.  That was tested on NAG and it works as expected there.  Closes #676 if accepted.

- Update the namelist read to make the group order flexible.
- Remove recent update to namelist read that traps bad lines, it conflicts
  with flexibility to read groups in random order picked up by NAG.
- change print* statements to write(nu_diag,*)

### DIFF
--- a/cicecore/cicedynB/analysis/ice_diagnostics.F90
+++ b/cicecore/cicedynB/analysis/ice_diagnostics.F90
@@ -1706,7 +1706,7 @@
 ! dynamics (transport and/or ridging) causes the floe size distribution to become non-normal
 !         if (tr_fsd) then
 !         if (abs(sum(trcrn(i,j,nt_fsd:nt_fsd+nfsd-1,n,iblk))-c1) > puny) &
-!            print*,'afsdn not normal', &
+!            write(nu_diag,*) 'afsdn not normal', &
 !                 sum(trcrn(i,j,nt_fsd:nt_fsd+nfsd-1,n,iblk)), &
 !                     trcrn(i,j,nt_fsd:nt_fsd+nfsd-1,n,iblk)
 !         endif
--- a/cicecore/cicedynB/analysis/ice_history.F90
+++ b/cicecore/cicedynB/analysis/ice_history.F90
@@ -93,8 +93,7 @@
       integer (kind=int_kind), dimension(max_nstrm) :: &
          ntmp
       integer (kind=int_kind) :: nml_error ! namelist i/o error flag
-      character(len=char_len_long) :: tmpstr2 ! for namelist errors
-      
+
       character(len=*), parameter :: subname = '(init_hist)'
 
       !-----------------------------------------------------------------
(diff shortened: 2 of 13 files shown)


apcraig:
I decided to code up option 2 in #676.  If we decide to go a different direction with namelist, we can close this PR and I'm happy to recode as needed.

eclare108213 (commented):
This looks fine to me, as long as it works as intended.  Some of those print statements have been in there forever!  I always have trouble finding the compare-source-codes functionality in github - how do I compare this PR with the code prior to the namelist PR #671?

phil-blain:
>  how do I compare this PR with the code prior to the namelist PR #671?

You can construct the URL yourself, as long as you know the commit hashes:
~~~
$ git log --oneline upstream/main -5
dfc6a113 (upstream/main, upstream/HEAD) Add CESM1_PIO for fill value check (#675)
4c9024c4 NUOPC/CMEPS cap updates (#670)
ddc1cfcc Main namelist debug (#671)
2b851261 (fix-output-conversion-errors) nuopc/cmeps driver updates (#668)
162aee96 Fix multi-pe advection=none bug (#664)
~~~

OK, the commit hash of the squash-merge for #671 is ddc1cfcc and the  parent of ddc1cfcc is 2b851261. Now, the commit hash of the tip of Tony's branch for this here PR is https://github.com/CICE-Consortium/CICE/pull/677/commits/33a3c2bd4107a9d566078bdea3a6a40e8903f3ef (see above in the PR page).

So the compare URL is https://github.com/CICE-Consortium/CICE/compare/2b851261..33a3c2bd4107a9d566078bdea3a6a40e8903f3ef (it works with either full or abbreviated hashes). Note that that view will also show changes made to master in the meantime, i.e. 4c9024c4 and dfc6a113.

Hope that helps:)

eclare108213:
Awesome.  Amazing.  Etc.  
Thanks!

eclare108213 (approved):

dabail10 (approved):
Looks good. It is really only the namelists in ice_init.F90 that are the problem. The others are in different modules and ice_in gets rewound to the top. Nonetheless it is good to be consistent.

apcraig:
@dabail10, that's correct.  In fact, I only use rewind when we read a namelist and didn't just open the file.  The namelists in ice_init were the ones that had to be in a certain order before (undocumented).  The others already worked fine in any order.

PR #727, opened 2022-06-02
Update cice.t-test.py to use cartopy instead of basemap.

daveh150:

## PR checklist
Update cice.t-test.py to use cartopy instead of basemap.

- [x] Developer(s): 
    David Hebert
     
    - [ x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x ] No
    - [ ] Yes
    - [ x] No
    - [ x] Yes
        - [ ] Yes
        - [ x] No 

### DIFF
--- a/configuration/scripts/tests/QC/cice.t-test.py
+++ b/configuration/scripts/tests/QC/cice.t-test.py
@@ -546,12 +546,15 @@ def plot_data(data, lat, lon, units, case, plot_type):
     #plt.suptitle('CICE Mean Ice Thickness\n{}'.format(case), y=0.95)
     plt.suptitle(f'CICE Mean Ice Thickness\n{case:s}')
 
-    # adjust subplot up. 
+    # add more whitespace between plots for colorbar. 
     plt.subplots_adjust(wspace=0.4)
 
+    # add separate axes for colorbars
+    # first get position/size of current axes
     pos1 = ax1.get_position()
     pos2 = ax2.get_position()
 
+    # now add new colormap axes using the position ax1, ax2 as reference
     cax1 = fig.add_axes([pos1.x0+pos1.width+0.03,
                          pos1.y0,
                          0.02,
@@ -563,15 +566,12 @@ def plot_data(data, lat, lon, units, case, plot_type):
                          pos2.height])
 
 
-    #cbar_ax = fig.add_axes([0.11,0.1,0.8,0.05])
     if '\n- ' in case:
         # If making a difference plot, use scientific notation for colorbar
-        #cbNH = plt.colorbar(scNH, cax=ax1, orientation="horizontal", format="%.2e")
-        #cbSH = plt.colorbar(scSH, cax=ax2, orientation="horizontal", format="%.2e")
         cbNH = plt.colorbar(scNH, cax=cax1, orientation="vertical", 
-                            pad=0.1, shrink=0.6, format="%.1e")
+                            pad=0.1, format="%.1e")
         cbSH = plt.colorbar(scSH, cax=cax2, orientation="vertical", 
-                            pad=0.1, shrink=0.6, format="%.1e")
+                            pad=0.1, format="%.1e")
 
     else:
         #pass
--- a/configuration/scripts/tests/QC/cice.t-test.py
+++ b/configuration/scripts/tests/QC/cice.t-test.py
@@ -394,7 +394,7 @@ def plot_data(data, lat, lon, units, case, plot_type):
     spstereo = ccrs.SouthPolarStereo(central_longitude= 90.0) # define projection
    
     # define figure
-    fig = plt.figure(figsize=[14,8])  
+    fig = plt.figure(figsize=[14,7])  
 
     # add axis for each hemishpere
     ax1 = fig.add_subplot(121,projection=npstereo)
@@ -424,31 +424,61 @@ def plot_data(data, lat, lon, units, case, plot_type):
                        draw_labels=True,
                        x_inline=False,y_inline=False)
 
-    g1 = ax2.gridlines(xlocs=mpLons,ylocs=mpLats,
+    g2 = ax2.gridlines(xlocs=mpLons,ylocs=mpLats,
                        draw_labels=True,
                        x_inline=False,y_inline=False)
 
 
+    # Specify Min/max colors for each hemisphere
+    # check for minus to see if it is a difference plot
+    if '\n- ' in case: # this is a difference plot
+        # specify colormap
+        mycmap = 'seismic' # blue,white,red with white centered colormap
+
+        # determine max absolute value to use for color range
+        # intent is use same min/max with center zero
+        dmin = np.abs(data.min())
+        dmax = np.abs(data.max())
+        clim = np.max([dmin,dmax])
+
+        # this specifies both hemishperes the same range.
+        cminNH = -clim
+        cmaxNH =  clim
+        cminSH = -clim
+        cmaxSH =  clim
+
+    else:  # not a difference plot
+        # specify colormap
+        mycmap = 'jet'
+
+        # arbitrary limits for each Hemishpere
+        cminNH = 0.0
+        cmaxNH = 5.0
+        cminSH = 0.0
+        cmaxSH = 2.0
+
     if plot_type == 'scatter':
         # plot NH
-        sc = ax1.scatter(lon,lat,c=data,cmap='jet',s=4,edgecolors='none',
-                         transform=ccrs.PlateCarree())
+        scNH = ax1.scatter(lon,lat,c=data,cmap=mycmap,s=4,edgecolors='none',
+                           vmin=cminNH, vmax=cmaxNH,
+                           transform=ccrs.PlateCarree())
         
         # plot SH
-        sc = ax2.scatter(lon,lat,c=data,cmap='jet',s=4,edgecolors='none',
-                        transform=ccrs.PlateCarree())
+        scSH = ax2.scatter(lon,lat,c=data,cmap=mycmap,s=4,edgecolors='none',
+                           vmin=cminSH, vmax=cmaxSH,
+                           transform=ccrs.PlateCarree())
 
     else: 
         if plot_type == 'contour':
             print("contour plot depreciated. using pcolor.")
             
-        sc = ax1.pcolormesh(lon,lat,data,cmap='jet',
-                            vmin=data.min(), vmax=data.max(),
-                            transform=ccrs.PlateCarree())
+        scNH = ax1.pcolormesh(lon,lat,data,cmap=mycmap,
+                              vmin=cminNH, vmax=cmaxNH,
+                              transform=ccrs.PlateCarree())
         
-        sc = ax2.pcolormesh(lon,lat,data,cmap='jet',
-                            vmin=data.min(), vmax=data.max(),
-                            transform=ccrs.PlateCarree())
+        scSH = ax2.pcolormesh(lon,lat,data,cmap=mycmap,
+                              vmin=cminSH, vmax=cmaxSH,
+                              transform=ccrs.PlateCarree())
 
     #else:
         # # Create new arrays to add 1 additional longitude value to prevent a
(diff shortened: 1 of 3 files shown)


eclare108213:
@daveh150
These are fairly extensive changes -- it took some work to do this.  Thanks!  Do you have sample output, before and/or at least after?

For future reference (and you can edit your text above) - deleting the space from the checkbox code when you add an x will make them work...   [x] instead of [ x] etc.

daveh150:
Thank you. The contour plot section was the challenging part, needing both +/- 180 and 0-360 lat/lons.

I admittedly did not test the plot in this exact script, I do not have a perturbation run to test with (I likely missed how to set that up on the testing documentation).  I was testing plots outside of this script using output from one the 'smoke' tests.  If anyone has suggestion on how to make a perturbation test I can test this.

daveh150:
Oh, I think I see why I couldn't run a QC test before. Previously I got a message that my test and baseline were bit-for-bit, so no qc test was required. But I did not make any changes to the ice_in, so the runs were identical. I'll try to change the shortwave and ktherm in the 'test' run and see if I can run the QC test to get example plots from this script. Sorry about that.

daveh150:
Does anyone use the contour option in these plots? That is by far the most challenging, the scatter and pcolor seem to do what we want. I'd like to just remove the contour plots if they aren't used.

eclare108213:
I didn't know that contour plotting was even available.  As far as I'm concerned, color maps should be sufficient and you can comment the contouring part out.  (Am I correctly understanding the issue here?) Check the docs to see if it's mentioned, and if so then comment that out too.  Thx

apcraig:
This looks fine to me, I'll try to run a t-test on cheyenne and make sure everything is working to independently confirm.  I've struggled at times getting the python environment setup, has anything changed in that realm?

apcraig:
Thanks @daveh150, those plots look pretty good overall.  No idea about the missing data in the contour plots.

daveh150:
I made an update where if contour is selected, the plots will revert to pcolor. Eventually that whole section of the code could be removed once we are sure no one wants to use it. It cleans up the plotting code significantly. I did not see a need to add the cyclic arrays with the pcolormesh.

dabail10:
Just got a chance to look at this. This is great! Thanks @daveh150 for doing this. Here are the plots from my gx3 B vs C grid comparison. A couple of notes for enhancements later.

1. This does seem to be in the right sense of the differences I would expect. That is the B grid solution is thinner.
2. However, It would be nice if the colorbar could be reset a bit to bring out the differences. The "red" area is basically zero everywhere and so a red-white-blue colormap might be nice for the difference plot. Then instead of having the maximum distance as the colorbar, maybe a one-sigma difference is better. That is computing the spatial variance of the distances and using that as the max on the colorbar instead of the absolute max difference. Perhaps even two-sigma. This would help bring out the smaller differences better.
3. Also, it would be nice if the colorbar for the two cases was the same. This way one could see the differences pretty clearly in these plots.

ice_thickness_gx3testb
ice_thickness_gx3testc
ice_thickness_gx3testb_minus_gx3testc

dabail10 (approved):
This looks great. I added some additional comments for enhancements.

dabail10:
> I made an update where if contour is selected, the plots will revert to pcolor. Eventually that whole section of the code could be removed once we are sure no one wants to use it. It cleans up the plotting code significantly. I did not see a need to add the cyclic arrays with the pcolormesh.

I agree that it is probably not worth chasing down the issues with the contour plots. Typically we don't see issues like this except on tripolar grids. It also looks like the line of missing values goes along the 90E/W line. It would make more sense if it was related to 180/0. I recall that Matt had designed it so that one could choose in case the pcolor doesn't work for some reason. What happens if you don't use filled contours?

apcraig:
@dabail10, thanks for reviewing and testing the scripts.  @daveh150, Are we ready to merge?  It seems like the outstanding requests may be

- update environment.yml, add cartopy, remove basemap
- update the colorbar

I think those are two good suggestions.  If we prefer to defer, I'm also OK merging.

apcraig:
@daveh150, just checking in again to see how we should proceed.

apcraig:
Thanks @daveh150.  @dabail10, do you have some QC data you can run the updated scripts on to verify they work and look good?  If not, I will try to generate some data and do the same.  thanks!

dabail10:
I do have some runs I can test this on.

dabail10:
These are great! It is great to see such a large area of white where the differences are close to zero. We might even be able to set the min/max on the difference plots to even smaller, but this is great for now.

ice_thickness_gx1testb_minus_gx1testc
ice_thickness_gx1testc
ice_thickness_gx1testb

daveh150:
@dabail10 , interesting, those plots are not what I was expecting. Do you have the latest version of the script? I set it up to have colorbars and colormaps like the attached. Maybe I didn't commit something correctly?

ice_thickness_nrlssc_gnu_smoke_gx3_20x1_qc qc_base

ice_thickness_nrlssc_gnu_smoke_gx3_20x1_qc qc_te

ice_thickness_nrlssc_gnu_smoke_gx3_20x1_qc qc_base_minus_nrlssc_gnu_smoke_gx3_20x1_qc qc_te

dabail10:
I downloaded from the PR tag (daveh150:update_plots).

daveh150:
Hmm, that still looks like the basemap version. I'll email you what I committed so you can make sure it is the right version.

dabail10:
There are two sets of plots here. The earlier ones were the old script.

phil-blain:
That might be something on m side, but with the script in this PR I get plots like this;

image


I haven't yet looked into why it's all blue...

configuration/scripts/tests/QC/cice.t-test.py
```diff
@@ -389,87 +389,200 @@ def plot_data(data, lat, lon, units, case, plot_type):
...
+        scSH = ax2.scatter(lon,lat,c=data,cmap=mycmap,s=4,edgecolors='none',
+                           vmin=cminSH, vmax=cmaxSH,
+                           transform=ccrs.PlateCarree())
 
+    else: 
         if plot_type == 'contour':
-            sc = m.contourf(x, y, d1, cmap='jet')
-        else:  # pcolor
-            sc = m.pcolor(x, y, d1, cmap='jet')
+            print("contour plot depreciated. using pcolor.")
```
phil-blain, line 473:
```suggestion
            print("contour plot deprecated. using pcolor.")
```

dabail10:
This all looks good to me. Not sure why Phillipe is getting blue everywhere. Is the colormap slightly different on your machine?

phil-blain:
This is with Cartopy 0.20.2, Matpltlib 3.5.2. What versions are your using ?

dabail10:
Cartopy 0.20.2 matplotlib 3.5.1

phil-blain:
I can investigate on my own as to not delay merging this.

phil-blain:
my plots are on the gx1 grid, maybe that's why I'm seeing something different?

dabail10:
Just ran it on one of my gx1 tests and I don't see the blue.

ice_thickness_gx1testc
ice_thickness_gx1testb_minus_gx1testc
ice_thickness_gx1testb

phil-blain:
In my plot I've just noticed there is also a dark longitudinal line near 30°E...

phil-blain:
I've spent more time debugging this. I *think* it's due to the fact that the `t_lat` and `t_lon` arrays are masked arrays due to land proc elimination:

`t_lat`:
image
`t_lon`:
image

If I do something like [`t_lat=t_lat.filled(0)`](https://numpy.org/doc/stable/reference/maskedarray.generic.html#filling-in-the-missing-data), and the same for `t_lon`, i.e. filling the masked values with zeros:
image

Then I get the expected plot:

image

So @dabail10 I guess my next question would be what was your MPI x OpenMP configuration for the gx1 runs you used to test the new script ? Would it be possible that there was no land procs?

phil-blain:
So I zoomed out a bit in my plot above and I don't think it's as easy as filling the lat/lon arrays with zeros:

image

we get spurious polygons, with different shapes depending on the filling value. I think we should use "complete" i.e. non masked `tlat`, `tlon`. This leads to the correct plot:

image


So one way to do it would be for the script to read the location of the input data directory (`CICE_data`) from the `env.machine_compiler` file  and then  read the grid from there to get the full lat/lon arrays, The only "problem" is that we currently do not have the gx1 grid as a NetCDF file.

daveh150:
@phil-blain @dabail10 if the issue is related to masked locations in the tlon/tlat but NOT the data, could you try this to get the mask the same. Then I think the pcolor will ignore the masked locations:

data.mask = np.logical_or(data.mask,lon.mask)

eclare108213:
In my opinion, tlat and tlon ought to be made 'complete' (no eliminated land blocks) in the output files.  One way to do this would be to write the original, global arrays, rather than writing them after the decomposition when the blocks are eliminated.  Not sure how easy this would be, or if it would work for PIO etc.

daveh150:
Thanks @phil-blain. I think you had an idea to read the gx1 grid file instead of from the history file, if it was netcdf? That should be straight forward, we just need to write the TGRID in addition the UGRID.

daveh150:
I made a grid_gx1.nc file. Hopefully it is attached, can y'all verify it is okay then perhaps add it to the CICE_data grid files for future use?
[grid_gx1.nc.gz](https://github.com/CICE-Consortium/CICE/files/9023675/grid_gx1.nc.gz)

Updating the cice.t-test.py to find this file will take a bit of thought, since it is not in the run directory.

daveh150:
Sorry for quick modification of the gx1 netcdf, I changed the variables to be all lowercase in accordance with popgrid_nc. Also changed the kmt data type to integer and unit to 'none'.

[grid_gx1.nc.gz](https://github.com/CICE-Consortium/CICE/files/9023726/grid_gx1.nc.gz)

phil-blain:
I implemented what we talked about in the previous meeting i.e. disabling land blocks elimination. 

@daveh150, you could update your branch to include [my commit](https://github.com/daveh150/CICE/compare/update_plots...phil-blain:CICE:update_plots) thusly:

```bash
git checkout update_plots
git pull git@github.com:phil-blain/CICE.git update_plots
git push
```

apcraig:
Thanks for doing that @phil-blain, I think I volunteered but then dropped the ball.  @daveh150, could you pull @phil-blain's change to your sandbox and retest?  @phil-blain, have you confirmed that this fixes the issues you were seeing?  If so, and everyone agrees, we can problem merge soon.  Thanks!

phil-blain:
I had already verified that using the full TLON/TLAT arrays made the script work correctly. I had done that by creating a gx1 test  with  `-p 1x1`, which then changed the block size and resulted in a single block and thus no land blocks elimination. I then manually loaded these TLON/TLAT arrays in Python and used those to create the "correct" maps in https://github.com/CICE-Consortium/CICE/pull/727#issuecomment-1170080421.

I just verified now that  with this updated `qc` option, i.e. including  ` distribution_wght = 'blockall'`, creating a test with `-p 44x1 -s qc` , i.e. [what the doc suggests using](https://cice-consortium-cice.readthedocs.io/en/master/user_guide/ug_testing.html#end-to-end-testing-procedure), the land blocks are indeed not eliminated and the TLON/TLAT arrays in the output files do not have any missing values. So I think this shows the fix is sufficient :)

apcraig:
Thanks @phil-blain, it makes sense the blockall would fix the problem.  Thanks for confirming.

@daveh150, if there are any mods related to pulling in a separate lon/lat file for plotting, those can be removed again.

@daveh150, @phil-blain's mod is a one line change.  If you prefer to implement it manually rather than pull/push, that would be fine too.  You can see the mod here, https://github.com/daveh150/CICE/compare/update_plots...phil-blain:CICE:update_plots, as noted by @phil-blain.

apcraig:
See #742

daveh150:
I've been on leave since last week. I did not have any other changes for the lat/lon issue, glad that the blockall solution by @phil-blain worked! Thank you.

Issue #599, opened 2021-04-22
More use of uninitialized arrays / variables (running 'base_suite' with `-init=snan,arrays`)


phil-blain:
Following #579, and to my request, the machine files for `daley` were updated to use the signalling NaN initialization for debug builds. 

I ran the 'base_suite' on `master` yesterday for the first time since then and noticed several more failures in `debug` cases due to the code using arrays or variable before they are initialized. @apcraig I know you did not activate the flag on other machines so I guess you must be aware of some of these issues, but I spent a little time documenting the failures I got and so I thought it would be worth it to list them here.

So here goes my findings:

### daley_intel_restart_gx3_6x2_alt01_debug_short
~~~
 (calc_timesteps) modified npt from        10 d with dt=       3600.00
 (calc_timesteps)                to       240 1 with dt=       3600.00
 (calc_timesteps) start time is  2005-01-01:00000
 (calc_timesteps)   end time is  2005-01-11:00000

forrtl: error (65): floating invalid
Image              PC                Routine            Line        Source
cice               00000000022DF5E4  Unknown               Unknown  Unknown
cice               0000000001B73480  Unknown               Unknown  Unknown
cice               00000000012B5FEF  ice_init_column_m         433  ice_init_column.F90
cice               0000000000403821  cice_initmod_mp_c         220  CICE_InitMod.F90
cice               0000000000401EB3  cice_initmod_mp_c          52  CICE_InitMod.F90
cice               000000000040168B  MAIN__                     43  CICE.F90
cice               00000000004015F2  Unknown               Unknown  Unknown
cice               00000000023C014F  Unknown               Unknown  Unknown
cice               00000000004014DA  Unknown               Unknown  Unknown
~~~

**array `apeffn` is used but is not initialized if `shortwave= 'ccsm3'`**

### daley_intel_restart_gx3_8x2_alt02_debug_short
~~~
 (calc_timesteps)   end time is  2005-01-11:00000

  Initial forcing data year =         2005
  Final   forcing data year =         2005

 Atmospheric data files:
 /home/ords/cmdd/cmde/sice500//CICE_data/forcing/gx3/JRA55/8XDAILY/JRA55_gx3_03hr_forcing_2005.nc
  Set current forcing data year =         2005
 (JRA55_data) reading forcing file 1st ts = /home/ords/cmdd/cmde/sice500//CICE_data/forcing/gx3/JRA55/8XDAILY/JRA55_gx3_03hr_forcing_2005.nc
[...]

PR #749, opened 2022-08-08
ice_grid: do call 'gridbox_verts' for rectangular grids

phil-blain:
## PR checklist
    title
- [X] Developer(s): 
    me
    base suite is bfb with CICE6.4.0, as expected:
```
349 of 349 tests PASSED
0 of 349 tests PENDING
0 of 349 tests MISSING data
0 of 349 tests FAILED
```

I also set up a case on  a rectangular grid (`gbox80`) and compiled with `-init=snan,arrays` and verified that the code does not crash as it did before (see "Problems in ice_grid.F90" in https://github.com/CICE-Consortium/CICE/issues/599#issue-865459873).

    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

At the end of subroutine ice_grid::gridbox_corners, the arrays
'lont_bounds' and 'lonu_bounds', which contain the longitude of the
corners of each grid cell on the T and U grids, are converted to to the
[0, 360] range.

In the case of rectangular grids ('grid_type = rectangular'), at the
point where 'gridbox_corners' is called in 'init_grid2', 'lont_bounds'
is not initialized, causing the code to abort if compiling with NaN
initialization. This is due to the fact that 'gridbox_verts', which
initializes 'lont_bounds' and 'latt_bounds', is not called in
'rectgrid', whereas it is called in 'popgrid[_nc]'.

Do call 'gridbox_verts' in 'rectgrid', so that 'lont_bounds' and
'latt_bounds' are correctly initalized in that case also.

Note that these calls are also missing in 'latlongrid' and 'cpomgrid',
but since these two subroutines are not used in standalone
configuration, let's not bother for now.

### DIFF
--- a/cicecore/cicedynB/infrastructure/ice_grid.F90
+++ b/cicecore/cicedynB/infrastructure/ice_grid.F90
@@ -1404,6 +1404,7 @@
          enddo
          work_g1(:,:) = work_g1(:,:) / rad_to_deg
       endif
+      call gridbox_verts(work_g1,lont_bounds)
       call scatter_global(ULON, work_g1, master_task, distrb_info, &
                           field_loc_NEcorner, field_type_scalar)
       call ice_HaloExtrapolate(ULON, distrb_info, &
@@ -1423,6 +1424,7 @@
          enddo
          work_g1(:,:) = work_g1(:,:) / rad_to_deg
       endif
+      call gridbox_verts(work_g1,latt_bounds)
       call scatter_global(ULAT, work_g1, master_task, distrb_info, &
                           field_loc_NEcorner, field_type_scalar)
       call ice_HaloExtrapolate(ULAT, distrb_info, &


apcraig (approved):

apcraig (approved):

phil-blain:
Maybe @dabail10 you would like to comment on the fact that these are also missing in  `latlongrid` ? I can update that subroutine also if you wish.

Or, we could refactor a little bit more, and have `gridbox_verts` be called in  `init_grid2`, instead of by each individual grid reading routine (thinking about it, that would be cleaner). But I don't know if it's mandatory for `gridbox_verts` to be called *before* [the call to `ice_HaloExtrapolate`](https://github.com/CICE-Consortium/CICE/pull/749/files#diff-5c8e815b29d7bb8283f9ed89427d8271562e374797b1a57fe8174825ff736b2fR1410)... If it does not matter then it would be safe to move the call to `gridbox_verts` to `init_grid2`, I would think.

dabail10:
Not sure what to say here. This subroutine latlongrid was created for the atmosphere grids and prescribed ice mode. I thought we were reading in all the information instead of computing it. However, since dynamics is off in this configuration, a lot of the B-grid velocity information is not needed. We have actually moved away from using latlongrid  now. We talked about removing it, but I think some groups still use it?

> Maybe @dabail10 you would like to comment on the fact that these are also missing in `latlongrid` ? I can update that subroutine also if you wish.
> 
> Or, we could refactor a little bit more, and have `gridbox_verts` be called in `init_grid2`, instead of by each individual grid reading routine (thinking about it, that would be cleaner). But I don't know if it's mandatory for `gridbox_verts` to be called _before_ [the call to `ice_HaloExtrapolate`](https://github.com/CICE-Consortium/CICE/pull/749/files#diff-5c8e815b29d7bb8283f9ed89427d8271562e374797b1a57fe8174825ff736b2fR1410)... If it does not matter then it would be safe to move the call to `gridbox_verts` to `init_grid2`, I would think.

phil-blain:
OK, thanks. I pinged you because it's in between `ifdef CESMCOUPLED`. If that the `latlongrid` subroutine is not used or in the process of being removed, we can just let it be, I would say.

PR #752, opened 2022-08-17
ice_calendar: zero-initialize 'nstreams'

phil-blain:
## PR checklist
 title 
- [X] Developer(s): 
me
    I checked that the code runs fine with the fix when compiling with `-init=huge`. It crashes without the fix.
I ran the `base`, `decomp` and `nothread` suites with the fix and  `-init=huge` and everything passed as expected.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

The variable ice_calendar::nstreams, which corresponds to the number of
output history streams to use for the run, is initialized in
ice_history::init_hist depending on the number of non-'x' elements in
'histfreq' in the namelist.

However, the code does use 'nstreams' before ice_history::init_hist is
called, in ice_calendar::calendar when called from
ice_calendar::init_calendar. Both 'init_calendar' and 'init_hist' are
called from CICE_InitMod::cice_init, in that order, such that the loop
that initializes 'write_history' in 'calendar' uses 'nstreams'
uninitialized.

'calendar' ends up being called at least once more during 'cice_init', from
ice_calendar::advance_timestep, at which point 'nstreams' is correctly
defined and 'write_history' is thus correctly initialized, before its
first use in 'accum_hist'.

To avoid using 'nstreams' uninitialized in the first call to 'calendar'
from 'init_calendar', initialize it to zero in 'init_calendar' before
calling 'calendar'.

This issue was discovered by compiling using the '-init=huge' flag of
the Intel compiler.

### DIFF
--- a/cicecore/shared/ice_calendar.F90
+++ b/cicecore/shared/ice_calendar.F90
@@ -204,6 +204,10 @@
       dt_dyn = dt/real(ndtd,kind=dbl_kind) ! dynamics et al timestep
       force_restart_now = .false.
 
+      ! initialize nstreams to zero (will be initialized from namelist in 'init_hist')
+      ! this avoids using it uninitialzed in 'calendar' below
+      nstreams = 0
+
 #ifdef CESMCOUPLED
       ! calendar_type set by coupling
 #else


dabail10 (approved):
Good catch.

apcraig (approved):

Issue #660, opened 2021-11-16
C-grid Code Cleanup 


apcraig:
key:
**_needs to be done in cice-consortium/main before cgrid fork_** 
_likely changes answers on B grid_

**style and consistency**
- [ ] variable names TUNE consistency
       - caps for TUNE (i.e. Tbu)
       - do we want to add U or T to existing variables?
       - do we want to change tmask to maskT?  tarea to areaT?  etc
- [x] there is a mix of `if (grid_ice == ‘CD’) then` and `select case (trim(grid_ice))`.  Do this consistently.  
- [x] rename strocn[x,y]T because it's different units from other strocn[x,y] terms (divides by ai for coupling output).  Better yet, move this to different location associate with output.
- [x] comment the code to explain that 00, 10, 01, 11 are i+0,j+0,i+1 and j+1
- [x] set capping variable based on namelist (or other, i.e. kdyn), make name more descriptive
- [ ] join iblk loops in ice_grid.F90 (and review loop structure everywhere), careful about answer changes
- [x] remove unused grid options in history file (i.e. ustr3Dz)
- [ ] white space, alignments, comments, etc
- [x] iceumask, icenmask, iceemask are logicals in ice_flux.F90.  icetmask is integer.  Make consistent?
- [x] reneame cicedynB to cicedyn and create soft link

**bugs**
- [x] uocn, vocn is on the T grid and never averaged to U grid (except for hadgem forcing)
- [x] #734
- [x] **`atm_data_dir = '/glade/u/home/tcraig/cice_data/'` in ice_in, needs to be 'unknown'**
- [x] #735
- [x] #736
- [x] #737
- [x] #707 

**better approaches**
- [x] _consolidate deltamin variables and update value of deltamin_
- [x] #738
- [ ] update dynamics driver, (see #619) 
- [ ] _new computation of dx and dy leveraging HTS, HTW_ (see #653)
- [ ] change grid_type = 'setmask' to use kmt_type instead (NUOPC / CMEPS) (see #805)
- [x] #739
- [x] move pre-transport shift of (uvel,vvel) from U to N,E points into a subroutine that can be called from the ends of eap, vp, evp
- [x] clean up geometry/forcing/ic settings primarily related to box configurations.  it's currently a function of ice_in, grid_type, atm/ocn/ice_data_type, kmt_type, boundary_type, close_boundaries.  Want to be able to define the geometry, mask, initial condition, and forcing separately and cleanly.
[...]

PR #762, opened 2022-09-22
Update grid averaging for tmass, aice, uvelT, vvelT

### LINKED ISSUE #735: change ice_step_therm1 to compute uvel, vvel at cell centers using new averaging routine

### LINKED ISSUE #736: change to_ugrid calculation of mass, aice to use T2US instead of T2UF (check and fix everywhere, eventually)

apcraig:
## PR checklist
    Update grid averaging for tmass, aice, uvelT, vvelT
- [X] Developer(s): 
    apcraig
    Everything runs, changes most results, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#658799554aafd24f9c2fd575b68b47b281b03027.  QC tests pass relative to current main (cice.tr502 is current main, #6587995
```
PASS cheyenne_intel_qcchk_gx1_144x1_medium_qc_qcchk compare cice.tr502 -1 -1 -1
PASS cheyenne_pgi_qcchk_gx1_144x1_medium_qc_qcchk compare cice.tr502 -1 -1 -1
PASS cheyenne_gnu_qcchk_gx1_144x1_medium_qc_qcchk compare cice.tr502 -1 -1 -1
```
    - [ ] bit for bit
    - [X] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [ ] No 

- Update tmass and aice T2U mapping, switch from "F" to "S", F was backwards compatible but not correct (changes answers), closes #736 
- Update ocean forcing T2U averaging in ocn_data_ncar_init, change "F" to "A".
- Update uvelT, vvelT averaging in step_therm1, change from 4 point average to U2TA (changes answers for highfreq=.true.), closes #735
- Remove history grids not needed (i.e. ustr3Dz), see #660

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_evp.F90
@@ -317,22 +317,22 @@
       ! convert fields from T to U grid
       !-----------------------------------------------------------------
 
-      call grid_average_X2Y('F', tmass    , 'T'          , umass   , 'U')
-      call grid_average_X2Y('F', aice_init, 'T'          , aiU     , 'U')
+      call grid_average_X2Y('S', tmass    , 'T'          , umass   , 'U')
+      call grid_average_X2Y('S', aice_init, 'T'          , aiU     , 'U')
       call grid_average_X2Y('S', uocn     , grid_ocn_dynu, uocnU   , 'U')
       call grid_average_X2Y('S', vocn     , grid_ocn_dynv, vocnU   , 'U')
       call grid_average_X2Y('S', ss_tltx  , grid_ocn_dynu, ss_tltxU, 'U')
       call grid_average_X2Y('S', ss_tlty  , grid_ocn_dynv, ss_tltyU, 'U')
 
       if (grid_ice == 'CD' .or. grid_ice == 'C') then
-         call grid_average_X2Y('F', tmass    , 'T'          , emass   , 'E')
-         call grid_average_X2Y('F', aice_init, 'T'          , aie     , 'E')
+         call grid_average_X2Y('S', tmass    , 'T'          , emass   , 'E')
+         call grid_average_X2Y('S', aice_init, 'T'          , aie     , 'E')
          call grid_average_X2Y('S', uocn     , grid_ocn_dynu, uocnE   , 'E')
          call grid_average_X2Y('S', vocn     , grid_ocn_dynv, vocnE   , 'E')
          call grid_average_X2Y('S', ss_tltx  , grid_ocn_dynu, ss_tltxE, 'E')
          call grid_average_X2Y('S', ss_tlty  , grid_ocn_dynv, ss_tltyE, 'E')
-         call grid_average_X2Y('F', tmass    , 'T'          , nmass   , 'N')
-         call grid_average_X2Y('F', aice_init, 'T'          , ain     , 'N')
+         call grid_average_X2Y('S', tmass    , 'T'          , nmass   , 'N')
+         call grid_average_X2Y('S', aice_init, 'T'          , ain     , 'N')
          call grid_average_X2Y('S', uocn     , grid_ocn_dynu, uocnN   , 'N')
          call grid_average_X2Y('S', vocn     , grid_ocn_dynv, vocnN   , 'N')
          call grid_average_X2Y('S', ss_tltx  , grid_ocn_dynu, ss_tltxN, 'N')
--- a/cicecore/cicedynB/analysis/ice_history_shared.F90
+++ b/cicecore/cicedynB/analysis/ice_history_shared.F90
@@ -156,57 +156,25 @@
          igrdz(nvar_grdz)      ! true if category/vertical grid field is written
 
       character (len=25), public, parameter :: &
-         tcstr = 'area: tarea'          , & ! vcellmeas for T cell quantities
-         ucstr = 'area: uarea'          , & ! vcellmeas for U cell quantities
-         ncstr = 'area: narea'          , & ! vcellmeas for N cell quantities
-         ecstr = 'area: earea'          , & ! vcellmeas for E cell quantities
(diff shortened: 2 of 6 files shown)


cicecore/cicedynB/dynamics/ice_dyn_evp.F90
```diff
@@ -317,22 +317,22 @@ subroutine evp (dt)
       ! convert fields from T to U grid
       !-----------------------------------------------------------------
 
-      call grid_average_X2Y('F', tmass    , 'T'          , umass   , 'U')
-      call grid_average_X2Y('F', aice_init, 'T'          , aiU     , 'U')
+      call grid_average_X2Y('S', tmass    , 'T'          , umass   , 'U')
+      call grid_average_X2Y('S', aice_init, 'T'          , aiU     , 'U')
```
eclare108213, line 321: The change here for tmass is right: to get the correct mass at the U point, we need to compute 
`sum(rhoi*vice*aice*cell area)/sum(aice*cell area)`
over the T cells [ignoring snow for now], since vice has units of volume per unit area.

But I do not think this is correct for aice_init.  Since ice concentration is assumed to be uniformly distributed over each grid cell, the concentration at U should be
`sum(aice_init*cell area)/sum(cell area)`
but I think the calculation here is 
`sum(aice_init*aice*cell area)/sum(aice*cell area)`.

apcraig: Just to clarify, The "S" mapping is masked area weighted while the "A" mapping is just area weighted.  When you use "aice" above, is that the ice fraction?  There are no ice fraction weighted averages built into CICE.  If we want to do that, we can use user defined averaging weights.  Again,

```
"S":  sum(var*Tmask*Tarea)/sum(Tmask*Tarea)
"A":  sum(var*Tarea)/sum(Tarea)
"F":  sum(var*Tarea_var)/Tarea_dst
```

Where Tarea and Tmask are the static grid cell areas and masks computed at initialization.  If we do want to have fraction weighted mapping, we can do that, but it's not the standard "S", "A", or "F".  To do that we'd do

```
mask_s(:,:,;) = 1
wght_s(:,:,:) = Tarea(:,:,:) * aice(:,:,:)
call grid_average_X2Y('A', tmass, 'T', wght_s, mask_s, umass, 'U')
```

and that would do

`umass = sum(tmass*Tarea*aice)/sum(Tarea*aice)
`

At this point, we NEVER weigh the grid averaging by the ice fraction anywhere.  Is that something we should be doing?

eclare108213: Sorry, I misunderstood "area weighted" in the doc to mean ice-area-weighted, but it clearly says grid cell area.  My mistake.  Let me think about this some more - what I wrote above is not quite correct.

eclare108213: Correction to my comment above: 

To get the correct mass at the U point, we need to compute 
`sum(rhoi*vice*Tarea)/sum(Tarea)` 
over the T cells [ignoring snow for now], since vice has units of volume per unit area and already incorporates aice (i.e. vice = hi * aice).  So the tmass line is correct.

Since ice concentration is assumed to be uniformly distributed over each grid cell, the initial concentration at U should be 
`sum(aice_init*Tarea)/sum(Tarea)` 
and that is what's being done in the S mapping. 

So I agree with both of these modifications.  Sorry for the confusion!

eclare108213 (approved):

cicecore/cicedynB/dynamics/ice_dyn_evp.F90
apcraig: The "S" mapping includes the mask, the "A" mapping does not.  I think we need to change "S" to "A" for umass and aiU if we want an unmasked weighted average as written above.  This only matters for gridcells near land.  So should the average be an average of only ocean cells ("S" mapping) or an average of all ocean/land cells ("A" mapping)?

One other thing that I just want reiterate.  Maybe there is a place for fraction weighted mapping in CICE.  We do that in the coupler when we map ice coupling fields to other grids.  We fraction area weight the mapping so a gridcell with fraction=0.9 has much greater weight than a gridcell with fraction=0.2 when interpolating from ice to atmosphere for example.  You can also think of it as a fraction weighted merge.  How we do that depends whether it's a flux (which has to be strictly conserved and has a particular mapping requirement) or a state (which could be done many ways). Maybe we need to do the same in CICE in some places for some fields??  I guess it partly depends whether we can define an ice fraction on the grid we're mapping from.  We'd have to think about it a bit.

eclare108213: I think S is correct in this case, with the mask included, although it would be good for @JFLemieux73 @dabail10 to confirm that it's consistent with other assumptions in the C-grid approach.  On a B-grid, the velocity being computed would be zero if any of the T cells is land (i.e. the point is moot).

apcraig: Great, thanks for the clarification.

eclare108213: I also wondered whether ice area weighting could/should be included as part of the averaging infrastructure.  For coupling, the actual averaging happens in the coupler with other components' data, so we can't control the entire average anyhow. Once the model starts being coupled with different grids, we should check whether we are calculating the coupling fields appropriately, but I think you've already thought through all of that.  In the ice model, we "aggregate" across ice thickness categories, but that's only within a particular cell, not for shifting things around horizontally (there are vertical grid remappings). The tracer loading/unloading also involves a lot of multiplying/dividing by "weights" (including ice area and other tracers).  Neither of these cases involve grid shifts.  So: I don't think we should worry about it unless we come across a particular case in which we need it.

PR #763, opened 2022-09-29
ice_dyn_vp: use `global_sum_*` for bit-for-bit reproducibility

phil-blain:
I'm opening this PR as draft mainly so we can discuss the performance regression I'm hitting, hence I did not fill the whole PR template. I'll do that when we are ready to merge the code (if we do).

Summary:
I refactored the VP solver code so that global sums are computed using the existing CICE `global_sum_*` subroutines, so that bit-for-bit reproducibility can be achieved if desired using the existing `bfbflag` namelist settting.

My adventures in doing so are detailed at:
- https://github.com/phil-blain/CICE/issues/39#issuecomment-1125463055 and following
- https://github.com/phil-blain/CICE/issues/40#issuecomment-1184774930 and following

As you can see from the changes in the PR, each call to `global_sum` is replaced by two calls to `global_sum_prod`, one for the array holding X components and a second for the array holding Y components.

I used the  Intel VTune profiler to profile both the old and new code, compared them, and got these results: https://github.com/phil-blain/CICE/issues/40#issuecomment-1188260610.

Seeing that the time spent in the MPI_ALLREDUCE in the new code was twice that of the old code, I tried to refactor the code again to do the global sums for both dimensions at the same time: 
- code: https://github.com/phil-blain/CICE/compare/repro-vp-no-pairs...repro-vp
- results: https://github.com/phil-blain/CICE/issues/40#issuecomment-1261514528

but that did not really help, which was disappointing... 

So I guess I'd like to have some input from others :)

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_vp.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_vp.F90
@@ -2780,6 +2780,7 @@
       real (kind=dbl_kind) :: &
          norm_residual   , & ! current L^2 norm of residual vector
          inverse_norm    , & ! inverse of the norm of a vector
+         norm_rhs        , & ! L^2 norm of right-hand-side vector
          nu, t               ! local temporary values
 
       integer (kind=int_kind) :: &
@@ -2818,6 +2819,25 @@
       workspace_y = c0
       arnoldi_basis_x = c0
       arnoldi_basis_y = c0
+
+      ! solution is zero if RHS is zero
+      !$OMP PARALLEL DO PRIVATE(iblk)
+      do iblk = 1, nblocks
+         call calc_L2norm_squared(nx_block       , ny_block      , &
+                                  icellu   (iblk),                 &
+                                  indxui (:,iblk), indxuj(:,iblk), &
+                                  bx(:,:,iblk)   , &
+                                  by(:,:,iblk)   , &
+                                  norm_squared(iblk))
+
+      enddo
+      !$OMP END PARALLEL DO
+      norm_rhs = sqrt(global_sum(sum(norm_squared), distrb_info))
+      if (norm_rhs == c0) then
+         solx = bx
+         soly = by
+         return
+      endif
 
       ! Residual of the initial iterate
 
--- a/cicecore/cicedynB/dynamics/ice_dyn_vp.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_vp.F90
@@ -783,9 +783,6 @@
          stress_Pr,   & ! x,y-derivatives of the replacement pressure
          diag_rheo      ! contributions of the rhelogy term to the diagonal
 
-      real (kind=dbl_kind), dimension (max_blocks) :: &
-         L2norm       ! array used to compute l^2 norm of grid function
-
       real (kind=dbl_kind), dimension (ntot) :: &
          res        , & ! current residual
          res_old    , & ! previous residual
@@ -821,7 +818,6 @@
 
       ! Initialization
       res_num = 0
-      L2norm  = c0
       Fx = c0
       Fy = c0
 
@@ -901,8 +897,7 @@
                                indxui     (:,iblk), indxuj    (:,iblk), &
                                bx       (:,:,iblk), by      (:,:,iblk), &
                                Au       (:,:,iblk), Av      (:,:,iblk), &
-                               Fx       (:,:,iblk), Fy      (:,:,iblk), &
-                               L2norm       (iblk))
+                               Fx       (:,:,iblk), Fy      (:,:,iblk))
          enddo
          !$OMP END PARALLEL DO
          nlres_norm = sqrt(global_sum_prod(Fx(:,:,:), Fx(:,:,:), distrb_info, field_loc_NEcorner) + &
@@ -1968,8 +1963,7 @@
                                indxui     , indxuj  , &
                                bx         , by      , &
                                Au         , Av      , &
-                               Fx         , Fy      , &
-                               sum_squared)
+                               Fx         , Fy        )
 
       integer (kind=int_kind), intent(in) :: &
          nx_block, ny_block, & ! block dimensions
@@ -1989,9 +1983,6 @@
          Fx      , & ! x residual vector, Fx = bx - Au (N/m^2)
          Fy          ! y residual vector, Fy = by - Av (N/m^2)
 
-      real (kind=dbl_kind), intent(out), optional :: &
-         sum_squared ! sum of squared residual vector components
-
       ! local variables
 
       integer (kind=int_kind) :: &
@@ -2000,12 +1991,8 @@
       character(len=*), parameter :: subname = '(residual_vec)'
 
       !-----------------------------------------------------------------
-      ! compute residual and sum its squared components
+      ! compute residual
       !-----------------------------------------------------------------
-
-      if (present(sum_squared)) then
-         sum_squared = c0
-      endif
 
       do ij = 1, icellu
          i = indxui(ij)
@@ -2013,9 +2000,6 @@
 
          Fx(i,j) = bx(i,j) - Au(i,j)
          Fy(i,j) = by(i,j) - Av(i,j)
-         if (present(sum_squared)) then
-            sum_squared = sum_squared + Fx(i,j)**2 + Fy(i,j)**2
-         endif
       enddo                     ! ij
 
       end subroutine residual_vec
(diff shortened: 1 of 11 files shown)


apcraig:
I had a quick look at your notes and implementation.  First,

I think you can reduce the new implementation from two MPI sums to one.  You should be able to compute "Fx*Fx+Fy*Fy" locally and put it in a temporary array (nx,ny,iblk).  Then call global_sum on that array.  I hope I understand that correctly.  And you can do that over icells saving some mult-adds.  That should save the chaos (masks, tripoles, etc) in global_sum_prod as well as reduce the MPI_Allreduce calls in half.  And that should still work on tripole because all the tripole logic for the global sums is in global_sum.

Separately, it might be worth recoding global_sum and global_sum_prod so the masking is outside the loops.  I suspect those methods could perform better with some refactoring, but we'd have to explore that.   That may be a second step to come later.

I think an important difference in the original and new implementation is that the old implementation was using global_sum_scalar for the reduction after computing a local sum.   Circling back, part of the difference is probably the cost of the local sum which does not seem particularly efficient for arrays in the global_sum method because of how the loops and if tests are structured.

apcraig:
One other comment, is it possible the time accounted in MPI is partly due to load imbalance elsewhere such that the AllReduce is waiting for all pes to arrive?

apcraig:
Just for fun, I ran a few cases with this branch and 

`./cice.setup -m cheyenne -e intel -g gx1 -p 72x1 -s dynpicard --test smoke --testid rv01
`

Here are the dynamics timings I got,

```
Timer   3:  Dynamics       5.45 seconds    ! old version of the code (current CICE main)
Timer   3:  Dynamics      10.85 seconds    ! out of the box (bfbflag='off')
Timer   3:  Dynamics       9.72 seconds    ! out of the box + simpler global_sum_prod_dble
Timer   3:  Dynamics      23.07 seconds    ! bfbflag='reprosum'
Timer   3:  Dynamics      22.79 seconds    ! bfbflag='reprosum'  + simpler global_sum_prod_dble
```

So, getting the if statements out of global_sum_prod_dble (I just commented them out for testing purposes), reduces the dynamics time by 10% for bfbflag='off'.  It has less impact with bfbflag='reprosum'.  We should recode the global_sum and global_sum_prod at some point.  Also 'reprosum' is quite a bit more expensive overall, but it doesn't have to be on in production, just for some test cases.  The next thing I might try to do is combine the two global_sum_prod calls into a single global_sum call in ice_dyn_vp and see the effect.  

I think a main point is that global reductions have never been evaluated for performance because they were just used for diagnostics and never were a bottleneck.  It's good we're looking at them now.

apcraig:
A few more timing data points for this version of the code,

```
Timer   3:  Dynamics       5.45 seconds    ! old version of the code (current CICE main)
Timer   3:  Dynamics      10.85 seconds    ! out of the box (bfbflag='off')
Timer   3:  Dynamics       9.72 seconds    ! bfbflag='off' + simpler global_sum_prod_dbl
Timer   3:  Dynamics      23.07 seconds    ! bfbflag='reprosum'
Timer   3:  Dynamics      22.79 seconds    ! bfbflag='reprosum'  + simpler global_sum_prod_dbl
Timer   3:  Dynamics       8.38 seconds    ! bfbflag='off' + 2->1 VP AllReduce
Timer   3:  Dynamics       7.24 seconds    ! bfbflag='off' + 2->1 VP AllReduce + simpler global_sum_dbl
```

So we were at 5.45s with old VP implementation, with 2->1 VP Allreduce calls and simpler global_sum (removed if checks in loops), it's now 7.24s.  I don't know if those differences could be explained by other changes in the VP implementation, but it suggests if we can refactor the two global_sum_prod calls to a single global_sum and improve the performance of the global_sum, we're starting to get close.  Also, the temporary arrays passed into global_sum are computed at all points, they could be masked saving still more time.  We still have to pay for some overhead in global_reductions, copying the array and then the call to compute_sums_dbl.  And maybe there are other things that could be done.

Separately, those intel performance tools looks cool.  Maybe we should use them to help us focus on other issues like vector performance.  Historically, those tools have required a bunch of spinup time and they didn't work well on real MPI applications (even though they work great on sample codes), but maybe CICE is a small enough application and the tools are getting better that it'd be very useful to go thru that exercise.

phil-blain:
Thanks a lot for taking a look.

> I had a quick look at your notes and implementation. First,
> 
> I think you can reduce the new implementation from two MPI sums to one. You should be able to compute "Fx_Fx+Fy_Fy" locally and put it in a temporary array (nx,ny,iblk). Then call global_sum on that array. I hope I understand that correctly. And you can do that over icells saving some mult-adds. That should save the chaos (masks, tripoles, etc) in global_sum_prod as well as reduce the MPI_Allreduce calls in half. And that should still work on tripole because all the tripole logic for the global sums is in global_sum.

Yes, I think that's right. 

> 
> Separately, it might be worth recoding global_sum and global_sum_prod so the masking is outside the loops. I suspect those methods could perform better with some refactoring, but we'd have to explore that. That may be a second step to come later.
> 

Yes, I was suspecting that also. I haven't yet confirmed it. 

> I think an important difference in the original and new implementation is that the old implementation was using global_sum_scalar for the reduction after computing a local sum. Circling back, part of the difference is probably the cost of the local sum which does not seem particularly efficient for arrays in the global_sum method because of how the loops and if tests are structured.

Yes, indeed. Another thing is that the computation of the local sum that was1 OpenMP-accelerated, but this is not the case for `global_sum`, which makes sense as we want to be able to get the same answer with different thread counts.

phil-blain:
I'll make some more tweaks and test on my side with a single AllReduce instead of two, and see if I can tweak `global_sum` to make use of the existing `icellu` array to sum only on ice points...

Thanks for your input!

apcraig:
Hi @phil-blain, thanks for continuing to push on this.  

I don't know if it'll make sense to push the icellu mask down into the global sum.  That might require some pretty significant changes in those methods all the way down to ice_repro_sum.  My suggestion is to use the icellu mask first just to reduce the local Fx*Fx+Fy*Fy computation but to leave the array passed into global_sum the regular size without passing icellu.  Using icellu as a "mask" or "lmask" in the global_sum is also just going to slow it down.  Those masks just zero out the cells, they don't reduce the size of the array to sum over.

It seems to me the first thing is to convert the 2 global_sum_prods into 1 global_sum by computing Fx*Fx+Fy*Fy efficiently outside the global_sum.  The second step is to refactor the global_sum methods to pull the if checks outside the loops.  That second step could be done as a separate PR over the next few weeks if you prefer, I think we've identified the problem there and know we need to fix it for VP at least.  Otherwise, it might be reasonable to merge this PR soon?  I think it's unlikely we'll be able to duplicate the baseline performance.

However, let me propose another option.  That would be to use the original approach when we don't need bit-for-bit but use the standard global_sum (of the array) when we need bit-for-bit reproducible.  That will ensure the production runs can go as fast as possible while still retaining ability to do all the checks we want in a bit-for-bit mode.  That would probably require a "vp_global_sum" subroutine in VP to leverage code reuse.

phil-blain:
> I don't know if it'll make sense to push the icellu mask down into the global sum. That might require some pretty significant changes in those methods all the way down to ice_repro_sum. My suggestion is to use the icellu mask first just to reduce the local Fx_Fx+Fy_Fy computation but to leave the array passed into global_sum the regular size without passing icellu. Using icellu as a "mask" or "lmask" in the global_sum is also just going to slow it down. Those masks just zero out the cells, they don't reduce the size of the array to sum over.

I was not thinking I would have to pass `icellu`  down to ice_reprosum. My idea was that since a lot of blocks have only ocean and no ice, we are summing a lot of zeros in `global_sum_dbl` before handling the work to `compute_sums_dbl`.  If the loop is done on `icellu`, then we do really skip the ocean cells, we are not only masking them. So I have a feeling it could be faster. I'd like to try it just to measure the difference. The array that would be passed to `compute_sums_dbl` would stay the same shape, so no further changes down the call stack would be necessary.

> 
> It seems to me the first thing is to convert the 2 global_sum_prods into 1 global_sum by computing Fx_Fx+Fy_Fy efficiently outside the global_sum. The second step is to refactor the global_sum methods to pull the if checks outside the loops. That second step could be done as a separate PR over the next few weeks if you prefer, I think we've identified the problem there and know we need to fix it for VP at least. Otherwise, it might be reasonable to merge this PR soon? I think it's unlikely we'll be able to duplicate the baseline performance.

Yes, if we do refactor the `global_sum`, I agree it could be in a subsequent PR. And then we can merge this one soon-ish, once we are confident we do not degrade the timings too much.

> 
> However, let me propose another option. That would be to use the original approach when we don't need bit-for-bit but use the standard global_sum (of the array) when we need bit-for-bit reproducible. That will ensure the production runs can go as fast as possible while still retaining ability to do all the checks we want in a bit-for-bit mode. That would probably require a "vp_global_sum" subroutine in VP to leverage code reuse.

This is exactly what @JFLemieux73 suggested when we talked  last week, and then he suggested reaching out to you to get a second opinion.  So I think this is an option worth exploring, too. I think if I refactor things a bit, we can keep the code clean and still retain the same performance, or very close (we would have an additional `if` somewhere to dispatch appropriately to the original or new code depending on `bfbflag`, but apart from that the performance should be pretty similar).

apcraig:
That all sounds great.  I think global_sum is just doing a copy into a 1d array so I'm not sure icellu would help much.  It might help if we were still using global_sum_prod.  I think we're also on the same page with respect to supporting two global sum approaches within vp.  As I noted before, the global sum has not been on the critical path for performance nor has it played a role in results beyond just diagnostics.  Having a fast custom global sum where it is on the critical path is a totally reasonable approach.

phil-blain:
> That all sounds great. I think global_sum is just doing a copy into a 1d array so I'm not sure icellu would help much. It might help if we were still using global_sum_prod. 

Oh, you are totally right, this is what I had in mind indeed. If we simply use `global_sum` and not `global_sum_prod`. then it does not make sense to use `icellu`. 

One thing that I thought of to speed up `global_sum` and friends is to just add more subroutines to the module procedure. We could have one version without any mask, one with the multiplicative mask, one with the logical mask, etc. Then the dispatching would be done at compile time, which is even better. And we could gather the common code (dealing with the tripole blocks) in a separate subroutine and simply call it from all global_sum_* subroutines.

apcraig:
Very good ideas for refactoring global_sum routines.  My thought was just pull out the if blocks and to do the masking separately only when needed, but having more interfaces might work even better, just have to see how much duplication we end up with and how much code reuse we can have.

phil-blain:
Final PR is here: https://github.com/CICE-Consortium/CICE/pull/774

Issue #67, opened 2018-01-06
track the coupling fluxes separately, to avoid confusion before/after scale_fluxes


eclare108213:
Some (but not all) fluxes are divided by aice before being sent to the cesm coupler.  This is highly confusing because normally, multiplying a non-flux value (e.g. thickness) specific to the ice-covered area by aice (or similarly area-averaging over categories) produces a grid-cell-mean value.  The subroutine scale_fluxes divides by aice, so multiplying by aice then just brings it back to the ice-covered area value.  It would be better to save the coupling fluxes separately so that the physical interpretations of the primary variables aren't changing.  E.g. from a user's question on how to compute net longwave from the history output:

flwdn in history is flw elsewhere in the code, and that is the value for any point in the grid cell, whether it’s ice or ocean.  It’s not multiplied by aice, but it is still a grid-cell average because it’s the same over ice and ocean alike.

flwup in history is flwout elsewhere in the code, and in the code calculations this is the value only over sea ice. However it’s later divided by aice for the cesm coupler, and that happens before it’s sent to history.  flwup_ai in history is flwout*aice,  so it’s back to the only-over-ice value. 

So the net longwave is 

over ice:  flwnet = flwup_ai-flwdn
over the grid cell:  flwnet(ice) + flwnet(ocn) = (flwup_ai-flwdn)*aice + (sigma*SST^4 - flwdn)*(1-aice)

Is that right?

czender:
That looks right to me. If I understand logic then `flwup_ai` in history is `(flwout/aice)*aice`. A minor point is that floating point arithmetic does not guarantee that `(flwout/aice)*aice == flwout`. Thus splitting `flwout` into two versions, one for the coupler and the other for internal use and history would improve the BFBness of history archived in double precision (default in MPAS?).

dabail10:
I think we should have a conversation about this. It has just come up again in the context of the ponds. So, the pond fraction is:

apond = sum(apondn(n)*aicen(n)), n = 1, ncat

However, there is a history variable apond_ai. That is, we take apond*aice. However, do we ever divide apond by aice in the code? If not, that means apond_ai has aice multiplied twice effectively. We have a couple variables like this that are not sent to the coupler and hence are not scaled.

dabail10:
Note that I am using apondn as the short form for trcrn(:,:,nt_apnd,:,:) here.

czender:
Pinging @zswolff

dabail10:
[...]

PR #764, opened 2022-10-05
Refactor strocnxT, strocnyT implementation

### LINKED ISSUE #761: Check strocnxT implementation
I'm looking at the implementation of strocnxT, strocnyT in CICE.  This is defined as the ice/ocn stress on the T grid and is something that is passed out of the model for coupling.   It is computed by taking the strocnxU, strocnyU, normalizing the ice fraction on the U grid (aiU) and then averaging to the T grid.  Normalization by ice fraction is needed for the coupling field, many coupling fields are normalized the same way, see subroutine scale_fluxes.  However, strocnxT and strocnyT are computed in the dynamics, not in scale fluxes.  I'm planning to clean up the implementation, see #660, and move the computation into scale fluxes.

I do have one question.  I think there may be a bug.  strocnxT and strocnyT are also passed into icepack_step_therm1.  They are used in the frzmlt_bottom_lateral subroutine in computation of ustar,

```
         ! strocnx has units N/m^2 so strocnx/rho has units m^2/s^2                                                                                                                                           
         ustar = sqrt (sqrt(strocnxT**2+strocnyT**2)/rhow)
```

My question is whether strocnxT and strocnyT here should be the stresses normalized by the fraction or not.  In the current implementation, they are.  But I'm pretty sure that's wrong.  Thoughts?  @eclare108213 @dabail10 

FYI, strocnxT and strocnyT are computed as follows,

```
      ! strocn computed on U, N, E as needed. Map strocn U divided by aiU to T                                                                                                                                
      ! TODO: This should be done elsewhere as part of generalization?                                                                                                                                        
      ! TODO: Rename strocn[x,y]T since it's different than strocn[x,y][U,N,E]                                                                                                                                
      ! conservation requires aiU be divided before averaging                                                                                                                                                 
      work1 = c0
      work2 = c0
      !$OMP PARALLEL DO PRIVATE(iblk,ij,i,j) SCHEDULE(runtime)                                                                                                                                                
      do iblk = 1, nblocks
      do ij = 1, icellu(iblk)
         i = indxui(ij,iblk)
         j = indxuj(ij,iblk)
         work1(i,j,iblk) = strocnxU(i,j,iblk)/aiU(i,j,iblk)
         work2(i,j,iblk) = strocnyU(i,j,iblk)/aiU(i,j,iblk)
      enddo
      enddo
      !$OMP END PARALLEL DO                                                                                                                                                                                   
      call ice_HaloUpdate (work1,              halo_info, &
                           field_loc_NEcorner, field_type_vector)
      call ice_HaloUpdate (work2,              halo_info, &
                           field_loc_NEcorner, field_type_vector)
      call grid_average_X2Y('F', work1, 'U', strocnxT, 'T')    ! shift                                                                                                                                        
      call grid_average_X2Y('F', work2, 'U', strocnyT, 'T')
```

apcraig:
## PR checklist
    Refactor strocnxT, strocnyT implementation
- [X] Developer(s): 
    apcraig
    Bit-for-bit on cheyenne full test suite with 3 compilers if strocnxT and strocnyT computation is preserved, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#036f1f72d5ec5955fc0c8ba76843ea038a83d59a.  
    - [X] bit for bit
    - [ ] more substantial
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

- add aiU to ice_state
- migrate computation of strocnxT and strocnyT to ice_step, needed for thermodynamics, better code reuse.
- add strocnxT_iavg, strocnyT_iavg as coupling field, could be computed differently than the thermodynanics version, but didn't make that change here.  We will be migrating other coupling fields to use the _iavg naming convention over time.  The _iavg field is also passed into icepack as currently computed.
- Update the coupling layers to use the _iavg version of the fields.  See also #67.
- https://github.com/CICE-Consortium/CICE/issues/761 suggests the values of strocnxT, strocnyT maybe shouldn't be scaled for use in thermodynamics.  This commit **does not change** the values after some further discussion.
- These changes are bit-for-bit for a full test suite on cheyenne with 3 compilers without the change to strocnxT, strocnyT.

Closes #761
Also see #660

### DIFF
--- a/cicecore/cicedynB/general/ice_flux.F90
+++ b/cicecore/cicedynB/general/ice_flux.F90
@@ -35,6 +35,16 @@
       ! Dynamics component
       ! All variables are assumed to be on the atm or ocn thermodynamic
       ! grid except as noted
+      !
+      ! scale_fluxes divides several of these by aice "in place", so
+      ! the state of some of these variables is not well defined.  In the
+      ! future, we need to refactor and add "_iavg" versions of the
+      ! fields to clearly differentiate fields that have been divided
+      ! by aice and others that are not.  The challenge is that we need
+      ! to go thru each field carefully to see which version is used.
+      ! For instance, in diagnostics, there are places where these
+      ! fields are multiplied by aice to compute things properly.
+      ! strocn[x,y]T_iavg is the first field defined using _iavg.
       !-----------------------------------------------------------------
 
       real (kind=dbl_kind), dimension (:,:,:), allocatable, public :: &
@@ -56,14 +66,12 @@
 
        ! out to ocean          T-cell (kg/m s^2)
        ! Note, CICE_IN_NEMO uses strocnx and strocny for coupling
-         strocnxT_sf, & ! ice-ocean stress, x-direction at T points, per ice fraction (scaled flux)
-         strocnyT_sf    ! ice-ocean stress, y-direction at T points, per ice fraction (scaled flux)
+         strocnxT_iavg, & ! ice-ocean stress, x-direction at T points, per ice fraction (scaled flux)
+         strocnyT_iavg    ! ice-ocean stress, y-direction at T points, per ice fraction (scaled flux)
 
        ! diagnostic
 
       real (kind=dbl_kind), dimension (:,:,:), allocatable, public :: &
-         strocnxT, & ! ice-ocean stress, x-direction at T points, computed at end of dynamics
-         strocnyT, & ! ice-ocean stress, y-direction at T points, computed at end of dynamics
          sig1    , & ! normalized principal stress component
          sig2    , & ! normalized principal stress component
          sigP    , & ! internal ice pressure (N/m)
@@ -391,10 +399,8 @@
          hwater     (nx_block,ny_block,max_blocks), & ! water depth for seabed stress calc (landfast ice)
          strairxT   (nx_block,ny_block,max_blocks), & ! stress on ice by air, x-direction
          strairyT   (nx_block,ny_block,max_blocks), & ! stress on ice by air, y-direction
-         strocnxT   (nx_block,ny_block,max_blocks), & ! ice-ocean stress, x-direction
-         strocnyT   (nx_block,ny_block,max_blocks), & ! ice-ocean stress, y-direction
-         strocnxT_sf(nx_block,ny_block,max_blocks), & ! ice-ocean stress, x-direction, per ice area
-         strocnyT_sf(nx_block,ny_block,max_blocks), & ! ice-ocean stress, y-direction, per ice area
+         strocnxT_iavg(nx_block,ny_block,max_blocks), & ! ice-ocean stress, x-direction, per ice area
+         strocnyT_iavg(nx_block,ny_block,max_blocks), & ! ice-ocean stress, y-direction, per ice area
          sig1       (nx_block,ny_block,max_blocks), & ! normalized principal stress component
          sig2       (nx_block,ny_block,max_blocks), & ! normalized principal stress component
          sigP       (nx_block,ny_block,max_blocks), & ! internal ice pressure (N/m)
@@ -769,8 +775,8 @@
       ! fluxes sent to ocean
       !-----------------------------------------------------------------
 
-      strocnxT_sf (:,:,:) = c0 ! ice-ocean stress, x-direction (T-cell)
-      strocnyT_sf (:,:,:) = c0 ! ice-ocean stress, y-direction (T-cell)
+      strocnxT_iavg (:,:,:) = c0 ! ice-ocean stress, x-direction (T-cell)
+      strocnyT_iavg (:,:,:) = c0 ! ice-ocean stress, y-direction (T-cell)
       fresh   (:,:,:) = c0
       fsalt   (:,:,:) = c0
       fpond   (:,:,:) = c0
--- a/cicecore/cicedynB/general/ice_flux.F90
+++ b/cicecore/cicedynB/general/ice_flux.F90
@@ -56,12 +56,14 @@
 
        ! out to ocean          T-cell (kg/m s^2)
        ! Note, CICE_IN_NEMO uses strocnx and strocny for coupling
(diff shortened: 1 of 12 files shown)


apcraig:
This is my proposed changes for strocnxT and strocnyT.  This does not yet include the "fix" to the computation of strocnxT and strocnyT, I will make that change next and retest.  I expect results to change, but just wanted to verify that the refactoring was bit-for-bit first.

apcraig:
This is ready for review.  It includes the changes discussed in #761 which is answer changing.  QC tests pass.

cicecore/cicedynB/general/ice_flux.F90
```diff
@@ -56,12 +56,14 @@ module ice_flux
 
        ! out to ocean          T-cell (kg/m s^2)
        ! Note, CICE_IN_NEMO uses strocnx and strocny for coupling
-         strocnxT, & ! ice-ocean stress, x-direction at T points, per ice fraction
-         strocnyT    ! ice-ocean stress, y-direction at T points, per ice fraction
+         strocnxT_sf, & ! ice-ocean stress, x-direction at T points, per ice fraction (scaled flux)
```
eclare108213, line 59: The "scaled flux" terminology is confusing.  Scaled how?  "per ice fraction" means that the quantity is the mean (of whatever) over just the ice fraction of the grid cell.  That should be enough information, but if it's not clear, then say "averaged over the ice fraction".  This comment implies that the stress terms have been divided by the ice fractional area, so that they are not averaged over the full grid cell area.  Instead of _sf, maybe use _ice (although that's confusing too, with the history output standard of using _ai to mean the quantity includes a multiplicative factor of aice)

eclare108213 (commented):
I like that a bunch of duplicated code is being consolidated into one place.  I am not at all convinced that the change in the stress scalings for the thermodynamics is correct.  See my comment at https://github.com/CICE-Consortium/CICE/issues/761#issuecomment-1263958492.  The idea with the thermodynamics is that the calculations are done for a single point in the grid cell, which means that quantities are (or should be) scaled to that point by dividing by the relevant area.  E.g. we use hi=vicen/aicen for the thermodynamic calculations in each thickness category.  The assumption is that the calculated value applies over the whole area of ice being considered (a thickness category or whatever), which is then used in the 'aggregate' step to get the average over the grid cell (or over the ice by then dividing by the total ice area fraction).  So I will need to be convinced that the grid-averaged ocean stress should be used here rather than the local value.  E.g. how does that compare, scaling-wise, with what's done for the wind stress?  It's calculated locally in the thermodynamics, then effectively multiplied by aice for use in the dynamics.  The ocean stress is calculated in the dynamics, including the aice factor (which might be implicit according to the theory from the Connolley et al paper, I don't remember exactly), and which is then divided out for use in the thermo.  Am I thinking about this wrong?

apcraig:
I can't say what the correct strocnxT,strocnyT is for use in icepack.  That's partly why I changed that as a separate commit, it would be very easy to undo that change and keep the rest.  I am looking for guidance on the correct science implementation.

I also agree there are several issues with scale fluxes and variable names.  I used _sf but it would be easy to change.  I recognize _ai is not correct for these fields.  I actually have a really big complaint about how scale_fluxes works.  Fields are passed in and then divided by aice and passed back out.  I think this is confusing.  That means at some places in the code, those variables are divided by aice and other places not.  That's exactly how we get into a situation where strocnxT, strocnyT can mean two different things and we get confused.  And it makes it very difficult to understand what is contained in variables at any given point in the code.  I think we need to have clear naming conventions that differentiate fields that are multiplied by aice and fields that are divided by aice vs fields that are not.  And we should be clear why we need them.  We do not have that now, maybe we're close with _ai, but I'm not sure.  I'd be happy to go through the code and clarify this stuff.  It would require definition of several new variables and extra memory.  But I think that's OK.  This is going to become more and more critical as we shift to coupling fields on multiple grids like (truly) B and C.

apcraig:
As we discussed, I renamed strocn[x,y]T_sf to strocn[x,y]T_iavg.  I also reverted the strocn[x,y]T fields passed into icepack thermodynamics to be consistent (bit-for-bit) with the current version on main.

I started to refactor scale_fluxes and add _iavg fields, but it's more work than I realized.  As we do that for the 25 or so fields that are modified by scale_fluxes, we need to carefully check the code to see whether all uses of those fields are before or after dividing by aice and then make appropriate changes.  For instance, I know some of the diagnostics use fields that are divided by aice so the fields are explicitly multiplied by aice in the diagnostics to get the correct results.  Again, this bigger cleanup has to be done carefully and should be a separate PR.   I noted that again in #67.

I am retesting the changes and will report test results when I have them.

apcraig:
This update PR is bit-for-bit with main, ready for review.

eclare108213 (approved):
Looks good.  My one request is that the main PR info be updated to reflect what's actually in the PR.  The QC testing info is useful -- maybe keep that there or move it to a comment, but make sure it's clear that the strocn fields sent to thermo have not changed in this PR.  Thank you!

apcraig:
**While this change was not implemented at this time**, if we change the computation of strocnxT, strocnyT passed into icepack_step_therm1, results are different, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#84ce3b23b3d76ee1147f19cab02737f6cf41bf2e.  QC passes on cheyenne with that change,
```
PASS cheyenne_gnu_qcchk_gx1_144x1_medium_qc_qcchk compare cice.036f1f72d5.221001-085009 -1 -1 -1
PASS cheyenne_pgi_qcchk_gx1_144x1_medium_qc_qcchk compare cice.036f1f72d5.221001-085009 -1 -1 -1
PASS cheyenne_intel_qcchk_gx1_144x1_medium_qc_qcchk compare cice.036f1f72d5.221001-085009 -1 -1 -1

apcraig:
I updated the PR template.  Thanks for reminding me, that's important.

PR #766, opened 2022-10-06
ice_init: do broadcast 'default_season'

phil-blain:
## PR checklist
    title
- [x] Developer(s): 
    me
Base suite is b4b with main on `robert_intel`:
~~~
349 of 349 tests PASSED
0 of 349 tests PENDING
0 of 349 tests MISSING data
0 of 349 tests FAILED
~~~
    - [X] bit for bit **(since no test in the base suite change `default_season`)**
    - [x] more substantial **if changing `default_season`**
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

When the 'default_season' namelist setting was added in 01494c75 (Nml settings (#208), 2018-10-19) to replace 'l_winter' and 'l_spring', a call to 'broadcast_scalar' was forgotten, such that the 'default_season' value from the namelist is only used on the first MPI process; all other processes get the hardcoded default value 'winter' defined in 'ice_init::input_data', resulting in different initialization across the grid for several variables if anything other than 'winter' is used in the namelist.

Fix that by broadcasting 'default_season' to all MPI procs.

### DIFF
--- a/cicecore/cicedynB/general/ice_init.F90
+++ b/cicecore/cicedynB/general/ice_init.F90
@@ -980,6 +980,7 @@
       call broadcast_scalar(albsnowi,             master_task)
       call broadcast_scalar(ahmax,                master_task)
       call broadcast_scalar(atmbndy,              master_task)
+      call broadcast_scalar(default_season,       master_task)
       call broadcast_scalar(fyear_init,           master_task)
       call broadcast_scalar(ycycle,               master_task)
       call broadcast_scalar(atm_data_format,      master_task)


apcraig:
This is definitely a bug that could affect users if they set default_season to something other than the internal default.  Just didn't affect any of our testing.  Maybe we should try to fill that whole in the testing.  default_season seems to set initial forcing values.  Are they ever used and if so, under what conditions?

eclare108213:
The default_season 'forcing' serves several purposes -- initialization of potentially missing forcing fields, simple values for short-period testing, and order-of magnitude guidance for users' own forcing fields.  Thank you for fixing this bug, @phil-blain!  @apcraig, if you want to add a test case for a different season, it could also test starting the code on a date other than Jan 1.

phil-blain:
I can add a new test to this PR if we want to do that. I noticed that we do have tests with different `default_season`, but only in the Icepack base suite, not CICE's.

Issue #737, opened 2022-07-29
variable drag coefficients are computed on T-cell centers, need to be moved to velocity locations


eclare108213:

apcraig:
I'm happy to help implement this.  Can someone provide a little guidance about what needs to change and where in the code.  thanks.

eclare108213:
Search for cdn_atm and cdn_ocn, which are (neutral) drag coefficients calculated in icepack_atmo.F90, called from step_therm1.  Basically, all of the _thermodynamic_ calculations that use them should be done at cell centers, and they (or variables derived from them) only need to be shifted for the momentum equation.  E.g. cdn_atm is used for the atmo turbulent boundary layer calculations, from which we get atmo surface stress.  I think the stress calculations are done in T cells then shifted to the velocity points before being used for solving the dynamics equations, so they should be okay as-is (but check).  cdn_atm also appears in init_coupler_flux, for RASM -- I'm not sure what's needed there.  This github issue is probably mainly about cdn_ocn in the ice_dyn_*.F90 modules, for which an equivalent shifting procedure should be done as for the atmo stresses.  It is more complicated, though, because the ocean stress term is split up in the code, which will require a some care to get right.

apcraig:
I had a look thru the code and think the above assessment is correct.  It looks like all the cdn variables are defined on the T-grid.  They are computed in icepack_atmo, including the cdn_ocn terms.  And all the air/ice stresses are also computed in icepack on the T grid.  Those are mapped to the U, E, and N grid as needed.  

Only cdn_ocn is used directly in the dynamics in dyn_finish, stepu and similar.  It is passed into subroutines with related variables already on the U, E, or N grid.  So I think we need to create cdn_ocn[U,E,N] at the start of the dynamics and then instead of using cdn_ocn, use cdn_ocn[U,E,N].  I assume an appropriate mapping would be "S" which is a masked area-weighted average.  The ocean stress terms are computed separately on the U, E, or N grid, I think there is no complication with passing in the appropriate cdn_ocn and using it.

Unless there is any disagreement, I'll make the modifications and create a PR.  @JFLemieux73.  This will also changes answers for the B grid.

Issue #734, opened 2022-07-29
kinetic energy diagnostic mixes T and U variables for B grid


eclare108213:

apcraig:
I see this in ice_diagnostics.F90.  I assume what we need to do is compute work1 on either the T or U grid consistently?  It seems like it would be better to compute the KE on the U grid, moving the T variables to the U grid?  For the C grid, would we compute the "u" part of KE on the E face and the "v" part of KE on the N face then add them?  Or is it better to just average the U,V to T and compute KE on T?


```
      ! total ice-snow kinetic energy                                                                                                                                                                         
      !$OMP PARALLEL DO PRIVATE(iblk,i,j)                                                                                                                                                                     
      do iblk = 1, nblocks
         do j = 1, ny_block
         do i = 1, nx_block
            work1(i,j,iblk) = p5 &
                           * (rhos*vsno(i,j,iblk) + rhoi*vice(i,j,iblk)) &
                           * (uvel(i,j,iblk)**2 + vvel(i,j,iblk)**2)
         enddo
         enddo
      enddo
      ! Eventually do energy diagnostic on T points.                                                                                                                                                          
!     if (grid_ice == 'CD') then                                                                                                                                                                              
!     !$OMP PARALLEL DO PRIVATE(iblk,i,j)                                                                                                                                                                     
!     do iblk = 1, nblocks                                                                                                                                                                                    
!        do j = 1, ny_block                                                                                                                                                                                   
!        do i = 1, nx_block                                                                                                                                                                                   
!           call grid_average_X2Y('E2TS',uvelE,uvelT)                                                                                                                                                         
!           call grid_average_X2Y('N2TS',vvelN,vvelT)                                                                                                                                                         
!           work1(i,j,iblk) = p5 &                                                                                                                                                                            
!                          * (rhos*vsno(i,j,iblk) + rhoi*vice(i,j,iblk)) &                                                                                                                                    
!                          * (uvelT(i,j,iblk)*uvelT(i,j,iblk) &                                                                                                                                               
!                          +  vvelT(i,j,iblk)*vvelT(i,j,iblk))                                                                                                                                                
!        enddo                                                                                                                                                                                                
!        enddo                                                                                                                                                                                                
!     enddo                                                                                                                                                                                                   
!     endif                                                                                                                                                                                                   
!     !$OMP END PARALLEL DO                                                                                                                                                                                   
      ketotn = c0
      ketots = c0
[...]

Issue #655, opened 2021-11-03
C-grid output


eclare108213:
_We will need proper C-grid output early on in the development process, for checking our test results._

**History Files**
The grids supported by history files will need to be expanded.  This is defined in ice_history_shared (i.e. tcstr2D).  ice_history_write will also need to be expanded to write the new grids as well as support the new vcoord values.  Unfortunately, this will need to be done in the binary, netcdf, and pio versions separately and then tested.  

History variables will need to be associated with the appropriate grid and this will probably depend on the grid system for some variables.  

There is likely an opportunity to refactor some of the history implementation to provide greater flexibility for adding fields/grids and improving code reuse, but this may also be a big distraction.  We will take a simpler, direct approach to expand features, for now.  

- [x] Determine scope of history output changes needed
- [X] Define and write new grids, etc
- [x] Associate history variables with the appropriate grid (use grid system namelist flag)
- [x] Review current grid settings for history variables (uatm, vatm), which might be buggy.
- [ ] Review documentation and alter as necessary.

**Diagnostics and budgets**
The current implementation of global sums and other reductions should be directly usable on the N and E grids.  Individual diagnostics need to be reviewed and may require distinct implementations for the B-grid and C-grid as a result of different staggering of quantities.  The grid_system variable can be used to control diagnostic computations that are different.

Diagnostics associated with budget conservation is a larger concern.  As new algorithms are implemented in CICE, diagnostics associated with conservation and the general budget will need to be revised.  It’s strongly recommended that the budget diagnostics be modified as the C-grid solvers are implemented.  The grid_system namelist can be used to trigger different budget diagnostic implementations.

- [x] Determine scope of diagnostic and budget output changes needed
- [ ] Check conservation diagnostics
- [ ] Modify diagnostics as needed (use grid system namelist)
- [ ] Review documentation and alter as necessary.

eclare108213:
Some notes on diagnostics:
- Total kinetic energy of ice+snow is calculated using (uvel, vvel) incorrectly (mass and velocity are not co-located). The T-cell area is used, which is fine.  Velocity should be interpolated to T-cell centers.  _**NOTE:  use vector averages (magnitude and direction) rather than averaging components!**_
- Maximum ice speed assumes uvel and vvel are co-located.  This is okay for the CD grid, but we'll need to take the maximum over all of the N and E edges.
- Likewise, there is a root mean square ice speed diagnostic.
[...]

PR #771, opened 2022-10-12
Update computation of cdn_ocn for use in dynamics

apcraig:
## PR checklist
    Update computation of cdn_ocn for use in dynamics
- [X Developer(s): 
    apcraig
    All tests pass but answers have changed as expected, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#af8cc84308beb3dd959639f0142a8cbb29a9a5e6.  QC tests pass,
```
PASS cheyenne_intel_qcchk_gx1_144x1_medium_qc_qcchk compare cice.422117fc5d.221008-085008 -1 -1 -1
PASS cheyenne_pgi_qcchk_gx1_144x1_medium_qc_qcchk compare cice.422117fc5d.221008-085008 -1 -1 -1
PASS cheyenne_gnu_qcchk_gx1_144x1_medium_qc_qcchk compare cice.422117fc5d.221008-085008 -1 -1 -1
```
    - [ ] bit for bit
    - [X] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

- Compute and used cdn_ocn on U, E, and N cell locations as needed for dynamics.
- Add halo updates in dynamics before calling grid_average.  This doesn't change answers, but is the safe thing to do in general.  Performance does not seem to be affected.

### DIFF
--- a/cicecore/cicedynB/dynamics/ice_dyn_eap.F90
+++ b/cicecore/cicedynB/dynamics/ice_dyn_eap.F90
@@ -177,6 +177,7 @@
          vocnU      , & ! j ocean current (m/s)
          ss_tltxU   , & ! sea surface slope, x-direction (m/m)
          ss_tltyU   , & ! sea surface slope, y-direction (m/m)
+         cdn_ocnU   , & ! ocn drag coefficient
          tmass      , & ! total mass of ice and snow (kg/m^2)
          waterxU    , & ! for ocean stress calculation, x (m/s)
          wateryU    , & ! for ocean stress calculation, y (m/s)
@@ -186,7 +187,9 @@
          umassdti       ! mass of U-cell/dte (kg/m^2 s)
 
       real (kind=dbl_kind), allocatable :: &
-         fld2(:,:,:,:)  ! temporary for stacking fields for halo update
+         fld2(:,:,:,:), & ! temporary for stacking fields for halo update
+         fld3(:,:,:,:), & ! temporary for stacking fields for halo update
+         fld4(:,:,:,:)    ! temporary for stacking fields for halo update
 
       real (kind=dbl_kind), dimension(nx_block,ny_block,8):: &
          strtmp         ! stress combinations for momentum equation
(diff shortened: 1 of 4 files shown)


apcraig:
Please check if this seems correct.  This closes #737.

eclare108213:
I'm confused about when/how variable form drag is invoked -- this does not seem to be documented well in Icepack.  E.g. `formdrag`=T should turn on the formdrag parameterization, which adjusts the ocean drag coefficients for various ice characteristics (in icepack_atmo.F90, not the most intuitive place), but if it is not turned on then `Cdn_ocn` is set to a constant value.  There is also an option `calc_dragio`, which when true apparently varies the ocean drag coefficient in some other way (and might conflict with `formdrag`?).  If neither of these parameterizations is turned on, then `Cdn_ocn` should be constant (=`dragio`=5.36e-3), and so for the B grid I would not expect the answers to change.  Do they?

apcraig:
The B cases do changes answers.  Looking closer at output, I think we're getting a round off level change in the cdn_ocn when we average a constant field to another grid.  On the first timestep, the only differences in the log file are in KE, ice spped, and arwt ocn heat flux and those are roundoff different in the global sum.  The difference then grows.  So I think that's consistent with our expectation and it's OK.

I don't have any insight into the underlying computation of cdn_ocn, just focused on the grid location.  Happy to make other changes, in documentation or otherwise, as needed.

eclare108213:
So on the B-grid it's a roundoff change...  Is the "more substantial" tickmark above for C-grid simulations?

I found the `calc_dragio` documentation and will make an issue to make the documentation more transparent (`calc_dragio` isn't mentioned, which is why I didn't find it initially).  In fact it's also a constant value (as long as the ice roughness and upper ocean grid spacing are constants), but it is computed from those parameters rather than assigned.

apcraig:
Would this also address tasks in #655?  Or are there still more things to do there.  And if so, should I include them now?

eclare108213:
Were the two ice speed diagnostics fixed?  Sorry I've lost track...  And you're working on TKE also, now?

apcraig:
Sorry, the KE/TKE is not related to this PR, my mistake.  Lets move discussion to #734.

eclare108213 (approved):

PR #774, opened 2022-10-14
ice_dyn_vp: allow for bit-for-bit reproducibility under `bfbflag`

phil-blain:
## PR checklist
    title
- [X] Developer(s): 
   me
```
350 measured results of 350 total results
348 of 350 tests PASSED
0 of 350 tests PENDING
0 of 350 tests MISSING data
2 of 350 tests FAILED
```

```
$ ./results.csh|\grep FAIL
FAIL ppp6_intel_smoke_gx3_4x1_dynpicard complog 007fbff different-data
FAIL ppp6_intel_smoke_gx3_4x1_dynpicard compare 007fbff 2.58 0.75 1.14 different-data
```

more details below.

    - [ ] bit for bit
    - [X] different at roundoff level
    - [X] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

### Summary

This PR refactors the VP solver to allow bit-for-bit reproducibility when using `bfbflag`. Currently on `main`, global sums are accumulated locally before calling `global_sum` on a scalar value. With the new code, when `bfbflag` is not `off`, we call `global_sum` on an array, allowing the existing algorithms (reprosum, ddpdd, etc) to be used for b4b reproducibility. See also previous implementations and design decisions in #763. In this new version, the code has the same number of global sums as the `main` version.

### PR overview

First we update the docs a little:
- 1/15 c9144ea doc: fix typo in index (bfbflag)
- 2/15 545377e doc: correct default value of 'maxits_nonlin'
- 3/15 399d873 doc: VP solver is validated with OpenMP

Then we update our machines files:

- 4/15 75b934a machines: eccc : add ICE_MACHINE_MAXRUNLENGTH to ppp[56]
- 5/15 405df63 machines: eccc: use PBS-enabled OpenMPI for 'ppp6_gnu'
- 6/15 de08b34 machines: eccc: set I_MPI_FABRICS=ofi
- 7/15 f7cbcb2 machines: eccc: set I_MPI_CBWR for BASEGEN/BASECOM runs

Then we fix a bug we discovered along the way:

- 8/15 fd54d5d ice_dyn_vp: fgmres: exit early if right-hand-side vector is zero

Then the core of the changes, first introducing two new functions and then refactoring the code to use them:
- 9/15 b96ea5b ice_dyn_vp: add global_norm, global_dot_product functions
- 10/15 c37c645 ice_dyn_vp: use global_{norm,dot_product} for bit-for-bit output reproducibility
- 11/15 33ccce2 ice_dyn_vp: do not skip halo updates in 'pgmres' under 'bfbflag'
- 12/15 6aab357 ice_dyn_vp: use global_{norm,dot_product} for bit-for-bit log reproducibility

Some clean-up and doc updates:
- 13/15 5a8b2d1 ice_dyn_vp: remove unused subroutine and cleanup interfaces
- 14/15 8901899 doc: mention VP solver is only reproducible using 'bfbflag'

And finally a tweak to the default solver settings:
- 15/15 9f1efbb ice_dyn_vp: improve default parameters for VP solver

### Testing

- I ran the `decomp_suite` with `-s dynpicard,reprosum` (i.e. `bfbflag=reprosum`) and all tests pass
  - I also ran it with `bfbflag=ddpdd`, all tests pass
  - I also ran it with `bfbflag=lsum16`, all tests pass
- I ran the `nothread_suite` with `-s dynpicard,reprosum` and all tests pass
- I also ran both these suites twice from the same commit and compared the results using `bgen/bcmp` to also ensure reproducibility on the same decomposition + MPI distribution. All pass.
- I ran the QC test for the failing test in the base_suite. I ran it on 4 pairs of results.
  - `main` code vs new code
  - `main` code vs new code with `bfbflag=lsum8`, so exercising the array version of `global_sum`
  - `main` code with the same changes to the solver parameters as in 15/15 vs new code
  - `main` code with the same changes to the solver parameters as in 15/15 vs new code with `bfbflag=lsum8`

The first two pairs pass the 2-stage test and the NH quadratic skill test **but fail in the SH**.
The last two **pass all 3 tests** (2-stage, quadratic skill NH, quadratic skill SH).

I don't think it's too surprising that the first two pairs fail the test, as with 4 nonlinear iterations the solution is not really converged at all from what we've seen. With 10 iterations we usually get at least one significant digit of convergence.

### Performance
- The old and new code both took 3 hours to simulate the 5 years QC data (with the new solver parameters)
- The new code with `bfbflag=lsum8` (i.e. array version of `global_sum`) and the new solver parameters) took 2.5 hours, so it's actually faster in this case!

This is not what I was expecting since my previous timings showed a modest time increase with the `lsum8` case, but those timings were done on single-day runs, so the 5 year runs are maybe more "realistic" timings. So in the end we might want to default to the array version ? What do others think ?

### Questions 
- After 10/15, the `global_allreduce_sum` function is not used anymore. It was added by me for the original implementation of the VP solver but is not used anywhere else in the code. I can remove it if we feel we do not need it. My opinion is it does not hurt to keep it for now.
- The doc mentions 

    > For best performance, the `off` (or `lsum8` which is equivalent) setting is recommended.
    
    but after these changes, `off` is not the same as `lsum8` for the VP solver, since with `lsum8`  we use the array version of `global_sum`... I don't think it matters that much, but I could also just remove that parenthesis from the doc.


apcraig:
This all looks great!  A couple questions/comments,

Under performance you say the old code time was 3 hours and the new 3.5 hours and then suggest new is faster.  Something doesn't seem consistent, are I misunderstanding?

Overall, I sort of vote for removing global_sum_prod subroutines if we don't need them anymore, and we don't really see a need for them at this point.  It's just more code to test/maintain if we don't need it.  Also, they exist in the serial and mpi side, so removing them reduces the burden on keep the two versions synced up a bit.  They are tested in the sumchk unit tester, so if we remove them, we need to update the sumchk test.  However, if others prefer we keep them, I'm OK with that.  Should we create an issue and defer for now?

With regard to documentation of global sums, I'm trying to think about the best way to update the usage and documentation overall.  In some ways, we've overloaded the bfbflag.  It means one thing in global_sums and a slightly different but mostly overlapping thing in the vp solver.  Thinking out loud, do we need a new bfbflag option (offvp?) that sets bfbflag=off and uses the scalar global sum in vp while all other bfbflag settings (including off) would call into the standard global_sum from vp?  Will there ever be a time where we want the vp global_sum to be set different from the diagnostics global sum (maybe vp=off but diagnostics=bfb)?  If so, then we need a new namelist in the dynamics for the vp global sum setting.  I'm largely happy with what's there now, but just throwing out some ideas.  If we keep what we have, then I suggest we keep the documentation simple and clean up what we can (remove suggestion that sum8==off).

Have all the OpenMP comments been turned back on, are all block loops shared parallel now?

apcraig:
Quick follow up on OpenMP, what about line 337 in ice_dyn_vp.F90?

phil-blain:
> Under performance you say the old code time was 3 hours and the new 3.5 hours and then suggest new is faster. Something doesn't seem consistent, are I misunderstanding?

Sorry, my mistake. The new code with `bfbflag=lsum8` is actually took 2.5 hours. So yes it's faster.

> 
> Overall, I sort of vote for removing global_sum_prod subroutines if we don't need them anymore, and we don't really see a need for them at this point. It's just more code to test/maintain if we don't need it. Also, they exist in the serial and mpi side, so removing them reduces the burden on keep the two versions synced up a bit. They are tested in the sumchk unit tester, so if we remove them, we need to update the sumchk test. However, if others prefer we keep them, I'm OK with that. Should we create an issue and defer for now?

Just to be clear, here I'm talking about the `global_allreduce_sum` module procedure, not `global_sum_prod`. `global_sum_prod` is used at a few places (ice_diagnostics, ice_transport_driver). `global_allreduce_sum` is not used anymore and could be removed if we wish.

> 
> With regard to documentation of global sums, I'm trying to think about the best way to update the usage and documentation overall. In some ways, we've overloaded the bfbflag. It means one thing in global_sums and a slightly different but mostly overlapping thing in the vp solver. Thinking out loud, do we need a new bfbflag option (offvp?) that sets bfbflag=off and uses the scalar global sum in vp while all other bfbflag settings (including off) would call into the standard global_sum from vp? Will there ever be a time where we want the vp global_sum to be set different from the diagnostics global sum (maybe vp=off but diagnostics=bfb)? If so, then we need a new namelist in the dynamics for the vp global sum setting. I'm largely happy with what's there now, but just throwing out some ideas. If we keep what we have, then I suggest we keep the documentation simple and clean up what we can (remove suggestion that sum8==off).

I think I would keep things as simple as possible and just remove the parenthesis in the doc :)

> Have all the OpenMP comments been turned back on, are all block loops shared parallel now?
> Quick follow up on OpenMP, what about line 337 in ice_dyn_vp.F90?

Oh yeah, this:

https://github.com/CICE-Consortium/CICE/blob/5a8b2d18f9fca45035c4adde35201cb2e9eda9c2/cicecore/cicedynB/dynamics/ice_dyn_vp.F90#L335-L342

This dates back to when `ice_dyn_vp` was created by copying `ice_dyn_evp`. Then when you recently fixed a bunch of OpenMP loops in d1e972a (Update OMP (#680), 2022-02-18), you fixed the private variables in `ice_dyn_vp` for this loop as in `ice_dyn_evp` but kept the loop unthreaded, since (I'm guessing) we already knew that the VP solver had other b4b issues. Now that you mention it, I checked and that line is the only remaining `TCXOMP` loop in the code.

I'll do a few more tests with an actual `!$OMP` directive over that loop and report back.

apcraig:
Thanks for clarifications.

Sorry for my confusion, I would get rid of the global_allreduce_sum methods.  I don't think those are going to be needed in the future.  This is tested in sumchk, so we would need to update that as well.  Again, maybe best to be a new issue and done separately.

I'm fine with the plan for bfbflag and documentation of it.

Correct, when I worked on the OpenMP validation, vp had other problems that made it difficult for me to validate the OpenMP in vp.  I commented out OpenMP directives that were clearly problematic in vp to improve overall robustness and then I think vp passed the OpenMP testing generally.  But we need to revisit that.  I think the first question is that one open OpenMP loop that is commented out.  Also, we should make sure there are no other iblk loops in vp that should have OpenMP added.  And we should try to get all loops on and validated with OpenMP.  If a loop cannot have OpenMP, lets try to add a comment why.  Again, this could be a separate PR but would be great for it to be settled now.  Thanks!

phil-blain:
> Sorry for my confusion, I would get rid of the global_allreduce_sum methods. I don't think those are going to be needed in the future. This is tested in sumchk, so we would need to update that as well. Again, maybe best to be a new issue and done separately.

OK. I can remove them and update the unit test. And it's relatively easy to re-add them if needed from the Git history :)


> I'm fine with the plan for bfbflag and documentation of it.

 OK. Will update the doc.

> Correct, when I worked on the OpenMP validation, vp had other problems that made it difficult for me to validate the OpenMP in vp. I commented out OpenMP directives that were clearly problematic in vp to improve overall robustness and then I think vp passed the OpenMP testing generally. But we need to revisit that.

My understanding is that after this PR, everything is validated as far as I know. The `decomp` suite has a lot of different MPI x OpenMP distributions in additions to the various block decompositions, and as mentioned all these tests pass with `bfbflag=reprosum`.

> I think the first question is that one open OpenMP loop that is commented out.

Yes, I'm testing that and will update the PR with the results. 

> Also, we should make sure there are no other iblk loops in vp that should have OpenMP added. And we should try to get all loops on and validated with OpenMP. If a loop cannot have OpenMP, lets try to add a comment why. Again, this could be a separate PR but would be great for it to be settled now. 

I just had a look and the only loops that are not threaded are those that compute `ntot` and the ones in  `vec_to_arrays` and `arrays_to_vec`. I'm planning to remove these subroutines when I go back to work on the parallelization of the Anderson solver. Right now these loops are only involved in the Anderson solver code, not in the Picard solver. So I would keep them as is for now.

phil-blain:
> > I think the first question is that one open OpenMP loop that is commented out.
> 
> Yes, I'm testing that and will update the PR with the results.

Great news, `decomp` suite passes with that loop OpenMP'd (with `bfbflag=reprosum`, as in the rest of my tests). So I'll rework a but my first commits to include that change (next week). Have a good week-end! :)

apcraig:
Thanks @phil-blain.  All sounds good.  If you have questions/problems with the unit tester, feel free to defer.  The unit testers are very useful, but pretty rough in some ways.

eclare108213 (commented):
This all looks good to me.  I'll let the experts review/approve/merge.

phil-blain:
I've just pushed a v2 addressing all feedback.

apcraig:
This looks great to me, I'm going to run a few tests and then will formally approve and merge.

apcraig (approved):
All my testing looks good.  Very happy this is all sorted out!!  Thanks.

apcraig:
I will merge in the next day or so unless someone says "not yet".  Thanks!

PR #788, opened 2022-11-14
Added check for individual namelist variables

daveh150:

## PR checklist
Added check for namelist variables
- [x] Developer(s): 
    David Hebert
quick_suite passed. Only ran small test since this is a namliest check, not physics:
16 measured results of 16 total results
16 of 16 tests PASSED
0 of 16 tests PENDING
0 of 16 tests MISSING data
0 of 16 tests FAILED

    - [ ] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 
Issue #677 changed how namelists were read to allow random reading, and removed previous namelist checks. I added namelist checks back with these updates. I ran quick_suite, then introduced errors in each of the namelists and re-ran. That worked with my gnu compiler, logs show results like this:

 (abort_ice)ABORTED:
 (abort_ice) called from /home/hebert/git/daveh150_new/cicecore/cicedynB/analysis/ice_histo\
ry_drag.F90
 (abort_ice) line number          107
 (abort_ice) error = (init_hist_drag_2D)ERROR: icefields_drag_nml reading     test_icefields_drag_nml = .true.

 I think #677 was targed at NAG, does someone have NAG compiler to test with?  Thanks!

### DIFF
--- a/cicecore/cicedynB/analysis/ice_history.F90
+++ b/cicecore/cicedynB/analysis/ice_history.F90
@@ -102,7 +102,9 @@
       character (len=25) :: &
          cstr_gat, cstr_gau, cstr_gav, &     ! mask area name for t, u, v atm grid (ga)
          cstr_got, cstr_gou, cstr_gov        ! mask area name for t, u, v ocn grid (go)
-      character(len=char_len) :: description
+
+      character(len=char_len)      :: description
+      character(len=char_len_long) :: tmpstr2 ! for namelist check
 
       character(len=*), parameter :: subname = '(init_hist)'
 
@@ -239,11 +241,18 @@
          nml_error =  1
          do while (nml_error > 0)
             read(nu_nml, nml=icefields_nml,iostat=nml_error)
+            ! check if error
+            if (nml_error /= 0) then
+               ! backspace and re-read erroneous line
+               backspace(nu_nml)
+               read(nu_nml,fmt='(A)') tmpstr2
+
+               call abort_ice(subname//'ERROR: icefields_nml reading ' // &
+                    trim(tmpstr2), &
+                    file=__FILE__, line=__LINE__)
+            endif
          end do
-         if (nml_error /= 0) then
-            call abort_ice(subname//'ERROR: icefields_nml reading ', &
-               file=__FILE__, line=__LINE__)
-         endif
+
          close(nu_nml)
          call release_fileunit(nu_nml)
       endif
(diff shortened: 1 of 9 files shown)


apcraig:
Can you check your #667 reference, I think you're referring to a different PR/Issue.

apcraig:
It's #677.

daveh150:
Oh, yes, sorry, I updated the original PR text to say #677.

apcraig:
This implementation looks similar to #671.  This will not work on the NAG compiler.  Because we support namelist groups in any order in the namelist file, the new strategy will trap an error on NAG if it reads in the wrong namelist group, which is something we need to support.  That's why we have a "do while (nml_error>0)" with an abort outside the do loop.  It cycles until it finds the right namelist group in the file.  Adding the "if (nml_error /=0)" in the middle of the loop is guaranteed to cause an abort if the next namelist group is not the right one.

Original implementation:

         nml_error =  1
         do while (nml_error > 0)
            read(nu_nml, nml=icefields_nml,iostat=nml_error)
         end do
         if (nml_error /= 0) then
            call abort_ice(subname//'ERROR: icefields_nml reading ', &
               file=__FILE__, line=__LINE__)
         endif

New implementation proposed here:

         nml_error =  1
         do while (nml_error > 0)
            read(nu_nml, nml=icefields_nml,iostat=nml_error)
            ! check if error
            if (nml_error /= 0) then
               ! backspace and re-read erroneous line
               backspace(nu_nml)
               read(nu_nml,fmt='(A)') tmpstr2

               call abort_ice(subname//'ERROR: icefields_nml reading ' // &
                    trim(tmpstr2), &
                    file=__FILE__, line=__LINE__)
            endif
         end do

I think this PR should be closed.

daveh150:
Yes, true, the strategy is the same. I guess I misunderstood the additions with all the rewinds. I thought it was rewinding to the start of the file after every namelist read. Then reading the specified namelist, where the additional check was entered. I guess if that doesn't work with NAG then yea, this can be discarded. Thanks you.

apcraig:
No problem, thanks @daveh150.  I really wish there were a good way to do what you propose.  It would be very helpful to have something like this.  Lets keep exploring.  I'll close this now.

Issue #798, opened 2022-11-30
tmpstr not defined in ice_init.F90


dabail10:
Looks like this code is scheduled for deprecation, but it does not compile because tmpstr is not defined.

! each task gets unique ice log filename when if test is true, for debugging
         if (1 == 0) then
            call get_fileUnit(nu_diag)
            write(tmpstr,'(a,i4.4)') "ice.log.task_",my_task
            open(nu_diag,file=tmpstr)
         endif

eclare108213:
Just need to change tmpstr to tmpstr2.  Amazing this hasn't been caught before now.  Which compiler?

dabail10:
This was on cheyenne intel. I can't recall if I had optimization turned off.

dabail10:
This was within CESM, so CESMCOUPLED is turned on.

dabail10:
I have a PR coming with the saltflux stuff and I will fix it there.

apcraig:
I knew it was in CESM, but was just pointing out that the standalone model would not have caught this.  I just checked RASM.  The code is there and CESMCOUPLED is on.  We have a slightly older version.  tmpstr is defined there.  My guess is that when I was cleaning up compiler warning messages about unused variables and such, I removed tmpstr from the module.  This would not have been needed for standalone.  I suspect it's my fault and probably happened fairly recently.  Sorry about that.  Thanks @dabail10 for fixing this.

dabail10:
No worries. Testing my changes now.

dabail10:
Fixed in #799.

Issue #783, opened 2022-11-03
OPENACC code in ice_dyn_evp_1d.F90


dabail10:
Apparently the NVIDIA compiler is flagging the OPENACC directives with a syntax error:

NVFORTRAN-S-0034-Syntax error at or near ACCLOOP (/glade/work/jedwards/sandboxes/cesm2_x_alpha/components/cice/src/cicecore/cicedynB/dynamics/ice_dyn_evp_1d.F90: 906)
0 inform, 0 warnings, 1 severes, 0 fatal for evp1d_halo_update

using: Currently Loaded Modules:

ncarenv/1.3 3) nvhpc/22.2 5) netcdf-mpi/4.8.1 7) esmf-8.4.0b08_casper-ncdfio-openmpi-O 9) pio/2.5.9
cmake/3.18.2 4) openmpi/4.1.4 6) pnetcdf/1.12.3 8) ncarcompilers/0.5.0

This is the acc loop starting at line 906. @TillRasmussen , can you have your guys look at this?

TillRasmussen:
ok. I think that I have the reason but I need to talk to our it department tomorrow in order to get acces to a nvidia compiler. I will test it then. If you need a quick fix then try  the following

From line 899. Se bold line below. This should be replaced with 
      **!$acc present(uvel, vvel)**
Thus "&" is removed
#ifdef _OPENACC
      !$acc parallel &
      **!$acc present(uvel, vvel) &**
      !$acc loop
      do iw = 1, NAVEL_len
         if (halo_parent(iw) == 0) cycle
         uvel(iw) = uvel(halo_parent(iw))
         vvel(iw) = vvel(halo_parent(iw))
      end do
      !$acc end parallel
#else

dabail10:
I wondered about the continuation characters. I'll mention it to our software engineers.

dabail10:
Fixed in PR #799 .

Issue #706, opened 2022-03-29
Add true salt flux computation.


dabail10:
This method would compute the salt and freshwater flux using the actual bulk salinity in the sea ice instead of 4 psu.

This is essentially computing d(hS)/dt or salt volume change in the thermodynamics.

PR #799, opened 2022-12-02
Saltflux option for CICE

dabail10:

## PR checklist
This adds the saltflux option in CICE. Also, a couple cleanup things.
- [X] Developer(s): 
dabail10 (D. Bailey)
Bfb with the aux_cice tests in CESM.
Also, did base_suite on cheyenne using intel, pgi, gnu. Added salt test (only in gnu base_suite here).
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#f19faa770d7e06903bbae621687adadc0bce42ba
    - [X] bit for bit
    - [ ] more substantial 
    - [X] Yes
    - [ ] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 
In addition to the saltflux option (Issue #706 ) I have added a couple cleanup things:

1. Fix an ACC directive in ice_dyn_evp_1d.F90 (issue #783)
2. Fix tmpstr. Issue #798 
3. Remove the CESM1_PIO stuff in io_pio2/ice_restart.F90. This is not necessary now that we are not using landblock elimination. (Issue #675)
4. Put an ifdef around the OpenMP initialization code in ice_init.F90. Not sure what is going on here yet. (Issue #800 )
5. I have added an option (wrapped in #CESMCOUPLED) to initialize ice in the 'default' case based on Tair instead of sst. When coupling in CESM, the SST has not been received from the ocean on initialization, but it has received Tair.

### DIFF
--- a/cicecore/cicedyn/analysis/ice_history.F90
+++ b/cicecore/cicedyn/analysis/ice_history.F90
@@ -420,10 +420,6 @@
          f_taubyE = f_tauby
       endif
 
-#ifndef ncdf
-      f_bounds = .false.
-#endif
-
       ! write dimensions for 3D or 4D history variables
       ! note: list of variables checked here is incomplete
       if (f_aicen(1:1) /= 'x' .or. f_vicen(1:1) /= 'x' .or. &
--- a/doc/source/science_guide/sg_horiztrans.rst
+++ b/doc/source/science_guide/sg_horiztrans.rst
@@ -35,8 +35,11 @@ versions but have not yet been implemented.
 
 Two transport schemes are available: upwind and the incremental
 remapping scheme of :cite:`Dukowicz00` as modified for sea ice by
-:cite:`Lipscomb04`. The upwind scheme is naturally suited for a C grid discretization. As such, the C grid velocity components (i.e. :math:`uvelE=u` at the E point and :math:`vvelN=v` at the N point) are directly passed to the upwind transport scheme. On the other hand, if the B grid is used, :math:`uvel` and :math:`vvel` (respectively :math:`u` and :math:`v` at the U point) are interpolated to the E and N points such that the upwind advection can be performed. Conversely, as the remapping scheme was originally developed for B grid applications, :math:`uvel` and :math:`vvel` are directly used for the advection. If the remapping scheme is used for the C grid, :math:`uvelE` and :math:`vvelN` are first interpolated to the U points before performing the advection.
+:cite:`Lipscomb04`. 
 
+- The upwind scheme uses velocity points at the East and North face (i.e. :math:`uvelE=u` at the E point and :math:`vvelN=v` at  the N point) of a T gridcell.  As such, the prognostic C grid velocity components (:math:`uvelE` and :math:`vvelN`) can be passed directly to the upwind transport scheme.  If the upwind scheme is used with the B grid, the B grid velocities, :math:`uvelU` and :math:`vvelU` (respectively :math:`u` and :math:`v` at the U point) are interpolated to the E and N points first.  (Note however that the upwind scheme does not transport all potentially available tracers.)
+
+- The remapping scheme uses :math:`uvelU` and :math:`vvelU` if l_fixed_area is false and :math:`uvelE` and :math:`vvelN` if l_fixed_area is true.  l_fixed_area is hardcoded to false by default and further described below.  As such, the B grid velocities (:math:`uvelU` and :math:`vvelU`) are used directly in the remapping scheme, while the C grid velocities (:math:`uvelE` and :math:`vvelN`) are interpolated to U points first.  If l_fixed_area is changed to true, then the reverse is true.  The C grid velocities are used directly and the B grid velocities are interpolated.
 
 The remapping scheme has several desirable features:
 
@@ -464,14 +467,14 @@ In general, the fluxes in this expression are not equal to those implied
 by the above scheme for locating departure regions. For some
 applications it may be desirable to prescribe the divergence by
 prescribing the area of the departure region for each edge. This can be
-done in CICE 4.0 by setting `l\_fixed\_area` = true in
+done by setting `l\_fixed\_area` = true in
 **ice\_transport\_driver.F90** and passing the prescribed departure
 areas (`edgearea\_e` and `edgearea\_n`) into the remapping routine. An extra
 triangle is then constructed for each departure region to ensure that
 the total area is equal to the prescribed value. This idea was suggested
 and first implemented by Mats Bentsen of the Nansen Environmental and
 Remote Sensing Center (Norway), who applied an earlier version of the
-CICE remapping scheme to an ocean model. The implementation in CICE v4.0
+CICE remapping scheme to an ocean model. The implementation in CICE
 is somewhat more general, allowing for departure regions lying on both
 sides of a cell edge. The extra triangle is constrained to lie in one
 but not both of the grid cells that share the edge. Since this option


cicecore/cicedyn/analysis/ice_diagnostics.F90
```diff
@@ -1310,6 +1333,12 @@ subroutine init_mass_diags
       toten = global_sum(work1, distrb_info, field_loc_center, tarean)
       totes = global_sum(work1, distrb_info, field_loc_center, tareas)
 
+      ! north/south salt
+      call total_salt (work2)
+      totsn = global_sum(work2, distrb_info, field_loc_center, tarean)
+      totss = global_sum(work2, distrb_info, field_loc_center, tareas)
```
apcraig, line 1339: Looks like you could reuse work1 here rather than add new array, work2.

cicecore/cicedyn/general/ice_init.F90
```diff
@@ -260,6 +260,7 @@ subroutine input_data
         highfreq,       natmiter,        atmiter_conv,  calc_dragio,    &
         ustar_min,      emissivity,      iceruf,        iceruf_ocn,     &
         fbot_xfer_type, update_ocn_f,    l_mpond_fresh, tfrz_option,    &
+        saltflux_option, ice_ref_salinity, &
```
apcraig, line 263: ice_ref_salinity could be shifted left one space, align &.

cicecore/cicedyn/general/ice_init.F90
```diff
@@ -1414,6 +1419,12 @@ subroutine input_data
             write(nu_diag,*) subname//' WARNING:   For consistency, set tfrz_option = mushy'
          endif
       endif
+      if (ktherm == 1 .and. trim(saltflux_option) /= 'constant') then
```
apcraig, line 1422: should this be an error or just a warning that nobody will see?

cicecore/shared/ice_fileunits.F90
```diff
@@ -116,7 +116,11 @@ subroutine init_fileunits
          ice_IOUnitsInUse(ice_stdout) = .true. ! reserve unit 6
          ice_IOUnitsInUse(ice_stderr) = .true.
          if (nu_diag >= 1 .and. nu_diag <= ice_IOUnitsMaxUnit) &
-            ice_IOUnitsInUse(nu_diag) = .true. ! reserve unit nu_diag
+              ice_IOUnitsInUse(nu_diag) = .true. ! reserve unit nu_diag
+#ifdef CESMCOUPLED
+         ! CESM can have negative unit numbers.
+         if (nu_diag < 0) nu_diag_set = .true.
```
apcraig, line 122: Had a quick look at "nu_diag_set" usage and don't love it overall.  Can we put #ifdef CESMCOUPLED" around it everywhere, so it's clear it has nothing to do with CICE.

apcraig: including in the declaration in ice_fileunits.F90.

cicecore/shared/ice_fileunits.F90
```diff
@@ -239,7 +243,8 @@ subroutine release_all_fileunits
          call release_fileunit(nu_rst_pointer)
          call release_fileunit(nu_history)
          call release_fileunit(nu_hdr)
-         if (nu_diag /= ice_stdout) call release_fileunit(nu_diag)
+         ! CESM can have negative unit numbers
+         if (nu_diag > 0 .and. nu_diag /= ice_stdout) call release_fileunit(nu_diag)
```
apcraig, line 247: Maybe rather than change this logic, it should just be

#ifndef CESMCOUPLED
  if (nu_diag /= ice_stdout) call release_fileunit(nu_diag)
#endif

cicecore/shared/ice_fileunits.F90
apcraig: or could we get rid of nu_diag_set and use CESMCOUPLED cpp to "turn off" whatever CICE is doing to set/release nu_diag?  With CESMCOUPLED, if we can assume nu_diag is set by the coupler layer, we shouldn't need "nu_diag_set".

apcraig (commented):
I think we need to do some standalone CICE testing to make sure we understand what the results are going to be.  I think the log comparisons are going to fail because of the new diagnostics even if it's all bit-for-bit.  I can help with this if needed.

cicecore/cicedyn/analysis/ice_diagnostics.F90
dabail10: Maybe. I thought they needed to be independent.

cicecore/cicedyn/general/ice_init.F90
dabail10: Ok

cicecore/cicedyn/general/ice_init.F90
dabail10: I'm basically going with the example of tfrz_option here. It is not advisable to use ktherm = 1 and saltflux_option /= 'constant', but maybe people could do it if they wanted.

cicecore/shared/ice_fileunits.F90
dabail10: This is fine.

cicecore/shared/ice_fileunits.F90
dabail10: We still need to make sure not to release the negative file units here. However, I am fine with wrapping in a CESMCOUPLED.

cicecore/cicedyn/analysis/ice_diagnostics.F90
dabail10: In this code, we do not need work2, but in the print_points code we do. The array work2 is used in a number of places, so it is not really saving us memory.

apcraig: OK, it just looks like work2 was added in init_mass_diags just for the new salt diagnostic.  work2 must exist separately in print_points and elsewhere.

cicecore/cicedyn/general/ice_init.F90
apcraig: fine by me, was just asking.

dabail10:
I think this is a good idea.

apcraig:
Github Actions indicates the code is not compiling.  Recommend standalone testing.

dabail10:
I fixed the issue with nu_diag_set in ice_init.F90. Should compile now. Although, I don't think it will actually run, because it requires the new icepack submodule?

apcraig:
I was just noticing that it needs the updated Icepack.

dabail10:
I will do some standalone testing though.

apcraig:
You should be able to update Icepack on your branch and push to update the PR.

How about also changing the name of the test from salt to saltflux.  Thanks.

apcraig:
I think the Icepack update is incorrect.  GHActions is also failing.  You could try the following and see if it helps,

- mv icepack icepack.old
- git submodule update --init
- cd icepack
- git pull origin main
- git log (should be #8637a2d)
- cd ../
- git add icepack
- git commit -a "update icepack"
- git push ...

dabail10:
Hmm. I thought I had the right one. I will check it.

dabail10:
Sorry. I was pointing to ESCOMP instead of Consortium main for Icepack.

dabail10:
Ok. Updating icepack is creating some conflicts with the current Icepack hash.

apcraig:
I think Icepack is OK now.  Will run some tests just to be sure.

apcraig:
Full test suite on cheyenne is all bit-for-bit, 3 compilers.

PR #808, opened 2023-01-05
Fix for rare instabilities in basal stress.

kshedstrom:
- Don't do log with v_i negative and don't let sigma_i be zero.
- For SIS2, I was having issues with thin new ice, 100% of the first category.

## PR checklist
    Fix for rare basal stress instability.
- [x] Developer(s): 
    Kate Hedstrom
    ENTER INFORMATION HERE 
    - [ ] bit for bit
    - [ x] different at roundoff level
    - [ ] more substantial 
    - [ ] Yes
    - [ x] No
    - [ ] Yes
    - [ x] No
    - [ ] Yes
    - [x ] No
    - [ ] Yes
        - [ ] Yes
        - [ x] No 

### DIFF
--- a/cicecore/cicedyn/dynamics/ice_dyn_shared.F90
+++ b/cicecore/cicedyn/dynamics/ice_dyn_shared.F90
@@ -1484,7 +1484,11 @@
             v_i = v_i - m_i**2
 
             mu_i    = log(m_i/sqrt(c1 + v_i/m_i**2)) ! parameters for the log-normal
-            sigma_i = sqrt(log(c1 + v_i/m_i**2))
+            if (v_i > 0.0) then
+              sigma_i = max(sqrt(log(c1 + v_i/m_i**2)), puny)
+            else
+              sigma_i = puny
+            endif
 
             ! max thickness associated with percentile of log-normal PDF
             ! x_kmax=x997 was obtained from an optimization procedure (Dupont et al.)


dabail10:
I'm curious to know how often this occurs?

dabail10 (approved):
I think this looks fine. I'm curious to how many cases with the basalstress formulation will be impacted by this?

kshedstrom:
I had a simulation fail after two years of trouble-free running. That one failure is too many, right?

apcraig:
@JFLemieux73 @phil-blain any concerns about this change?

@kshedstrom can you clarify when answers are changed in general.  Does it change answers only when landfastice is turned on and in all landfastice cases in general (depending on v_i > 0)?

kshedstrom:
You would need to have the newer landfast ice on to change answers. It would only change in very rare circumstances, yes depending on v_i >= 0. It is hard but not impossible to get a v_i < 0.

JFLemieux73:
I just got back from holiday break. Let me chat with Fred and Dany about this.

dupontf:
Sorry Kate and JF, is it possible that `cicecore/cicedyn/dynamics/ice_dyn_shared.F90` at line 1484 is incorrect as it should be normalized by `n_cat`, as it intends to compute the variance from the square mean and mean squared? Please also remined me why the square sum is normalized by `max(acat(n), puny)` at line 1482... Thanks!

dupontf:
hum, sorry I will recall that last comment about normalization by `n_cat` and `acat(n)`. Checking the formulae, it all seems correct although the quadratic summation is approximated. So I guess we are down to a discretization issue with possible negative values in certain conditions...

dupontf:
Also, Kate, @kshedstrom can you document the thickness distribution you are using that is at fault here? Your correction looks fine to me.

kshedstrom:
@dupontf Sorry, I'm not actually testing this within CICE, but rather within SIS2. I am running a pan-Arctic domain with ncat_ice = 10 and the new landfast ice. I have the ridging from Icepack on. I don't remember what condition lead to v_i going negative, but having very thin ice all in the first category at 100% ice cover lead to v_i being zero.

dupontf:
Thanks Kate, I agree. As far as I can see the one category case is the only one capable of producing a zero v_i. In fact, I have not found cases for negative values (phew!) with more categories. Still, zero (or less due to round-off errors) is not something we want in the log formula, so correction welcome!

JFLemieux73:
Thanks Kate for finding this. Did you run the base suite tests to verify that all the tests are BFB except maybe the ones with seabed_stress_method = 'probabilistic'? 

@apcraig, am I right that the base suite includes tests for seabed_stress_method = 'probabilistic'?

JFLemieux73:
I notice that the Dupont paper is "in prep" in the documentation. This should be modified as it has been published. I could take care of that.

kshedstrom:
@JFLemieux73 I have not run any test suites or CICE6 at all ever, in fact. I assume others are more set up to do so than I am.

JFLemieux73:
Ok no problem. I will test it.

JFLemieux73:
Elizabeth had another idea to avoid the if statement. Would the following also work? 

v_i = max( (v_i - m_i**2), puny )

dupontf:
good point. Yes, that does it too, with less operations.

kshedstrom:
Yes, good idea.

dupontf:
For the record, using only one category and assuming equiprobable PDF between $h_1$ and $h_2$, the square variance should be $1/12 (h_2-h_1)^2$ if my calculation is correct, i.e., not 0. Just to illustrate that our variance calculation is appoximate, i.e., it is equivalent in this instance to assume a impulse PDF (i.e., zero variance). As you add more categories, the problem is less acute.

JFLemieux73:
This PR can be closed as it has been replaced by PR #810.

eclare108213:
replaced by PR https://github.com/CICE-Consortium/CICE/pull/810

PR #812, opened 2023-01-26
Some small CESM updates.

dabail10:

## PR checklist
Removes a comment about the OMP initialization not working in CESM and adds back in some fixes for restarting in CESM.
- [X] Developer(s): 
dabail10 (D. Bailey)
 https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#8afd0be23357b7160e1c9fb5a228c7a0b60c78e4
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

### DIFF
--- a/cicecore/cicedyn/analysis/ice_history.F90
+++ b/cicecore/cicedyn/analysis/ice_history.F90
@@ -81,6 +81,7 @@
       use ice_history_fsd, only: init_hist_fsd_2D, init_hist_fsd_3Df, &
           init_hist_fsd_4Df, f_afsd, f_afsdn
       use ice_restart_shared, only: restart
+      use ice_fileunits, only: goto_nml
 
       real (kind=dbl_kind), intent(in) :: &
          dt      ! time step
@@ -104,7 +105,9 @@
          cstr_got, cstr_gou, cstr_gov        ! mask area name for t, u, v ocn grid (go)
       character (len=25) :: &
          gridstr2D, gridstr                  ! temporary string names
-      character(len=char_len) :: description
+      character(len=char_len)      :: description
+      character(len=char_len_long) :: tmpstr2 ! for namelist check
+      character(len=char_len)      :: nml_name ! text namelist name
 
       character(len=*), parameter :: subname = '(init_hist)'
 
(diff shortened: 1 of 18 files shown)


DeniseWorthen:
@dabail10 could you make one small change that I need also? I found I needed the nu_diag_set parameter to be available because it is used now in the cap code, even w/o cesmcoupled being active. The change is
```
@@ -81,10 +80,8 @@ module ice_fileunits
       integer (kind=int_kind), public :: &
          nu_diag = ice_stdout  ! diagnostics output file, unit number may be overwritten

-#ifdef CESMCOUPLED
       logical (kind=log_kind), public :: &
          nu_diag_set = .false. ! flag to indicate whether nu_diag is already set
-#endif
```

dabail10:
Sounds good.

eclare108213:
> 
> ```
> -#ifdef CESMCOUPLED
>        logical (kind=log_kind), public :: &
>           nu_diag_set = .false. ! flag to indicate whether nu_diag is already set
> -#endif
> ```

Could this change adversely affect other codes?  Is this parameter used in any other codes?

DeniseWorthen:
@eclare108213 The nu_diag_set is only used by the cmeps NUOPC cap. It is false by default, so it shouldn't impact any other use case. However, if it is an issue, an alternative fix would be to wrap the code related to nu_diag_set in the cap with CESMCOUPLED. It's just a bit messier.

apcraig:
The nu_diag_set implementation should be refactored at some point, it's quite awkward as it stands now.

Why remove the comment about the OpenMP block that doesn't work in CESM?  Does it work now, if so, should the ifdef _OPENMP be removed?  We should have a comment there that indicates why that particular openmp loop is an ifdef block, don't you think?

dabail10:
It does work in CESM now. I'm not sure about the ifdef. I guess it is redundant? Or actually, we don't want this section to print when OpenMP is not active correct?

apcraig:
You're right, we do want the _OPENMP ifdef there.  I guess I thought it was related to CESM without looking at it.  I now remember that this was debugged and fixed a while ago, guess the comment should have been removed then.  thanks.

apcraig (approved):

PR #816, opened 2023-02-28
Add time_period_freq

dabail10:

## PR checklist
Adds the field time_period_freq to the history file metadata.
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#347689413d1b678ddb576befa3bcbbc710f46d9b
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
This is a requirement for CESM. Perhaps other groups would be interested in this as well?

### DIFF
--- a/cicecore/cicedyn/analysis/ice_history.F90
+++ b/cicecore/cicedyn/analysis/ice_history.F90
@@ -81,6 +81,7 @@
       use ice_history_fsd, only: init_hist_fsd_2D, init_hist_fsd_3Df, &
           init_hist_fsd_4Df, f_afsd, f_afsdn
       use ice_restart_shared, only: restart
+      use ice_fileunits, only: goto_nml
 
       real (kind=dbl_kind), intent(in) :: &
          dt      ! time step
@@ -104,7 +105,9 @@
          cstr_got, cstr_gou, cstr_gov        ! mask area name for t, u, v ocn grid (go)
       character (len=25) :: &
          gridstr2D, gridstr                  ! temporary string names
-      character(len=char_len) :: description
+      character(len=char_len)      :: description
+      character(len=char_len_long) :: tmpstr2 ! for namelist check
+      character(len=char_len)      :: nml_name ! text namelist name
 
       character(len=*), parameter :: subname = '(init_hist)'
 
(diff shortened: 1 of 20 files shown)


apcraig (commented):
Is there a reason this is only useful for CESMCOUPLED?  Why not implement generally?

dabail10:
I just wasn't sure if this was of interest for other groups. I guess I could do a namelist flag instead of CESMCOUPLED?

DeniseWorthen:
We don't use the PIO interface so this doesn't impact us. In that respect, the ifdef is not required. But the question I have is why is this needed? Doesn't the file name tell you what the time period is? If there is a good reason to have this added to the file, maybe it could be added to the io_netcdf interface also.

dabail10:
Fair question. The issue is that the post-processing should not necessarily trust the filename. Also, sometimes the files get concatenated and that information might be lost in the filename. Regardless, we were required to add this to our metadata. Good point that you do not use PIO. I can certainly add this to io_netcdf if others want it.

DeniseWorthen:
I see how that could be useful w/ concatenated files or post jobs. I would have no objection to having it added to the io_netcdf.

phil-blain:
A quick note, I took a quick look at the CF conventions https://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html and did not find a standard name for that information. I agree with Denise that it might make sense to also add it to the `io_netcdf` backend.

dabail10:
Happy to add it by default if people don't mind. @eclare108213 do you have a strong opinion here?

daveh150:
I think it is a good idea to add it to the io_netcdf as well. Good to have that info within the NetCDF metadata.

eclare108213:
I agree with the others:  put it in both netcdf and pio without the ifdef. There's no need for a namelist flag unless there's code in CICE that would specifically look for it in pre-existing files, which would not be backwards compatible -- that seems unlikely in this case.

dabail10:
Great. Thanks for all the feedback.

dabail10:
I updated the io_netcdf and removed CESMCOUPLED. I reran the io_suite comparing against the baselines. Everything is bfb. Not sure why the report_results.csh script is failing, but here are the test results:

609 measured results of 609 total results
609 of 609 tests PASSED
0 of 609 tests PENDING
0 of 609 tests MISSING data
0 of 609 tests FAILED

DeniseWorthen:
This is good on my end; compiles and writes the expected global attribute to the history file.

DeniseWorthen (approved):

apcraig (commented):
Thanks for doing that.  One more thought, I think time_period_freq needs a default like "unknown" either before the "select case" section or as a "case default".  "case default" could even be an abort if we think that is how it should behave.  We don't want select case to not select anything then try to write an uninitialized variable.

dabail10:
I have initialized time_period_freq to 'none' in the declaration. Then I check for 'none' when writing it out.

dabail10:
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#957ce31398e11dd7539b594a340aef3fffce8ae3

apcraig (approved):
Thanks.

PR #830, opened 2023-05-25
Fix for mesh check in CESM driver

dabail10:

## PR checklist
Fix mesh check when comparing CICE longitude to mesh longitude.
- [X] Developer(s): 
@jedwards4b (J. Edwards)
Only impacts NUOPC/CMEPS driver. Tested in CESM and UFS. Just changes the log output.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
It was something weird with the mod function where it was not detecting that 359.9 and 0. were the same.

### DIFF
--- a/cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90
+++ b/cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90
@@ -583,7 +583,7 @@ contains
     real(dbl_kind)               :: diff_lon
     real(dbl_kind)               :: diff_lat
     real(dbl_kind)               :: rad_to_deg
-    real(dbl_kind)               :: tmplon, eps_imesh
+    real(dbl_kind)               :: eps_imesh
     logical                      :: isPresent, isSet
     logical                      :: mask_error
     integer                      :: mask_internal
@@ -643,7 +643,7 @@ contains
                diff_lon = diff_lon - c360
              endif
              if (abs(diff_lon) > eps_imesh ) then
-                write(6,100)n,lonMesh(n),tmplon, diff_lon
+                write(6,100)n,lonMesh(n),lon(n), diff_lon
                 !call abort_ice(error_message=subname, file=__FILE__, line=__LINE__)
              end if
              diff_lat = abs(latMesh(n) - lat(n))
--- a/cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90
+++ b/cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90
@@ -638,7 +638,7 @@ contains
              lat(n) = tlat(i,j,iblk)*rad_to_deg
 
              ! error check differences between internally generated lons and those read in
-             diff_lon = abs(mod(lonMesh(n) - lon(n),360.0))
+             diff_lon = mod(abs(lonMesh(n) - lon(n)),360.0)
              if (diff_lon > c180) then
                diff_lon = diff_lon - c360
              endif
--- a/cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90
+++ b/cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90
@@ -559,7 +559,7 @@ contains
 
     ! Check CICE mesh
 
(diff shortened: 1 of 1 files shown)


dabail10:
It is working in CESM, but I don't understand why.

apcraig:
This looks sort of OK.  My suggestion though is that you get rid of the variable tmplon since you don't seem to be using it anymore.

But this type of calculation is always challenging.  What you have works if lon is always going to be greater than -720.  The other thing is that you run into problems if the lon differs by epsilon across 0/360.  Say one is 0.0001 and the other is 359.999.  Then the diff will be > 359 but in fact the result you want is 0.001 or something.  So for lon differences, you sometimes have to correct one value relative to the other.   Several ways to do it.  One way is to check whether abs(diff_lon) is greater than 180 and if so, adjust one value +- 360 and do the diff again and use the smaller diff.  Another option would be to correct one value to be 0,360 and the other value is corrected to -180, 180 if the first value is <90, to 0,360 if the first value is >90 and <270 and 180,540 if the first value is >270.  Just some thoughts.  Depends how important the diagnostic is.  And it looks like you're not aborting, so will anyone see the diagnostics or care?

This only affects the cmeps coupling layer, so it's largely up to you how to best implement.

cicecore/drivers/nuopc/cmeps/ice_mesh_mod.F90
```diff
@@ -641,7 +641,7 @@ subroutine ice_mesh_check(gcomp, ice_mesh, rc)
              if(tmplon < c0)tmplon = tmplon + c360
```
jedwards4b, line 641: tmplon can be removed

jedwards4b:
@apcraig My original PR did remove tmplon, I guess David reworked it?   We had some 18,000 lines of cice errors at the top of the cesm log, with this change it went to 0.  If you have a fix that better handles all of the edge cases please feel free to submit it.   I think we can put the abort back in if we get this right.   I don't think that we want to release a model that has a log beginning with 18,000 lines of "CICE: ERROR" messages.

apcraig:
So, a more general implementation might be something like

             diff_lon = (mod(c2*c360+lonMesh(n),c360) - mod(c2*c360+lon(n),c360))
             if (diff_lon > c180) then
               diff_lon = diff_lon - c360
             elseif (diff_lon < -c180) then
               diff_lon = diff_lon + c360
             endif
             if (abs(diff_lon) > eps_imesh ) then
                write(6,100)n,lonMesh(n),lon(n), diff_lon
                call abort_ice(error_message=subname, file=__FILE__, line=__LINE__)
             end if

I have not compiled or tested this.

dabail10:
I like this idea of checking diff_lon after it is computed. We don't even need the 2*360 in this case. Testing now.

jedwards4b (commented):
tmplon variable should be removed at line 586

dabail10:
Sorry. Caught that and issued another commit.

apcraig:
This scares me a bit.  I never trust mod of negative values.

   diff_lon = abs(mod(lonMesh(n) - lon(n),360.0))

I would write it

   diff_lon = mod(abs(lonMesh(n) - lon(n)),360.0)

dabail10:
Fixed

apcraig:
The abort is still off.  Will you update that later or do you want to fix that now?

dabail10:
Actually, that makes sense. We removed the abort since the 360 / 0 thing was causing a problem. I guess I can put that back in.

apcraig (approved):

dabail10:
@DeniseWorthen are you comfortable with this?

DeniseWorthen:
@dabail10 I saw this this weekend and was wondering if anything had changed since I had tested it. Should I test again?

dabail10:
I added back in the abort statements.

DeniseWorthen:
@dabail10 I've rechecked a couple of cases and this still works for us. Thanks.

PR #831, opened 2023-05-25
Add JRA55do to forcing

daveh150:

## PR checklist
    Add JRA55do forcing
- [x] Developer(s): 
    David Hebert
base_suite results
 419 measured results of 419 total results
419 of 419 tests PASSED
0 of 419 tests PENDING
0 of 419 tests MISSING data
0 of 419 tests FAILED

- How much do the PR code changes differ from the unmodified
-  code? 
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [x] Yes
        - [ ] No 

Added JRA55do option to ice_forcing.F90. Started from the JRA55 forcing and modified. 

In this JRA55do case I kept the JRA55do variable names, and also used individual variables for each variable instead of one large file for each year.  They are defined in the JRA55do subroutine.

### DIFF
--- a/cicecore/cicedyn/general/ice_forcing.F90
+++ b/cicecore/cicedyn/general/ice_forcing.F90
@@ -281,11 +281,14 @@
             file=__FILE__, line=__LINE__)
       endif
 
-      if (use_leap_years .and. (trim(atm_data_type) /= 'JRA55_gx1' .and. &
-                                trim(atm_data_type) /= 'JRA55_gx3' .and. &
-                                trim(atm_data_type) /= 'JRA55_tx1' .and. &
-                                trim(atm_data_type) /= 'hycom' .and. &
-                                trim(atm_data_type) /= 'box2001')) then
+      if (use_leap_years .and. (trim(atm_data_type) /= 'JRA55_gx1'   .and. &
+                                trim(atm_data_type) /= 'JRA55_gx3'   .and. &
+                                trim(atm_data_type) /= 'JRA55_tx1'   .and. &
+                                trim(atm_data_type) /= 'JRA55do_gx1' .and. &
+                                trim(atm_data_type) /= 'JRA55do_gx3' .and. &
+                                trim(atm_data_type) /= 'JRA55do_tx1' .and. &
+                                trim(atm_data_type) /= 'hycom'       .and. &
+                                trim(atm_data_type) /= 'box2001'))   then
          write(nu_diag,*) 'use_leap_years option is currently only supported for'
          write(nu_diag,*) 'JRA55, default , and box2001 atmospheric data'
          call abort_ice(error_message=subname, file=__FILE__, line=__LINE__)
@@ -308,6 +311,12 @@
          call JRA55_gx3_files(fyear)
       elseif (trim(atm_data_type) == 'JRA55_tx1') then
          call JRA55_tx1_files(fyear)
+      elseif (trim(atm_data_type) == 'JRA55do_gx1') then
+         call JRA55do_gx1_files(fyear)
+      elseif (trim(atm_data_type) == 'JRA55do_gx3') then
+         call JRA55do_gx3_files(fyear)
+      elseif (trim(atm_data_type) == 'JRA55do_tx1') then
+         call JRA55do_tx1_files(fyear)
       elseif (trim(atm_data_type) == 'hadgem') then
          call hadgem_files(fyear)
       elseif (trim(atm_data_type) == 'monthly') then
@@ -654,6 +663,12 @@
          call JRA55_data
       elseif (trim(atm_data_type) == 'JRA55_tx1') then
          call JRA55_data
+      elseif (trim(atm_data_type) == 'JRA55do_gx1') then
+         call JRA55do_data
+      elseif (trim(atm_data_type) == 'JRA55do_gx3') then
+         call JRA55do_data
+      elseif (trim(atm_data_type) == 'JRA55do_tx1') then
+         call JRA55do_data
       elseif (trim(atm_data_type) == 'hadgem') then
          call hadgem_data
       elseif (trim(atm_data_type) == 'monthly') then
@@ -1605,6 +1620,18 @@
          i = index(data_file,'.nc') - 5
          tmpname = data_file
          write(data_file,'(a,i4.4,a)') tmpname(1:i), yr, '.nc'
+      elseif (trim(atm_data_type) == 'JRA55do_gx1') then ! netcdf
+         i = index(data_file,'.nc') - 5
+         tmpname = data_file
+         write(data_file,'(a,i4.4,a)') tmpname(1:i), yr, '.nc'
+      elseif (trim(atm_data_type) == 'JRA55do_gx3') then ! netcdf
+         i = index(data_file,'.nc') - 5
+         tmpname = data_file
+         write(data_file,'(a,i4.4,a)') tmpname(1:i), yr, '.nc'
+      elseif (trim(atm_data_type) == 'JRA55do_tx1') then ! netcdf
+         i = index(data_file,'.nc') - 5
+         tmpname = data_file
+         write(data_file,'(a,i4.4,a)') tmpname(1:i), yr, '.nc'
       else                                     ! LANL/NCAR naming convention
          i = index(data_file,'.dat') - 5
          tmpname = data_file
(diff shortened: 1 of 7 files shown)


apcraig:
Several questions/comments.
- Is the plan to keep JRA55 and JRA55do longer term?  Which would be the recommended JRA55 forcing dataset?
- We should add some test cases for JRA55do.
- I think we need to update the input datasets at zenodo, propagate those datasets, and do some testing before merging this.

daveh150:
For the JRA55do set I wrote individual NetCDF files for each variable instead of one large NetCDF. The variable names are different and there are no common files between this and the JRA55 files. Also my thought in this case was having 7x 1.4GB(ish) files might be more manageable than 1x 10 GB (ish) file, particularly to download from Zenodo. Though I can quickly make one file with NetCDF operators.

 

Sorry if I confused the BFB, but I don’t think that will hold using JRA55do vs. ‘old’ JRA55 forcing. The BFB was with the base suite, with intent the addition ‘did no harm’ to the rest of the CICE code. 

 

Agree we should have a comparison of tests with the JRA55 vs. JRA55do. I asked Dave Bailey if he could facilitate that since I think he has JRA55 runs/stats to compare with.

eclare108213:
> For the JRA55do set I wrote individual NetCDF files for each variable instead of one large NetCDF. The variable names are different and there are no common files between this and the JRA55 files. Also my thought in this case was having 7x 1.4GB(ish) files might be more manageable than 1x 10 GB (ish) file, particularly to download from Zenodo. Though I can quickly make one file with NetCDF operators. 

I see.  Would most of the data itself be the same (although with different variable names), if the files were formatted the same way?  I'm not suggesting you do that, just trying to understand what's different and where the overlaps are.

> Sorry if I confused the BFB, but I don’t think that will hold using JRA55do vs. ‘old’ JRA55 forcing. The BFB was with the base suite, with intent the addition ‘did no harm’ to the rest of the CICE code. 

I see, thanks!

daveh150:
They should be formatted the same with same units, etc, just in individual files vs. one large NetCDF (with different variable names). I thought it might be good to use the JRA55do variable names for clarity instead of the ones we used for JRA55, which are ones I copied from other internal NRL forcing.

 

Variable names should be straight forward to update/change, then we could make one large file for 2005 keep the same JRA55 forcing code if that is preferred.  That way would not require CICE code changes, provide we are okay with the hardwired JRA55 filename.

apcraig:
Or have a set of inputs that specify the forcing filename and names of variables on the file.  The thing that does stand out in these changes is how much is hard-wired.  However, I don't know that we can actually build something that is truly general and if we can't, it's probably better to just do it as we have than try to build something complicated and general that doesn't work very well.  If we generalize, we might have to get into time and space interpolation too.  There are tools out there like the CESM data streams that might be useful.  Prescribed CICE mode already uses that, so there is some precedent in CICE, but it would require more CESM infrastructure to be migrated into CICE (prescribed ice mode is only run within the CESM coupled framework so the infrastructure is available outside CICE).

daveh150:
In the JRA55/JRA55do case, in hindsight it seems easiest to adopt a variable name convention then based on the atm_data_dir definition one can tell if it JRA55 or JRA55do. I can keep what we have now for variable names and update the new JRA55do forcing to use the same ones.

 

I think I can also update the existing JRA55_??_files subroutines to check for a atm_dat_type = jra55do, then modify the base ‘uwind’ file to be ‘JRA55do_’

 

Let me have a hack on this the next few days and see if it can be simplified.

 

Thanks,

Dave

daveh150:
I consolidated the JRA55_files and JRA55_data subroutines in an attempt to standardize. In each subroutine it will search 'atm_data_type' to determine the grid and JRA55 or JRA55do. The downside of this is we will need to standardize our current the JRA55 files. Currently they are (clearly not thinking about standard file naming!)

gx1: JRA55_03hr_forcing_2005.nc
gx3: JRA55_gx3_03hr_forcing_2005.nc
tx1: JRA55_03hr_forcing_tx1_2005.nc

I updated code to search for JRA55_grd_03hr_forcing_yyyy.nc

Here are some quick results from the log, they are similar which is encouraging. Dave Bailey is on leave until the 10th he will probably have more tests for comparison.

total_snw_volume_2005


total_ice_extent_2005

total_ice_volume_2005

apcraig:
The code mods look good.  Have you checked everything is still bit-for-bit for the old option?  The documentation needs to be updated, and we need to decide if we're going to support old and new JRA55 data or just one.  If just new, we should change the default, deprecate the old option, and run a full test suite since all answers will change.  We may also want to do a QC test.  If we want to support both datasets, then we should decide which is default and then add some tests for the other.   The new datasets also need to be migrated to zenodo in any case.  It may be OK to support both options in terms of managing the dataset.  Any site will probably just have one set of data.  Zenodo already has the older JRA55 dataset, and we have to add the new one anyway.  We can have two sets on some test machines, at least for gx3.

daveh150:
Running base_suite test locally passed all tests (I only have gnu compiler locally). Two tests initially failed because they could not find the JRA55 forcing with 'new' naming conventions. Making symlinks and re-running solved that issue.

344 measured results of 344 total results
344 of 344 tests PASSED
0 of 344 tests PENDING
0 of 344 tests MISSING data
0 of 344 tests FAILED

For reference the tests that initially failed are:
restart_tx1_40x4_dsectrobin_medium
smoke_gx1_24x1_medium_run90day_yi2008

dabail10:
I am currently running two 5 year tests. The code here is fine. I will have to work up some new distribution for the files. Note that the names of the gx1 files have to change to include the 'gx1' grid name.

dabail10 (approved):

eclare108213 (approved):

daveh150:
I can add JRA55do tests. As it stands now looks like we will need to add items like set_nml.jra55do_gx1, set_nml.jra55do_gx3, set_nml.jra55do.tx1.  Then it will be one of the 'sets' in a test suite.

daveh150:
I added a few jra55do test cases to the base_suite and quick_suite. in base_suite I added jra55do to each of our grids (gx3, gx1, tx1). There are corresponding setup_nml.jra55do_XXX, where XXX is grid version. I tested them on our local system and they ran successfully.

apcraig:
Can someone summarize the current strategy.  Is the old JRA55 still the default?   Is it now using the individual yearly files?  Do we need to download new datasets for both JRA55 and JRA55do on all our test platforms?  Is the old JRA55 forcing going to stay the default for a while?

apcraig:
I would like to run a full test suite on cheyenne before merging.  But don't want to do that until the PR is ready.

dabail10:
In case it is helpful, the run with the JRA55do passed the QC test versus the standard JRA55 case.

Dave
ice_thickness_gx1testb_minus_gx1testc
ice_thickness_gx1testc
ice_thickness_gx1testb

daveh150:
Thanks Tony for asking about the strategy. As it stands now, JRA55 is the default, but needs to have new forcing files downloaded to work with the updated forcing code. 

Quite honestly I'm not sure what the default should be. It wouldn't be too hard to make JRA55do the default if that is the better dataset, should be just a matter of updating the forcing definitions in the test suites. Any opinions?

I am struggling with if we actually need both JRA55 and JRA55do? For our testing purposes can (or should) we simply support the JRA55do? Maybe that is for future release/testing. 

Also, a note I just went to the JRA55do [website](https://climate.mri-jma.go.jp/pub/ocean/JRA55-do/), and they say JRA55 is going to terminate sometime this year. Doesn't mean we can't use it for our testing years 2005-2011, just that it is going to go away.

apcraig:
It may be a little late, but I think it's a bummer to have to download the old JRA55 data split into files in order to user the newer version of the code.  Every test machine and every user will have to do this as soon as they update CICE with this PR.  And it doesn't change anything in terms of forcing data or results (hopefully).  But it's kind of a big job to ask of the community.  One way around this is to deprecate the JRA55 dataset in favor of the JRA55do and say that you have to download the new dataset.  But that could be problematic for folks that have established testing and development with JRA55 and want to stick with that.  From the user perspective, the best thing would be to preserve the current JRA55 forcing option with the current files and to add the JRA55do files (in whatever format we prefer) and make the JRA55do the default if we think it's better.  And yes, the JRA55 is being discontinued.  If the QC is passes with JRA55 and JRA55do, and JRA55 is being discontinued, remind me why we're doing this now.  Are simulations improved somehow?

What we may need to do is think about what comes after JRA55.  Or maybe it's fine to have 10 years of JRA55 forcing for the forseeable future since it's not supposed to be used for science anyway.

daveh150:
I think one could get by with making symlinks instead of re-downloading the old JRA55 forcing. The only difference is the filename having the grid resolution in a consistent location. This avoids having hard-coded filenames in the code, or an archaic if-else tree. 

Your bigger question is a good one... why are we doing this? I think the intent was to give option to use JRA55do as better for forcing ocean-ice models by making some corrections to JRA55. If the benefit really isn't worth the effort, maybe as you suggest perhaps it is better if we spend effort on a supported forcing set. We typically use CFSR when we need climate extended data, but we don't do a whole lot with that.

apcraig:
See also https://github.com/CICE-Consortium/CICE/issues/840.

What's happening with the old JRA55 datasets?  The files just need to be renamed or linked?  OK, that's less of an issue then, I thought there was some files were split into individual years or something.  That's true for all the different resolutions for the old JRA55 dataset?

I think we need to decide what we're doing here and how we're doing it, what's going to be the default, are we supporting both, and how to let the community know what's happening.

If the JRA55do provides some advantages, then we should make it the default.  And we should support the old datasets at all resolutions and with a namelist switch, with at most a requirement to add soft links in the local inputdata area.  And maybe the CICE scripts can automatically create those links for users if the new files/links don't exist, I could help with that.  If the JRA55do does not provide benefits, I would close this PR.  The only caveat might be if we think the JRA55do dataset could be our standard forcing dataset for many years to come, then maybe it's a good idea to update to JRA55do now (while still providing support for the prior JRA55 version with relative ease of use).

dabail10:
Basically the JRA55do is corrected version of the JRA55 that is specially designed for ice-ocean simulations. The JRA55do is what is used in CESM and other modeling centers. The Tsujino et al. article that was on our inputdata page specifically refers to the JRA55do. While that change in forcing passes the QC test (above), I do think we should change to JRA55do as the default. This is more consistent with other uses. I agree we will have to support the original JRA55 forcing for older versions of the code. I have created Zenodo entries for the JRA55do by year only, except for the gx3. These code mods also require that the string 'gx1_' be added to the gx1 forcing. I have an updated version (by year only) on Zenodo and created symbolic links on cheyenne.

apcraig:
I'm OK with making the JRA55do the default.  Just to be clear, that will change ALL test results and we will have to download the new data to all test platforms.

How about if we support the older JRA55 dataset without the links, using the old filenames directly in the code.  I think this should be easy to do without having to create a bunch of redundant subroutines.  Also, why don't we remove the grid name from the "atm_data_type" while we're doing this.  What I envision is

atm_data_type = [JRA55,JRA55do,ncar,default,box2001,...]
atm_data_dir = ICE_MACHINE_INPUTDATA/CICE_data/forcing/gx1

and then the code would construct the filenames.  So much is already hardcoded this should not be difficult.  We could do the same for the NCAR_bulk option too.  Let me checkout this code and prototype that implementation and see how it looks.  Then we can discuss.  I much prefer to make the code work with the current datasets than have to create links for the existing datasets.

dabail10:
I like this idea. However, the old JRA55 data should be using the gx1 in the filenames. The gx3 and tx1 have the grid name in the file names and it would be more consistent.

apcraig:
I totally understand that it would be more consistent to rename the old gx1 JRA55 dataset to have gx1 in the name.  But that's not what we have now and I really don't like the idea of the community struggling when the JRA55 gx1 case suddenly doesn't work just because we changed the names, especially if it's not the default anymore.  Why add that complication?  I think we want things to be backwards compatible when it's possible.

dabail10:
Could we support both names?

apcraig:
> Could we support both names?

I think we could.  Let me see if I can make this work.

apcraig:
I have created a PR to the daveh150 jra55do branch to generalize the implementation a bit.  The implementation is now flexible to find JRA55 or JRA55do filenames of various formats and with less redundant subroutines.  That PR also updates this PR to a more recent version of Consortium main.  But there is still going to be a conflicting file based on the PR merge this morning.  I can help fix that if needed.  Or we can cancel this PR and open a new one as needed and I can help with that too if it's better.  I suggest we get the changes in for jra55do then have a separate PR to change the default to jra55do.

daveh150:
@apcraig if your branch implements  jra55do that does not have a merge conflict it sounds like it will be easier to just work from your repo for a PR, and we can delete this one. That seems like the cleanest route unless I'm missing something. Thank you Tony!

apcraig:
OK, will setup a clean PR and then close this one.  Testing that now.

Several other issues.  I think it's important to keep the inputdata usage as simple as possible and backwards compatible.  To me, that means living with the current forcing filenames.  They are not totally self consistent, but they are all unique and that's the main thing.  And the new implementation will be fully compatible with all the filenames as they exist today.   So, we should not worry about new file names for current files, creating links to rename filenames, or anything else like that.  It just confuses usage.  To that end, 

- I have updated the directory structure information on this page, https://github.com/CICE-Consortium/CICE/wiki/CICE-Input-Data to add the JRA55do files, that hadn't been done yet.  I also reordered the information slightly so it's consistent between grids.
- I have removed the JRA55 gx1 version 2 zenodo link from that same page.  These files are the same files as the original versions, just different names.   The new implementation will deal with the various flavors of filenames automatically.   Unfortunately, zenodo still shows there are two versions, https://zenodo.org/record/8118239.  Both should work for users and both are the same data (so that's slightly confusing).  Could we delete the zenodo version 2 page?
- I have removed some of the links that were added on cheyenne at /glade/p/cesm/pcwg_dev/CICE_data/forcing.  OK, sort of makes the files look a little nicer, but this is not how the files are provided to the community.  We should be duplicating what we provide to the community on cheyenne to make sure it works.  @dabail10, I don't have permissions to remove the JRA55_tx1 links, can you do that?  Also maybe check permissions in the pcwg_dev/CICE_data and Icepack_data directories on cheyenne.  Group is "ncar" and some of it is group write-able.  But there is a mix of permissions, just think about what it should be and maybe check that it's OK.

apcraig:
Closing, see #843

eclare108213:
@daveh150 thanks for all your work getting this started, and for comparing the old and new JRA55 data sets (@dabail10 too!).  It would be good to include links to those results in https://github.com/CICE-Consortium/CICE/pull/843.  I'll work on that bit -

Issue #832, opened 2023-06-06
Set l_fixed_area=T when C grid (and remap) is used


JFLemieux73:
The code should be modified so that l_fixed_area is always set to True wtih the C grid (and remap). See issues #792, #809 and #811 for details. 

An additional test should be added to the suite ( C grid, remap l_fixed_area=T).

JFLemieux73:
add variable in module rmap and set l_fixed_area=F or T whether B or C grid in remap init

JFLemieux73:
I have modified the code so that l_fixed_area is set in init_remap depending of the grid. 

The baseline tests assume that  l_fixed_area=.false. So at first, I wanted to verify that the end-2-end testing procedure is BFB if l_fixed_area = .false. whatever the grid. So I tested with 

if (grid_ice == 'CD' .or. grid_ice == 'C') then
         l_fixed_area = .false.
else
         l_fixed_area = .false.
endif

It is indeed BFB:

349 measured results of 349 total results
349 of 349 tests PASSED
0 of 349 tests PENDING
0 of 349 tests MISSING data
0 of 349 tests FAILED

apcraig:
See #833

Issue #809, opened 2023-01-12
Impact of l_fixed_area flag on C grid gx3/gx1 simulations


JFLemieux73:
C grid simulations with l_fixed_area=.false. (default) exhibit checkerboard noise (issue #792) and thicker ice (compared to B grid) in convergence zones (issue #702).

Setting l_fixed_area=.true. eliminates the checkerboard noise (issue #792). 

We want to see what is the impact of  l_fixed_area flag on ice thickness in gx3/gx1 simulations.

JFLemieux73:
I use the latest code with commit ID 0bf0fdcabd00df95d1028a66f71784b0b68f1178

I set up the experiments by doing:

./cice.setup -c CgridFixedArea -g gx3 -m ppp6 -e intel -p 8x1 -s diag1,buildincremental

As in issue #702 I use these settings:

ndte=1200
visc_method = 'avg_zeta'
elasticDamp = 0.12d0
deltaminEVP = 2e-9
deltaminVP = 2e-9
capping = sum

JFLemieux73:
Things are always more complicated than expected...The C grid run (dt=3600) with l_fixed_area=.true. just crashed. I get the following message:

(abort_ice) error = (diagnostic_abort)ERROR: negative area (ice)

whlipscomb:
@JFLemieux73, I'm sorry to hear that. I guess the documentation was right to warn that the scheme hasn't been fully tested.

When I got that error, I would write out a lot of diagnostics for the cell in question. Once I understood the geometry of the situation, I could compare it to the code logic and figure out why the logic failed.

Have you set bugcheck = .true. in ice_transport_remap.F90?  You might get some useful info. For instance, there's some code to check whether the sum of triangle areas across the edge is equal to the prescribed edgearea.  If these don't match, it wouldn't be surprising to see a negative area.

One possibility is that the small midpoint displacement, which is supposed to give exactly the right area, pushes the midpoint from one side of the edge to the other. If so, we might need some logic to handle that situation differently. At the end of the day, we'll have to decide if setting l_fixed_area = .true. is a useful quick fix, or more of a bag of worms.

JFLemieux73:
Thanks Bill for your help. I will investigate and keep you informed.

JFLemieux73:
[...]

PR #833, opened 2023-06-06
Modification of edge mask computation when l_fixed_area=T in horizontal remapping

JFLemieux73:
## PR checklist
   gx3, gx1 C grid simulations were previously crashing with horizontal remapping and  l_fixed_area=T. This modification corrects this problem. See issue #809 for details.
- [x] Developer(s): 
    @JFLemieux73 
    @eclare108213 

349 of 349 tests PASSED
0 of 349 tests PENDING
0 of 349 tests MISSING data
0 of 349 tests FAILED

    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    Another PR will follow and will add a test case. I will also modify the code so that l_fixed_area=T when the C grid (and remap) is used. See issue #832.
    - [ ] Yes
        - [x] Yes
        - [ ] No 

### DIFF
--- a/cicecore/cicedyn/general/ice_init.F90
+++ b/cicecore/cicedyn/general/ice_init.F90
@@ -2532,7 +2532,7 @@
       use ice_domain, only: nblocks, blocks_ice, halo_info
       use ice_domain_size, only: ncat, nilyr, nslyr, n_iso, n_aero, nfsd
       use ice_flux, only: sst, Tf, Tair, salinz, Tmltz
-      use ice_grid, only: tmask, ULON, TLAT, grid_ice, grid_average_X2Y
+      use ice_grid, only: tmask, umask, ULON, TLAT, grid_ice, grid_average_X2Y
       use ice_boundary, only: ice_HaloUpdate
       use ice_constants, only: field_loc_Nface, field_loc_Eface, field_type_scalar
       use ice_state, only: trcr_depend, aicen, trcrn, vicen, vsnon, &
@@ -2720,6 +2720,7 @@
                              ilo, ihi,            jlo, jhi,            &
                              iglob,               jglob,               &
                              ice_ic,              tmask(:,:,    iblk), &
+                             umask(:,:,    iblk), &
                              ULON (:,:,    iblk), &
                              TLAT (:,:,    iblk), &
                              Tair (:,:,    iblk), sst  (:,:,    iblk), &
@@ -2742,10 +2743,10 @@
 
       if (grid_ice == 'CD' .or. grid_ice == 'C') then
 
-         call grid_average_X2Y('S',uvel,'U',uvelN,'N')
-         call grid_average_X2Y('S',vvel,'U',vvelN,'N')
-         call grid_average_X2Y('S',uvel,'U',uvelE,'E')
-         call grid_average_X2Y('S',vvel,'U',vvelE,'E')
+         call grid_average_X2Y('A',uvel,'U',uvelN,'N')
+         call grid_average_X2Y('A',vvel,'U',vvelN,'N')
+         call grid_average_X2Y('A',uvel,'U',uvelE,'E')
+         call grid_average_X2Y('A',vvel,'U',vvelE,'E')
 
          ! Halo update on North, East faces
          call ice_HaloUpdate(uvelN, halo_info, &
@@ -2759,7 +2760,6 @@
                              field_loc_Eface, field_type_scalar)
 
       endif
-
 
       !-----------------------------------------------------------------
       ! compute aggregate ice state and open water area
@@ -2819,8 +2819,9 @@
                                 ilo, ihi, jlo, jhi, &
                                 iglob,    jglob,    &
                                 ice_ic,   tmask,    &
-                                ULON, &
-                                TLAT, &
+                                umask, &
+                                ULON,  &
+                                TLAT,  &
                                 Tair,     sst,  &
                                 Tf,       &
                                 salinz,   Tmltz, &
@@ -2845,7 +2846,8 @@
          ice_ic      ! method of ice cover initialization
 
       logical (kind=log_kind), dimension (nx_block,ny_block), intent(in) :: &
-         tmask      ! true for ice/ocean cells
+         tmask  , & ! true for ice/ocean cells
+         umask      ! for U points
 
       real (kind=dbl_kind), dimension (nx_block,ny_block), intent(in) :: &
          ULON   , & ! longitude of velocity pts (radians)
@@ -3293,13 +3295,19 @@
             domain_length = dxrect*cm_to_m*nx_global
             period        = c12*secday               ! 12 days rotational period
             max_vel       = pi*domain_length/period
+
             do j = 1, ny_block
             do i = 1, nx_block
 
-               uvel(i,j) =  c2*max_vel*(real(jglob(j), kind=dbl_kind) - p5) &
-                         / real(ny_global - 1, kind=dbl_kind) - max_vel
-               vvel(i,j) = -c2*max_vel*(real(iglob(i), kind=dbl_kind) - p5) &
-                         / real(nx_global - 1, kind=dbl_kind) + max_vel
+               if (umask(i,j)) then
+                  uvel(i,j) =  c2*max_vel*(real(jglob(j), kind=dbl_kind) - p5) &
+                            / real(ny_global - 1, kind=dbl_kind) - max_vel
+                  vvel(i,j) = -c2*max_vel*(real(iglob(i), kind=dbl_kind) - p5) &
+                            / real(nx_global - 1, kind=dbl_kind) + max_vel
+               else
+                  uvel(i,j) = c0
+                  vvel(i,j) = c0
+               endif
             enddo               ! j
             enddo               ! i
          else
--- a/cicecore/cicedyn/dynamics/ice_transport_driver.F90
+++ b/cicecore/cicedyn/dynamics/ice_transport_driver.F90
@@ -37,9 +37,6 @@
                      ! 'upwind' => 1st order donor cell scheme
                      ! 'remap' => remapping scheme
 
-      logical, parameter :: &
-         l_fixed_area = .false. ! if true, prescribe area flux across each edge
-
 ! NOTE: For remapping, hice and hsno are considered tracers.
 !       ntrace is not equal to ntrcr!
 
@@ -81,6 +78,7 @@
       use ice_state, only: trcr_depend
       use ice_timers, only: ice_timer_start, ice_timer_stop, timer_advect
       use ice_transport_remap, only: init_remap
+      use ice_grid, only: grid_ice
 
       integer (kind=int_kind) ::       &
          k, nt, nt1     ! tracer indices
@@ -236,7 +234,7 @@
       endif ! master_task
  1000 format (1x,a,2x,i6,2x,i6,2x,i4,4x,l4)
 
-      if (trim(advection)=='remap') call init_remap    ! grid quantities
+      if (trim(advection)=='remap') call init_remap (grid_ice) ! grid quantities
 
       call ice_timer_stop(timer_advect)  ! advection
 
@@ -545,7 +543,6 @@
           call horizontal_remap (dt,             ntrace,         &
                                  uvel   (:,:,:), vvel   (:,:,:), &
                                  aim  (:,:,:,:), trm(:,:,:,:,:), &
-                                 l_fixed_area,                   &
                                  tracer_type,    depend,         &
                                  has_dependents, integral_order, &
                                  l_dp_midpt,     grid_ice,       &
@@ -554,7 +551,6 @@
           call horizontal_remap (dt,             ntrace,         &
                                  uvel   (:,:,:), vvel   (:,:,:), &
                                  aim  (:,:,:,:), trm(:,:,:,:,:), &
-                                 l_fixed_area,                   &
                                  tracer_type,    depend,         &
                                  has_dependents, integral_order, &
                                  l_dp_midpt,     grid_ice)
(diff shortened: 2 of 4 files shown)


eclare108213:
Looks good!  I suggest that you go ahead and add a little documentation (a couple of sentences?) and a test case to this PR.

JFLemieux73:
Could you please merge this PR? I will add a test case and modifications to the doc in another PR as I would also like to make more modifications to the code so that l_fixed_area is set to true when the C grid is used. I have created an issue not to forget about this...

https://github.com/CICE-Consortium/CICE/issues/832

TillRasmussen (approved):
Looks fine.
One note. I suppose it is not bit for bit as it does not crash.

apcraig (commented):
I would prefer if the do j, do i loops were aligned with respect to indentation like the other double nested loops in this subroutine.  But a minor quibble.

apcraig:
What test suite was run?  I should probably run a full test suite including all the C-grid tests when ready.

Also, should a default change?  Should the C-grid always be running with l_fixed_area=T?

JFLemieux73:
Hi Tony,

I ran the end-to-end testing procedure described in section 3.3.6.4 of the doc:

https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_testing.html#end-to-end-testing-procedure

Yes l_fixed_area should always be true when using the C grid and I guess we should keep it set to false (default) for the B grid. I was not sure what is the best way to do that. I don't think I would add it in ice_in.

apcraig:
You can add the namelist to

configuration/scripts/options/set_nml.gridc

when the C grid is set with cice.setup, it'll always then have the settings in set_nml.gridc.  If someone manually changes a case and turns on the C grid, then it's up to them to set l_fixed_area.  This should also probably be documented with the C grid.

TillRasmussen:
> Hi Tony,
> 
> I ran the end-to-end testing procedure described in section 3.3.6.4 of the doc:
> 
> https://cice-consortium-cice.readthedocs.io/en/main/user_guide/ug_testing.html#end-to-end-testing-procedure
> 
> Yes l_fixed_area should always be true when using the C grid and I guess we should keep it set to false (default) for the B grid. I was not sure what is the best way to do that. I don't think I would add it in ice_in.

If this is the case then it should not be a hard coded parameter. Instead something like if c grid equal to true else if B grid equal to false else (cd grid I suppose) true/false

JFLemieux73:
I am done with the modifications to the code and the doc (for l_fixed_area and C-grid). Let me know if these are ok.

apcraig:
I'll run a full regression test suite on cheyenne including C-grid tests.  More soon.

apcraig:
I tested the modifications on cheyenne.  Mostly all OK.  The cgrid results are largely not bit-for-bit, as expected.  Otherwise, things are bit-for-bit.  The only problem is that a few c and cd tests are failing that were not before,

```
FAIL cheyenne_intel_smoke_gbox80_11x3_boxslotcyl_gridc_reprosum_run10day run -1 -1 -1
FAIL cheyenne_intel_smoke_gbox80_8x1_boxslotcyl_cmplogrest_gridc_reprosum_run10day_thread run -1 -1 -1
FAIL cheyenne_intel_smoke_gbox80_1x1_boxslotcyl_gridc run -1 -1 -1
```

The error looks like 

```
 New mass < 0, i, j =           9           9
 Old mass =   1.00000000000000     
 New mass = -0.530255960863720     
 Net transport =  -1.53025596086372     
 (print_state) negative area (open
 (print_state) istep1, my_task, i, j, iblk:           1           0           9
           9          10
 ...
 (abort_ice) error = (diagnostic_abort)ERROR: negative area (open water)
```

We probably need to understand this a bit better before merging.

apcraig:
Also, the full test results are here,

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#3d26d4cdb6089467b3c3a0866cb8f99fdc103f07

JFLemieux73:
Thanks Tony. This is bad news...this is the problem we were trying to fix. So basically the new code fails for these test cases while the current main code works? 

I tried the regular box tests and gx3 configuration before the PR but I never tried the boxslotcyl...

JFLemieux73:
I can reproduce the failure. I notice that for this test case: kdyn=-1. I just created an issue to investigate this:

https://github.com/CICE-Consortium/CICE/issues/836

eclare108213:
> I can reproduce the failure. I notice that for this test case: kdyn=-1.

If kdyn=-1, then the dynamics should be turned off and ice velocities ought to be zero, unless they are being initialized to nonzero values for some reason.  Is this a problem with the test configuration rather than the code?

JFLemieux73:
I don't know yet but for sure this is very different than the normal use of the model: velocities are initialized everywhere (even where there is no ice) and remain constant (in time) during the simulation.

JFLemieux73:
I think this is not really an issue and that the boxsltcyl should be modified. The failure happens where there is no ice...In all the other idealized test cases or more realistic ones (gx3, gx1) the transport would not be calculated because aiu<amin. I coded something quickly just for testing and set the departure points to be zero (same as having uvel=vvel=0) when aiu is lower than amin (1e-03).  It does not crash anymore. Here is the hi field after 13 days (288 time steps = default for boxsltcyl) for the C grid simulation (with l_fixed_area=T):

CT_hi_2005-01-13-00000

As a reference here it is for the B grid simulation (l_fixed_area=F):

BF_hi_2005-01-13-00000

JFLemieux73:
Maybe the C grid with l_fixed_area=T is more prone to these negative mass problems when there is no ice....but do we really want to investigate this further as this should never happen in normal simulations?

apcraig:
is there a simple fix for the boxslotcyl test on the C-grid?  Change initialization or something else simple that won't affect the solver/solutions in general?

eclare108213:
I think we need to look more closely at this test case, but I suspect that what it was intended to do is test the advection using constant ice velocities with the dynamics turned off.  The slotted cylinder ought to move in a predictable way, based on the ice velocities.  So if B grid is working and C grid isn't, then the masking must be acting differently?

JFLemieux73:
Ok I think I understand what is happening. It is not a problem with remapping nor l_fixed_area. 

There are two things that need to be modified in the code. 

First with this test case some velocities are defined on land. The code should use umask and set uvel,vvel in set_state_var as:

if (umask(i,j)) then
                  uvel(i,j) =  c2*max_vel*(real(jglob(j), kind=dbl_kind) - p5) &
                            / real(ny_global - 1, kind=dbl_kind) - max_vel
                  vvel(i,j) = -c2*max_vel*(real(iglob(i), kind=dbl_kind) - p5) &
                            / real(nx_global - 1, kind=dbl_kind) + max_vel
else
                  uvel(i,j) = c0
                  vvel(i,j) = c0
endif

By doing this uvel(79,79) = 0 which is consistent with the landmask. uvel(78,78) is initialized to 2.27. As uvel(78,79) = 0 (no slip), uvelE(78,79) should be close to 1.135. uvelE(78,79) is however 2.27. To fix this, call grid_average_X2Y('S',uvel,'U',uvelE,'E')...in init_state should be replaced by call grid_average_X2Y('A',uvel,'U',uvelE,'E'). Same idea for the other grid_average_X2Y. 

I tested this and there is no more crash. The test case runs fine for both C grid (l_fixed_area=T) and B grid (l_fixed_area=F).

apcraig:
Looking at ice_init, the code in question is,

 ```
     call bound_state (aicen,        &
                        vicen, vsnon, &
                        ntrcr, trcrn)

      if (grid_ice == 'CD' .or. grid_ice == 'C') then

         call grid_average_X2Y('S',uvel,'U',uvelN,'N')
         call grid_average_X2Y('S',vvel,'U',vvelN,'N')
         call grid_average_X2Y('S',uvel,'U',uvelE,'E')
         call grid_average_X2Y('S',vvel,'U',vvelE,'E')

         ! Halo update on North, East faces                                                                                                                   
         call ice_HaloUpdate(uvelN, halo_info, &
                             field_loc_Nface, field_type_scalar)
         call ice_HaloUpdate(vvelN, halo_info, &
                             field_loc_Nface, field_type_scalar)

         call ice_HaloUpdate(uvelE, halo_info, &
                             field_loc_Eface, field_type_scalar)
         call ice_HaloUpdate(vvelE, halo_info, &
                             field_loc_Eface, field_type_scalar)

      endif
```

This will only affect the initial condition of the C and CD configurations.  I assume it will change all answers for C and CD and make them ALL more correct?

**Comment by JFLemieux73** [comment]
Am I right that boxslotcyl is the only test case that has non zero initial velocities?

**Comment by apcraig** [comment]
> Am I right that boxslotcyl is the only test case that has non zero initial velocities?

I think that is true right now.  See the end of set_state_var in ice_init.F90.  So we should be OK for other test cases.  I will test again when updates are complete.

**Comment by JFLemieux73** [comment]
I thought I had written a comment here but I can't find it anymore...

Please do not merge this code yet. The model ran fine for 4 years with gx3. But I decided to also run gx1...unfortunately there was a crash after 10 months of simulation. I am investigating. See issue #809 for details.

**Comment by JFLemieux73** [comment]
Hi everyone,

Please have a look at issue #809. To solve the new failure with l_fixed_area=T will require more changes to the code. These changes will change answers on non-uniform grids (even for the B-grid...). I propose that we merge this PR first. I could then modify the code and run the QA testing to make sure everything is ok.

**Comment by apcraig** [comment]
I agree that a separate PR to modify the calculations for areafact makes sense.   I have read thru a bit of #809 and that plan all sounds reasonable.  If it's relatively easy to do the better solution ("solution 2"), then I think we should, but if it requires extra communication or storage at each timestep for additional grid metrics, then I think it can also be deferred.  I have also reviewed this PR and think it looks fine.  

Just to confirm, is this PR complete?  If so, I will run a quick test suite on cheyenne just to confirm everything is working then if others agree, we can merge and then @JFLemieux73 can create the next PR.

**Comment by JFLemieux73** [comment]
Thanks Tony. Yes this PR is complete.

File: cicecore/cicedyn/general/ice_init.F90 [hunk 1]
```diff
@@ -2742,10 +2743,10 @@ subroutine init_state
 
       if (grid_ice == 'CD' .or. grid_ice == 'C') then
 
-         call grid_average_X2Y('S',uvel,'U',uvelN,'N')
-         call grid_average_X2Y('S',vvel,'U',vvelN,'N')
-         call grid_average_X2Y('S',uvel,'U',uvelE,'E')
-         call grid_average_X2Y('S',vvel,'U',vvelE,'E')
+         call grid_average_X2Y('A',uvel,'U',uvelN,'N')
+         call grid_average_X2Y('A',vvel,'U',vvelN,'N')
+         call grid_average_X2Y('A',uvel,'U',uvelE,'E')
```
eclare108213, line 2748: Why is this change in the averaging type needed?

eclare108213 (changes requested):
The plan to merge this PR prior to making more invasive changes makes sense to me.  However, since the code will still not be working robustly, I think the documentation changes here should be reverted to the previous version.

JFLemieux73:
To make sure it is more robust I should also change this in transport_remap:

if (grid_ice == 'CD' .or. grid_ice == 'C') then
         l_fixed_area = .true.
      else
         l_fixed_area = .false.
      endif

It should  be for the moment:

if (grid_ice == 'CD' .or. grid_ice == 'C') then
         l_fixed_area = .false.
      else
         l_fixed_area = .false.
      endif

Do you agree?

cicecore/cicedyn/general/ice_init.F90
JFLemieux73: This is just for the initialization of boxslotcyl test. uvelE and vvelN are defined from the uvel, vvel values. uvel(78,79) is initialized to 0 because of the land mask (no slip) while uvel(78,78) is initialized to 2.27. As uvelE is interpolated from uvel, uvelE(78,79) should be 2.27/2 =1.135. This is obtained by changing the averaging type.

eclare108213:
I think you can leave l_fixed_area as true, since you're testing with it, but force other users to use upwind instead of remapping unless they manually change the code and so are aware of the issues.

cicecore/cicedyn/general/ice_init.F90
eclare108213: Ah, thank you.

eclare108213:
Any changes to the defaults will affect @apcraig's tests, so decide what you want users to have while further work is done.

JFLemieux73:
The current default is l_fixed_area = .false. for all the grids. It is simpler for the moment just to have it to false.  For the users we should just keep the warning that they should use upwind for the moment.

JFLemieux73:
Ok this PR is ready. Let me know if there are any issues with the tests. Thanks!

apcraig:
Before I restart testing, let me just ask a few questions.  

Why are we passing in "grid_ice" into init_remap and changing that interface?   We should be able to code "use ice_grid, only : grid_ice" in ice_transport_remap.F90, either in the module or subroutine init_remap.  This would simplify the changes.

If l_fixed_area is now a module variable in ice_transport_remap, then you could (should?) remove l_fixed_area from the locate_triangles subroutine since it's available as module data.

I assume the "if code" that defines l_fixed_area=.false. will be used in the next PR and made a bit more interesting?

Also, GHActions is currently broken, so we should ignore those failures for the time being.

@JFLemieux73, just confirm again if/when this is ready for testing.  Thanks!

JFLemieux73:
Good points. I have removed l_fixed_area from the locate_triangles subroutine. I added use ice_grid, only : grid_ice in ice_transport_remap module. 

The if code is there for the next PR...

I think it is ready to be tested.

apcraig:
Looks good, I'll start testing.

apcraig:
Full test suite on cheyenne looks fine.  All tests pass and are bit-for-bit with current main except boxslotcyl which all pass but with different results.

7728 measured results of 7728 total results
7671 of 7728 tests PASSED
0 of 7728 tests PENDING
0 of 7728 tests MISSING data
57 of 7728 tests FAILED

apcraig (approved):

apcraig:
Will merge in the morning unless anyone disagrees.

Issue #828, opened 2023-05-18
History file time axis


dabail10:
Heads up here. I wanted to start a discussion in particular with @apcraig and @eclare108213. Others are welcome to chime in. CESM is requiring that we change the default time axis to be the middle of the averaging interval instead of the end of the averaging interval. This is very easy to just average time_beg and time_end for each stream. My question for everyone is do we still want the functionality to write the time axis at the end of the averaging interval? That is, I could add a namelist option to control this. The options would be something like:

time_avg_axis = 'begin'
time_avg_axis = 'middle'
time_avg_axis = 'end'

Then you would use time_beg, (time_beg+time_end)*0.5, or time_end (currently the default). Note that the filenames would not change. This only impacts the time axis in the netCDF file.

phil-blain:
Hi @dabail10, just to be clear, the change would be only under `hist_avg=.true.`, right? Because for snapshots, we would want the time axis to stay as is... (and the same for fields that are always snapshots even when averaging is active, like internal stresses if I remember correctly?)

dabail10:
That is correct. Also, since I changed hist_avg to an array, we can control this for each stream.

dabail10:
Currently in the monthly and daily. The filenames do not have information about where in the averaging interval it is. It just says 0001-01 and 0001-01-01. Now, the hourly is trickier. So that it would be 0001-01-01-01800 instead of 03600. However, for simplicity, I would not change this and leave it at the end of the averaging period for the seconds in the filename.

apcraig:
I would say if the filename is going to change, then there has to be an option to choose the old or new method.  If the internal time axis is going to change, then there has to be an option to choose the old or new method (as you propose).  Adding additional metadata or even axis information, like time bounds, should be fine to do in general, but it sounds like that's already there.

So I think what you're proposing is to leave the filenames alone and add new options for where in the average period the time axis is defined.  That's OK for me.  We definitely wants backwards compatibility.  If you decide to do something similar for filenames, then I think we need it to be controlled by a namelist too and independent of the time axis.

DeniseWorthen:
Regarding the filename change (since I brought up the MOM6 case), I think on our end the preference would be that the filename did not change, as you proposed. Generally they like the timestamp to reference the end time of the average ("forecast valid at 6Z"). On the workflow end, they rename the MOM6 output to get this ending time for the filename.

dabail10:
[...]

PR #839, opened 2023-06-21
Namelist option for time axis position.

dabail10:

## PR checklist
Added a namelist option for time axis position.
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#2efc26cddef840801d5ed1e5322e6502f1120683
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 
This is a first attempt to address the need for issue #828 . The default behavior is hist_time_axis = 'end' (backward compatible) and is only used when hist_avg is true.

### DIFF
--- a/cicecore/cicedyn/general/ice_init.F90
+++ b/cicecore/cicedyn/general/ice_init.F90
@@ -1575,7 +1575,8 @@
       endif
 
       if(trim(hist_time_axis) /= 'begin' .and. trim(hist_time_axis) /= 'middle' .and. trim(hist_time_axis) /= 'end') then
-         write (nu_diag,*) subname//' WARNING: hist_time_axis set to default end '
+         write (nu_diag,*) subname//' ERROR: hist_time_axis value not valid = '//trim(hist_time_axis)
+         abort_list = trim(abort_list)//":29"
       endif
 
       if(dumpfreq_base /= 'init' .and. dumpfreq_base /= 'zero') then
--- a/cicecore/cicedyn/general/ice_init.F90
+++ b/cicecore/cicedyn/general/ice_init.F90
@@ -325,7 +325,7 @@
       histfreq_base = 'zero' ! output frequency reference date
       hist_avg(:) = .true.   ! if true, write time-averages (not snapshots)
       history_format = 'default' ! history file format
-      hist_time_axis = 'end' ! History file time axis interval position
+      hist_time_axis = 'end' ! History file time axis averaging interval position
 
       history_dir  = './'    ! write to executable dir for default
       history_file = 'iceh'  ! history file name prefix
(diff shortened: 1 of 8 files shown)


eclare108213 (changes requested):
Wherever it appears, it would be helpful to add into the comments what this is begin/middle/end of: the history averaging interval.   Is it only used for averaging?

dabail10:
This is only specifying where in the averaging interval the time axis is based. The default currently has the timestep at the end of the averaging interval. I will add some more comments in the code.

apcraig:
Should there be a check that hist_time_axis is set to begin, middle, or end only somewhere, maybe during initialization?  Otherwise, someone that uses Begin or niddle will end up with "end".

dabail10:
The check is a good idea. I'm not sure about the netCDF attribute. This impacts the 'time' variable on the nextCDF file which has units of days since. So, for example, the January mean of year one is currently 31 days since 0001-01-01 00000. With the "middle" option this becomes 15.5 days since 0001-01-01 00000.

apcraig:
In terms of attribute, I was thinking more of a global attribute, just for documentation.  I guess it could also be a variable attribute associated with the time variable too.  But something like (time_axis_phase = "end";) would be the attribute.  Someone could query that and then, for example, shift the time axis based on the value of that attribute when plotting the data.  I don't think it should affect the time axis or overall implementation and/or general use.

dabail10:
Also, the tests above were the io_suite with intel and gnu on cheyenne.

apcraig:
OK, good on the testing.  I just looked quick and saw <100 test results, assumed it was the base_suite.  Thanks.

cicecore/cicedyn/general/ice_init.F90
```diff
@@ -1570,6 +1574,10 @@ subroutine input_data
          abort_list = trim(abort_list)//":24"
       endif
 
+      if(trim(hist_time_axis) /= 'begin' .and. trim(hist_time_axis) /= 'middle' .and. trim(hist_time_axis) /= 'end') then
+         write (nu_diag,*) subname//' WARNING: hist_time_axis set to default end '
```
apcraig, line 1578: I would abort here.  Change to something like

```
         write (nu_diag,*) subname//' ERROR: hist_time_axis value not valid = '//trim(hist_time_axis)
         abort_list = trim(abort_list)//":24"
```

Instead of 24, give it a unique number, can quickly great for abort_list to find one.

apcraig (approved):
Looks good, thanks!

dabail10:
@eclare108213 does this address your concerns above?

eclare108213 (approved):

PR #845, opened 2023-07-14
Change default forcing to JRA55do, changes answers

apcraig:
## PR checklist
    Change default forcing to JRA55do, changes answers
- [X] Developer(s): 
    apcraig, dbailey, davidh
    full test suite run on cheyenne, all tests pass, most tests change answers (box cases and unit tests generally don't).  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#f9d3002c86e11ca18b06382fc2d0676c9a945223
    - [ ] bit for bit
    - [X] more substantial, should be same climate via QC testing
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes, changes forcing default and adds JRA55 tests
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

This changes answers and all users will have to update the input data forcing to add JRA55do.


apcraig:
I need to update the github actions testing, need to pull the new dataset.

apcraig:
If we're going to do this, lets get this change in.  I'll need to update several machines with the new input data and we want to make this change well in advance of the release.  gx3, gx1, tx1 results will change.  If we're not going to do this, then lets just close the PR, and we can support the JR55do as an option.  Anyone changed their mind about moving this forward?  Anyone want to review the PR?

eclare108213:
See discussion and results at https://github.com/CICE-Consortium/CICE/pull/831 and https://github.com/CICE-Consortium/CICE/pull/843

eclare108213:
My inclination is to not make JRA55do the default, but to support it as an option, so that users don't have to download the data if the old JRA55 is sufficient for their purposes.  At some point in the future we might decide to make it the default, but if JRA55 is going away completely, perhaps we should wait for its replacement.  @dabail10 please weigh in.

daveh150:
I agree that to minimize community disruption making JRA55do an option, and not the default, makes sense. We could have set_nml options for JRA55do. It is worth while to add a separate small test suite like a 'quick jra55do' for reference?

dabail10:
This is a bit challenging. In a sense the original JRA55 datasets were a mistake. Most groups who use it do use the JRA55do which is specifically modified for running ice-ocean simulations. We use the JRA55do in CESM and I know a number of other groups who do this. So, in that sense I would vote for the JRA55do as the default. That's just my $0.02.

eclare108213:
There are strong arguments on both sides, and so I'd like other team members to weigh in.  @CICE-Consortium/devteam 
The quest is whether to change from JRA55 to JRA55do as the default forcing in our releases.

PROs:  
- JRA55do is specifically modified for running ice-ocean simulations
- many groups are using it

CONs:  
- all users would have to download the new data set
- JRA55* will not continue to be supported and will likely be replaced with a different product

TillRasmussen:
From DMI side there is no strong opinion about this. We primarily use ECMWF data and only use this data set for test when we push to the main branch.

dupontf:
Same here, I would not change this just for our own testing. As long as there is a disclaimer that the dataset is provided for testing only and not scientific studies, it should be clear enough.

eclare108213:
I think that making JRA55do the default forcing is too much of a disruption for not much gain.  We do not recommend that users run uncoupled/standalone production simulations.  If they want to do that, then they can set the forcing to JRA55do.  The Consortium would only be using it for our testing.  So my recommendation is to close this PR.  If we end up needing to make JRA55do the default, then we can reopen it or make a new one.

apcraig:
We can consider this anytime in the future.  For now, we're leaving JRA55 forcing as is with JRA55do an option.

PR #849, opened 2023-07-26
More accurate calculation of areafact in remapping

JFLemieux73:
Hi everyone,

@apcraig please do not merge this...this is for the moment just for review. @eclare108213, @apcraig, @TillRasmussen and @dabail10 feel free to give me your comments. 

This is solution1. This could be temporary if we decide to go with the more complicated solution2 later. 

I have tested solution1 in gx3 and gx1 simulations (C-grid, l_fixed_area=T). These simulations ran fine for 4 years. Looks good.

@whlipscomb, I have modified the calculations for areafact. I am however not sure if areafac_l, areafac_r and areafac_c in the l_fixed_area section should be modified. I have added the comment: ! jlem should this be modified? where I am not sure what to do. Please let me know. 

For the moment the code calculates:

areafac_c(i,j) = p5*(areafac_l(i,j) + areafac_r(i,j))
areafacS (i,j) = areafac_r(i,j) + areafac_l(i,j)

This is redundant. I will remove one array after I get your comments.

### DIFF
--- a/cicecore/cicedyn/dynamics/ice_transport_remap.F90
+++ b/cicecore/cicedyn/dynamics/ice_transport_remap.F90
@@ -317,7 +317,7 @@
       !-------------------------------------------------------------------
 
       if (grid_ice == 'CD' .or. grid_ice == 'C') then
-         l_fixed_area = .false. !jlem temporary
+         l_fixed_area = .true.
       else
          l_fixed_area = .false.
       endif
--- a/cicecore/cicedyn/dynamics/ice_transport_remap.F90
+++ b/cicecore/cicedyn/dynamics/ice_transport_remap.F90
@@ -2553,7 +2553,7 @@
             areafact(i,j,ng) = areatp
             !areafact(i,j,ng) = areafac_c(i,j)
 
-         elseif (ydl <= c0 .and. ydr <= c0 .and. ydm < c0) then
+         elseif (ydl <= c0 .and. ydr <= c0 .and. ydm <= c0) then ! Bill I also changed ydm < to <=
 
             ! compute areafac for quadrilateral cl-cr-dr-dl
             xpm = p25*(xcl + xcr + xdr + xdl)
@@ -2601,7 +2601,7 @@
             areafact(i,j,ng) = areatp
             !areafact(i,j,ng) = areafac_c(i,j)
 
-         elseif (ydl <= c0 .and. ydr <= c0 .and. ydm >= c0) then  ! rare
+         elseif (ydl <= c0 .and. ydr <= c0 .and. ydm > c0) then  ! rare ! Bill I also changed ydm >= to >
 
             ! compute areafac for quadrilateral cl-cr-dr-dl
             xpm = p25*(xcl + xcr + xdr + xdl)
@@ -2653,8 +2653,8 @@
          ! For these cases, one triangle is given the area factor associated
          !  with the adjacent corner, to avoid rare negative masses on curved grids.
 
-         elseif (ydl >= c0 .and. ydr < c0 .and. xic >= c0  &
-                                          .and. ydm >= c0) then
+         elseif (ydl > c0 .and. ydr < c0 .and. xic >= c0  &
+                                         .and. ydm >= c0) then
 
             ! compute areafac for left triangle cl-dl-ic
             xpm = p333*(xcl + xdl + xic)
@@ -2704,8 +2704,8 @@
             areafact(i,j,ng) = -areatp
             !areafact(i,j,ng) = -areafac_c(i,j)
 
-         elseif (ydl >= c0 .and. ydr < c0 .and. xic >= c0  &
-                                          .and. ydm < c0 ) then  ! less common
+         elseif (ydl > c0 .and. ydr < c0 .and. xic >= c0  &
+                                         .and. ydm < c0 ) then  ! less common
 
             ! compute areafac for left triangle cl-dl-ic
             xpm = p333*(xcl + xdl + xic)
@@ -2755,8 +2755,8 @@
             areafact(i,j,ng) = areatp
             !areafact(i,j,ng) = areafac_c(i,j)
 
-         elseif (ydl >= c0 .and. ydr < c0 .and. xic < c0   &
-                                          .and. ydm < c0) then
+         elseif (ydl > c0 .and. ydr < c0 .and. xic < c0   &
+                                         .and. ydm < c0) then
 
             ! compute areafac for right triangle ic-cr-dr
             xpm = p333*(xic + xcr + xdr)
@@ -2806,8 +2806,8 @@
             areafact(i,j,ng) = areatp
             !areafact(i,j,ng) = areafac_c(i,j)
 
-         elseif (ydl >= c0 .and. ydr < c0 .and. xic <  c0  &
-                                          .and. ydm >= c0) then  ! less common
+         elseif (ydl > c0 .and. ydr < c0 .and. xic <  c0  &
+                                         .and. ydm >= c0) then  ! less common
 
             ! compute areafac for right triangle ic-cr-dr
             xpm = p333*(xic + xcr + xdr)
@@ -2857,8 +2857,8 @@
             areafact(i,j,ng) = -areatp
             !areafact(i,j,ng) = -areafac_c(i,j)
 
-         elseif (ydl < c0 .and. ydr >= c0 .and. xic <  c0  &
-                                          .and. ydm >= c0) then
+         elseif (ydl < c0 .and. ydr > c0 .and. xic <  c0  &
+                                         .and. ydm >= c0) then
 
             ! compute areafac for right triangle ic-cr-dr
             xpm = p333*(xic + xcr + xdr)
@@ -2908,8 +2908,8 @@
             areafact(i,j,ng) = -areatp
             !areafact(i,j,ng) = -areafac_c(i,j)
 
-         elseif (ydl < c0 .and. ydr >= c0 .and. xic < c0  &
-                                          .and. ydm < c0) then ! less common
+         elseif (ydl < c0 .and. ydr > c0 .and. xic < c0  &
+                                         .and. ydm < c0) then ! less common
 
             ! compute areafac for right triangle ic-cr-dr
             xpm = p333*(xic + xcr + xdr)
@@ -2959,8 +2959,8 @@
             areafact(i,j,ng) = areatp
             !areafact(i,j,ng) = areafac_c(i,j)
 
-         elseif (ydl < c0 .and. ydr >= c0 .and. xic >= c0  &
-                                          .and. ydm <  c0) then
+         elseif (ydl < c0 .and. ydr > c0 .and. xic >= c0  &
+                                         .and. ydm <  c0) then
 
             ! compute areafac for left triangle cl-dl-ic
             xpm = p333*(xcl + xdl + xic)
@@ -3010,8 +3010,8 @@
             areafact(i,j,ng) = areatp
             !areafact(i,j,ng) = areafac_c(i,j)
 
-         elseif (ydl < c0 .and. ydr >= c0 .and. xic >= c0   &
-                                          .and. ydm >= c0) then  ! less common
+         elseif (ydl < c0 .and. ydr > c0 .and. xic >= c0   &
+                                         .and. ydm >= c0) then  ! less common
 
             ! compute areafac for left triangle cl-dl-ic
             xpm = p333*(xcl + xdl + xic)
--- a/cicecore/cicedyn/dynamics/ice_transport_remap.F90
+++ b/cicecore/cicedyn/dynamics/ice_transport_remap.F90
@@ -1762,6 +1762,8 @@
          ishift_br, jshift_br , & ! i,j indices of BR cell relative to edge
          ishift_tc, jshift_tc , & ! i,j indices of TC cell relative to edge
          ishift_bc, jshift_bc , & ! i,j indices of BC cell relative to edge
+         is_l, js_l           , & ! index shifts for TL1 and BL2 for tri area consistency 
+         is_r, js_r           , & ! index shifts for TR1 and BR2 for tri area consistency 
          ise_tl, jse_tl       , & ! i,j of TL other edge relative to edge
          ise_bl, jse_bl       , & ! i,j of BL other edge relative to edge
          ise_tr, jse_tr       , & ! i,j of TR other edge relative to edge
@@ -1906,6 +1908,13 @@
          ishift_bc =  0
          jshift_bc =  0
 
+         ! index shifts for TL1, BL2, TR1 and BR2 for triangle area consistency 
+
+         is_l = -1
+         js_l =  0
+         is_r = +1
+         js_r =  0
+
          ! index shifts for neighbor east edges
 
          ise_tl = -1
@@ -1949,6 +1958,13 @@
          jshift_tc =  0
          ishift_bc =  0
          jshift_bc =  0
+
+         ! index shifts for TL1, BL2, TR1 and BR2 for triangle area consistency 
+
+         is_l =  0
+         js_l = +1
+         is_r =  0
+         js_r = -1
 
          ! index shifts for neighbor north edges
 
@@ -2128,7 +2144,7 @@
             yp    (i,j,3,ng) = yic
             iflux   (i,j,ng) = i + ishift_tl
             jflux   (i,j,ng) = j + jshift_tl
-            areafact(i,j,ng) = areafac_ce(i+ise_tl,j+jse_tl)
+            areafact(i,j,ng) = areafac_c(i+is_l,j+js_l)
 
             ! BL1 (group 3)
 
@@ -2169,7 +2185,7 @@
             yp    (i,j,3,ng) = ydl
             iflux   (i,j,ng) = i + ishift_bl
             jflux   (i,j,ng) = j + jshift_bl
-            areafact(i,j,ng) = -areafac_ce(i+ise_bl,j+jse_bl)
+            areafact(i,j,ng) = -areafac_c(i+is_l,j+js_l)
 
          endif                  ! TL and BL triangles
 
@@ -2221,7 +2237,7 @@
             yp    (i,j,3,ng) = ydr
             iflux   (i,j,ng) = i + ishift_tr
             jflux   (i,j,ng) = j + jshift_tr
-            areafact(i,j,ng) = areafac_ce(i+ise_tr,j+jse_tr)
+            areafact(i,j,ng) = areafac_c(i+is_r,j+js_r)
 
             ! BR1 (group 3)
 
@@ -2262,7 +2278,7 @@
             yp    (i,j,3,ng) = yic
             iflux   (i,j,ng) = i + ishift_br
             jflux   (i,j,ng) = j + jshift_br
-            areafact(i,j,ng) = -areafac_ce(i+ise_br,j+jse_br)
+            areafact(i,j,ng) = -areafac_c(i+is_r,j+js_r)
 
          endif                  ! TR and BR triangles
 
--- a/cicecore/cicedyn/dynamics/ice_transport_remap.F90
+++ b/cicecore/cicedyn/dynamics/ice_transport_remap.F90
@@ -2375,7 +2375,7 @@
                endif
                yicr = c0
 
-            elseif (xic < c0) then  ! fix ICL = IC
+            elseif (xic < c0 .and. xic > xcl) then  ! fix ICL = IC ! Bill please check this
 
                xicl = xic
                yicl = yic
@@ -2410,7 +2410,7 @@
                endif
                yicr = c0
 
-            elseif (xic >= c0) then  ! fix ICR = IR
+            elseif (xic >= c0 .and. xic < xcr) then  ! fix ICR = IR ! Bill please check this
 
                xicr = xic
                yicr = yic
@@ -3101,7 +3101,7 @@
                                      (xp(i,j,3,ng)-xp(i,j,1,ng)) )   &
                                    * areafact(i,j,ng)
 
-            if (abs(triarea(i,j,ng)) < eps16*areafac_c(i,j)) then ! jf ici
+            if (abs(triarea(i,j,ng)) < eps16*areafac_c(i,j)) then ! Bill please check this
                triarea(i,j,ng) = c0
             else
                icells(ng) = icells(ng) + 1
@@ -3120,7 +3120,7 @@
             do ij = 1, icellsd
                i = indxid(ij)
                j = indxjd(ij)
-               if (abs(areasum(i,j) - edgearea(i,j)) > eps13*areafac_c(i,j)) then ! jf ici
+               if (abs(areasum(i,j) - edgearea(i,j)) > eps13*areafac_c(i,j)) then ! Bill please check this
                   write(nu_diag,*) ''
                   write(nu_diag,*) 'Areas do not add up: m, i, j, edge =',   &
                            my_task, i, j, trim(edge)
--- a/cicecore/cicedyn/dynamics/ice_transport_remap.F90
+++ b/cicecore/cicedyn/dynamics/ice_transport_remap.F90
@@ -63,7 +63,7 @@
                       ! if false, area flux is determined internally
                       ! and is passed out
 
-      logical (kind=log_kind), parameter :: bugcheck = .false.
+      logical (kind=log_kind), parameter :: bugcheck = .true.
 
 !=======================================================================
 ! Here is some information about how the incremental remapping scheme
@@ -1756,7 +1756,11 @@
          ishift_tr, jshift_tr , & ! i,j indices of TR cell relative to edge
          ishift_br, jshift_br , & ! i,j indices of BR cell relative to edge
          ishift_tc, jshift_tc , & ! i,j indices of TC cell relative to edge
-         ishift_bc, jshift_bc     ! i,j indices of BC cell relative to edge
+         ishift_bc, jshift_bc , & ! i,j indices of BC cell relative to edge
+         ise_tl, jse_tl       , & ! i,j of TL other edge relative to edge
+         ise_bl, jse_bl       , & ! i,j of BL other edge relative to edge
+         ise_tr, jse_tr       , & ! i,j of TR other edge relative to edge
+         ise_br, jse_br           ! i,j of BR other edge relative to edge
 
       integer (kind=int_kind) ::   &
          icellsd          ! number of cells where departure area > 0.
@@ -1767,12 +1771,12 @@
 
       real (kind=dbl_kind), dimension(nx_block,ny_block) ::   &
          dx, dy       , & ! scaled departure points
-         areafac_l    , & ! area scale factor at left corner
-         areafac_r    , & ! area scale factor at right corner
          areafac_c    , & ! (areafac_l + areafac_r)/2
-         areafacD         ! Difference of areafac_l and areafac_r
+         areafac_ce       ! areafac_c on other edge
 
       real (kind=dbl_kind) ::   &
+         areafac_l    , & ! area scale factor at left corner
+         areafac_r    , & ! area scale factor at right corner
          xcl, ycl     , & ! coordinates of left corner point
                           ! (relative to midpoint of edge)
          xdl, ydl     , & ! left departure point
@@ -1791,9 +1795,6 @@
          area1, area2 , & ! temporary triangle areas
          area3, area4 , & !
          area_c       , & ! center polygon area
-         areatp       , & ! temporary area calculation
-         areatplone   , & ! temporary area calculation for lone triangle
-         xpm          , & ! mean value of x points
          puny         , & !
          w1, w2           ! work variables
 
@@ -1863,10 +1864,9 @@
       if (icepack_warnings_aborted()) call abort_ice(error_message=subname, &
          file=__FILE__, line=__LINE__)
 
-      areafac_l(:,:) = c0
-      areafac_r(:,:) = c0
-      areafac_c(:,:) = c0
-      areafacD(:,:) = c0
+      areafac_c(:,:)  = c0
+      areafac_ce(:,:) = c0
+      
       do ng = 1, ngroups
          do j = 1, ny_block
          do i = 1, nx_block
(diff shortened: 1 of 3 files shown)


apcraig:
This seems OK to me, don't have much to say about the right approach.  When you are ready for me to run a full test suite, let me know.

whlipscomb:
@JFLemieux73, I see the five lines in the code where you left comments. The answer isn't obvious. I've been drawing pictures trying to wrap my mind around what we're doing.

To summarize the Bentsen strategy: We start off with two triangles in the central region, and we draw a third (Bentsen) triangle to match the prescribed edgearea exactly.  In the most common case (call it Case 1), when DL and DR lie on the same side of the edge, these three triangles are formed by connecting the points CL-CR-DL, CR-DR-DL, and DL-DM-DR. Here, DM starts out as the midpoint of the segment connecting DL to DR. It is then shifted in a direction perpendicular to the segment DL-DR to form a Bentsen triangle with nonzero area.

The geometry is different (Case 2) if segment DL-DR crosses the edge at point IC. Then the triangles are CL-IC-DL, CR-IC-DR, and DL-IC-DM, are some similar combination. The triangle containing DM is again the Bentsen triangle. One of these triangles lies on the opposite side of the edge from the Bentsen triangle. Let's call this the "lone triangle".

If all the central triangles (not including the lone triangle) are computed with the same area factor (areafac_c in the current code), the math is exact. The total fluxed area should agree with the prescribed edgearea.

Things get tricky, though, if each triangle's area factor is based on the x coordinate of the triangle midpoint. In that case, I don't think the formula used to compute the location of the shifted DM will result in exactly the desired fluxed area. That's because the current formula assumes that all the central triangles (excluding the lone triangle) have the same area factor. 

If you have to choose a single value of areafac, is there a better choice than areafac_c? I think so. I might choose the value of areafac associated with the midpoint of either (1) the quadrilateral formed by CL-CR-DR-DL, for case 1, or (2) the triangle formed by either CL-DL-IC or CR-DR-IC (whichever is not the lone triangle) for case 2. You would compute the x value of this midpoint, apply your linear formula, and use the resulting areafac in the equations.

In the lines of code you've highlighted, I think you would leave areafac_l and areafac_r alone, because these are lone triangles. But in the other three lines, you would need to replace areafac_c with the new value computed for the central region.

That's not all. You would need to replace areafac_c with this new areafac (instead of areatp) for all the central triangles (except the lone triangles) in the computations below.

It would be easy to make a mistake implementing the new logic. But if you do things incorrectly, I think you'll find that edgearea doesn't have the prescribed value. If edgearea has the prescribed value (within machine roundoff), that's a good indication that the code is correct.

JFLemieux73:
Hi @whlipscomb,

I have modified the calculation for cases 1 and 2 as you describe above. I have not tested it yet but will do it as soon as I can. Please have a look and let me know what you think. 

Thanks again for your help!
jf

JFLemieux73:
Ok I tested it (gx3 simulation) with bugcheck=T. Case 1 is ok. However, out of the 8 possibilities for case 2, 4 of them have area that do not add up. I am trying to understand why but so far it looks like the ones that fail have the lone triangle in the BC cell. All case 2 with lone triangle in the TC cell work...

JFLemieux73:
Ok I know what is going on. It happens when one departure points is 0 and the other one negative. 

First let's look at a CASE that works: ydl>0, ydr=0 and ydm>0. In the l_fixed_area section it goes in 

if (ydl*ydr >= c0) then   ! Both DPs lie on same side of x-axis 

This is CASE 1

In the locate triangle section it goes in:

if (ydl >= c0 .and. ydr >= c0 .and. ydm >= c0) then 

Which also corresponds to CASE 1...good!

JFLemieux73:
Now a case that does not work:

ydl=0, ydr< 0, ydm<0

In the l_fixed_area section it goes in

if (ydl*ydr >= c0) then ! Both DPs lie on same side of x-axis

This is CASE 1

In the locate triangle section (BC and TC) it goes in:

elseif (ydl >= c0 .and. ydr < c0 .and. xic < c0  .and. ydm < c0) then

This is CASE 2...it does not work!

JFLemieux73:
To fix the problem I need to change the conditions:
 elseif (ydl < c0 .and. ydr < c0 .and. ydm < c0) then
by
elseif (ydl <= c0 .and. ydr <= c0 .and. ydm < c0)

Same idea for:
 elseif (ydl < c0 .and. ydr < c0 .and. ydm >= c0) then  ! rare
by
elseif (ydl <= c0 .and. ydr <= c0 .and. ydm >= c0) then  ! rare

I just tested it and the areas add up...

whlipscomb:
@JFLemieux73, thanks for the great detective work. I can see that there was a problem with the inequality logic, and that your fix makes the l_fixed_area code section consistent with the locate-triangle section.

I might suggest a few other very minor changes. In the locate_triangle section, ll. 2656, 2707, 2758, and 2809, the logic is
     (elseif (ydl >= c0 .and. ydr < c0 .and. [other stuff]) then

I think this could be replaced by
     (elseif (ydl > c0 .and. ydr < c0 .and. [other stuff]) then
since the case (ydl = 0, ydr < 0) is now correctly included within Case 1 by the changes you've made above.

Similarly, look at ll. 2860, 2911, 2962, and 3013. In these lines, the logic is
     (elseif (ydl < c0 .and. ydr >= c0 .and. [other stuff]) then

This could be replaced by
     (elseif (ydl < c0 .and. ydr > c0 .and. [other stuff]) then

I don't think these replacements would change the answers; they would just make the logic a bit cleaner. Let me know if that makes sense.

JFLemieux73:
Thanks Bill. I will make these changes. 

The good news with my latest changes is that the areas add up.

The bad news is that I just ran a gx3 simulation and there is a negative area after 3 years...

I have to go back to my detective work :)

whlipscomb:
@JFLemieux73, you also have four lines where something like this:
elseif (xic < c0) then  ! fix ICL = IC

was replaced by something like this:
elseif (xic < c0 .and. xic > xcl) then  ! fix ICL = IC ! Bill please check this

I drew some pictures and was unable to figure out a case where we get to this line of the code with xic <= xcl. Say that initially you have xic < xcl, with ydl < 0 and ydr > 0.  Then DL would be redefined (ll. 2298–2291) so that ydl > 0, and ydl would then have the same sign as ydr.

If you've run across cases where xic < xcl at this point of the code, then I've overlooked something. But even if there are no such cases, I think the change is a good one. If the second part of the conditional were violated, then the code that follows wouldn't make any sense.

whlipscomb:
@JFLemieux73, I'm sorry you found another negative area. I was hoping that with the latest changes, the code would just work.

Have you run a clean multiyear simulation with gx1? In other words, is this problem specific to gx3?

JFLemieux73:
I had a failure as well with gx1...

whlipscomb:
@JFLemieux73, Yes, I think for clarity, it's better to have "elseif (xic < c0 .and. xic > xcl)"

Let me know what you find out about the negative areas.

JFLemieux73:
I just added the changes you propose. I also made two small modifs (that would not change the answer). Just look for these comments:

! Bill I also changed...

JFLemieux73:
I will let you know about the failure but my impression is that it is the same problem we had before (like explained in the powerpoint I sent you). Right now what we have is consistent with the l_fixed_area approach (i.e. areas add up) but the areafac of triangle coming in on the east edge might not match the areafac of the same triangle going out (south edge coming from the TR cell)...more info soon.

JFLemieux73:
Hi @whlipscomb ,

I can confirm that the failure is the same we have seen before (see our discussion in issue #809).  Basically the flux in depends on a triangle on the east edge and the flux out of a triangle on the south edge. The other triangles are not important (fluxes are zero because initial aice0 = 0 in the grid cell). Using local coordinates the triangle IN has a larger area than the triangle OUT. So it should not be a problem. But again the situation reverses due to the areafac. Now the triangle IN has a lower area than the triangle OUT and this leads to a 'negative area' after the transport.

JFLemieux73:
Basically we had this problem before. I corrected it with solution 1 but areas did not add up. With the latest changes, the areas add up but the initial problem is back!

JFLemieux73:
Right now the l_fixed_area correction has areafac that depends on the midpoint of the quadrilateral formed by CL-CR-DR-DL. This gives a value close to areafac_c. The triangle IN uses areafac~areafac_c. The problem is that the triangle OUT (TR cell) uses areafac~areafac_r. 

Would it be a good idea to use areafac_c for the TL, BL, TR and BR triangles?

whlipscomb:
areafac_c is relative to the edge. So if you have a triangle fluxing area into a cell across the E edge, and a triangle of very similar size fluxing area out across the south edge, the areafac's would be different.

Out of curiosity, how big is the error you're seeing? If it's a very small error and we understand its source, there are worse crimes than resetting the negative area to zero.

JFLemieux73:
Sorry this was not clear. I am thinking of using the same areafac_c for the east edge calculation and for the TR triangle. To answer your question:

New mass < 0, i, j =          80         115
 Old mass =  0.000000000000000E+000
 New mass = -6.167789746423259E-011
 Net transport = -6.167789746423259E-011

whlipscomb:
That's a pretty small error. I doubt I would lose sleep if we set negative areas of that magnitude to zero.

Let's follow up after your vacation. Bonnes vacances!

JFLemieux73:
Merci Bill! And thanks again for your help.

JFLemieux73:
image

JFLemieux73:
**Here is a possible idea to fix this negative mass issue:**

image

image

image

JFLemieux73:
First the good news:

As set up, gx3 ran for 4 years. The areas add up with l_fixed_area=T except on rare occasions. I investigated one of these. It's no big deal; it is just roundoff  errors associated with very small triangles. The case I investigated is at the ice edge with uvel=vvel=0 for the 4 corners except one uvel=2.4e-6 m/s. The two uvelE and two vvelN associated with this grid cell are zero (because of ice edge aice values). The error in the bugcheck disappears if 

if (abs(areasum(i,j) - edgearea(i,j)) > eps13*areafac_c(i,j)) then

is replaced by

if (abs(areasum(i,j) - edgearea(i,j)) > eps12*areafac_c(i,j)) then

JFLemieux73:
Before gx1 failed with a negative (open water) area failed after 10 months. The new code (as expected) is more robust and eliminates this failure. However, gx1 fails again with a negative (open water) area after 1.5 year. I think it is a different problem: it is  not for a grid cell touching land. All corner (uvel,vvel) velocities are non-zero. I will investigate.

JFLemieux73:
There is something wrong with the halo. I get different answers with different decompositions. While the standard gx1 (p 44x1) failed after 1.5 year, a 1 proc (p 1x1) simulation is still running after almost 2 years. Hopefully this halo problem explains both the different answers and the negative area failures. 

@whlipscomb and @apcraig could you please have a look at my latest commit and tell me if you find what's wrong?

JFLemieux73:
Ok it is clearly a halo problem. 

 ! area scale factor for other edge (north)

         do j = jb-1, je
         do i = ib, ie+1
            areafac_l = dxu(i-1,j)*dyu(i-1,j)
            areafac_r = dxu(i  ,j)*dyu(i  ,j)
            areafac_ce(i,j) = p5*( areafac_l + areafac_r )
         enddo
         enddo

In this loop to calc areafac_ce(i,j), ib and jb are 1. So it needs dxu(0,j)=dyu(0,j)=0

It is not what we want...

apcraig:
It sounds like dxu and dyu are not initialized on the halo.  They are just static grid parameters?  I'll take a look.

TillRasmussen:
Should the array run from 1 to nx_block? Then it is more an out of bound issue?

apcraig:
dxu and dyu are NOT valid on the halo.  I think even uarea is not.  We can do a halo update at initialization on other grid variables, that's no problem.

Good point, why are ib=jb=1?  0 is not a valid index in the arrays?  Have you compiled with debug flags to check bounds and so forth?

apcraig:
Looks like ib=ilo, jb=jlo.  So that's correct.  Are you sure they are 1 at runtime?  If it's ilo and jlo, then the indices are correct and what we need to do is halo update some grid variables.  What grid variables do you need to have halo values for?

TillRasmussen:
Assuming that I am looking at the correct place (line 1922/1948) then I think that ib is ilo-1

apcraig:
OK, here is more code.

This is quite awkward.  Is jb=jlo-nghost actually correct?  Why would you shift jb "nghost" values where nghost can be 1 or 100.  Should it be jb=jlo-1?  Up to now, nghost has always been 1 in practice.

I'm not in love with this coding.  I think it would be clearer and more consistent with the rest of the code if the loops were i=ilo,ihi, j=jlo,jhi, then the indexing in the arrays would be i,j-nghost.  Thoughts?

```
         ! loop size                                                                                                                                                                            

         ib = ilo
         ie = ihi
         jb = jlo - nghost            ! lowest j index is a ghost cell                                                                                                                          
         je = jhi

         ! index shifts for neighbor cells                                                                                                                                                      

...

         ! area scale factor                                                                                                                                                                    

         do j = jb, je
         do i = ib, ie
            areafac_l(i,j) = dxu(i-1,j)*dyu(i-1,j)
            areafac_r(i,j) = dxu(i  ,j)*dyu(i  ,j)
            areafac_c(i,j) = p5*(areafac_l(i,j) + areafac_r(i,j))
         enddo
         enddo

```

JFLemieux73:
I have not used the debug flag. I guess it is out of bounds as you guys suggest. dxu*dyu is currently used in remapping. 

Let me explain what I am trying to do and maybe you could guide me. Let's we are at the north edge. Before the code needed only areafac_c:

areafac_l = dxu(i-1,j)*dyu(i-1,j)
areafac_r = dxu(i  ,j)*dyu(i  ,j)
 
areafac_c(i,j) = p5*( areafac_l + areafac_r )

 Now at the north edge we also need values on the four east edges around:

areafac_ce(i-1,j), areafac_ce(i-1,j+1), areafac_ce(i,j+1), areafac_ce(i,j)

It when I calculate these areafac_ce that I have some dxu,dyu not defined.

apcraig:
@TillRasmussen there is north and east implementations.  In one case, i is shifted by nghost but the indices are always i,j-1.  In the other, j is shifted by nghost but the indices are i-1,j.  So I think the references are all within the array bounds.

apcraig:
@JFLemieux73.  I think the array references are OK in terms of bounds.  So I think areafac_l and areafac_r are computed from valid memory references.  If not, we would have caught this a long time ago.  My guess is that when i=1 then areafac_l is just zero (incorrect) but never used.  Now that you have areafac_c, that is being used at i=1 and so we're seeing the problems.  I could be wrong, haven't looked through the whole code, but the problem is NOT the computation of areafac_c, it has to be areafac_l that is the problem, and that error has probably existed forever, just never caught or as I speculate, played no role in the computation at i=1., ever.

So again, what grid arrays need the halo values?  Clearly dxu and dyu.  Anything else?

Also, please have a look at the use of nghost in the subroutine, I think that's wrong and it should probably be 1.

JFLemieux73:
jb = jlo - nghost was there before. We could fix it in this PR as well. Yes dxu and dyu would be needed in the halo. By the way areafac_c and areafac_ce are fixed and could be calculated only once and stored (with values in the halo).

TillRasmussen:
@apcraig I agree that we are in the east case but this is the only place where I can find 

! area scale factor for other edge (north)

     do j = jb-1, je
     do i = ib, ie+1
        areafac_l = dxu(i-1,j)*dyu(i-1,j)
        areafac_r = dxu(i  ,j)*dyu(i  ,j)
        areafac_ce(i,j) = p5*( areafac_l + areafac_r )
     enddo
     enddo

that @JFLemieux73  refers to (from line 1989. Here ib is ilo-nghost.

apcraig:
The nghost usage in the transport_remap is very odd.  The comments and the code don't seem to entirely agree.  And the implementation doesn't seem to be entirely consistent, again from a quick look.  Why is dpx only halo updated for nghost=1 but then dpx halo memory is referenced on the north and east edges in "Compute mask for edges with nonzero departure areas", with an nghost offset.  This only seems like it works because nghost=1 all the time.

eclare108213:
> This is quite awkward. Is jb=jlo-nghost actually correct? Why would you shift jb "nghost" values where nghost can be 1 or 100. Should it be jb=jlo-1? Up to now, nghost has always been 1 in practice.
> 
> I'm not in love with this coding. I think it would be clearer and more consistent with the rest of the code if the loops were i=ilo,ihi, j=jlo,jhi, then the indexing in the arrays would be i,j-nghost. Thoughts?

This is left over from POP days.  The intention with nghost is 2-fold:  first, it makes it clear that the shift in the loop index is referring to ghost (i.e. halo) cells, and second it more easily allows the depth of the halo to increase to 2, say, if we wanted to use a different advection scheme (or something) that needs a bigger halo.

cicecore/cicedyn/dynamics/ice_transport_remap.F90
```diff
@@ -1939,13 +1965,34 @@ subroutine locate_triangles (nx_block,     ny_block,   &
...
+         enddo
+         enddo
+
+         ! area scale factor for other edge (north)
+
+         do j = jb-1, je
+         do i = ib, ie+1
+            areafac_l = dxu(i-1,j)*dyu(i-1,j)
+            areafac_r = dxu(i  ,j)*dyu(i  ,j)
+            areafac_ce(i,j) = p5*( areafac_l + areafac_r )
```
apcraig, line 1995: What's happening here.  Why do you have a 2nd loop that overwrites most of the first loop.  Also, @TillRasmussen is correct, these indices are problematic.  Is there confusion about ib=ilo?  I'd love to see ib and ie removed and use ilo and ihi with explicit offsets instead to make things clearer.

apcraig:
> > This is quite awkward. Is jb=jlo-nghost actually correct? Why would you shift jb "nghost" values where nghost can be 1 or 100. Should it be jb=jlo-1? Up to now, nghost has always been 1 in practice.
> > I'm not in love with this coding. I think it would be clearer and more consistent with the rest of the code if the loops were i=ilo,ihi, j=jlo,jhi, then the indexing in the arrays would be i,j-nghost. Thoughts?
> 
> This is left over from POP days. The intention with nghost is 2-fold: first, it makes it clear that the shift in the loop index is referring to ghost (i.e. halo) cells, and second it more easily allows the depth of the halo to increase to 2, say, if we wanted to use a different advection scheme (or something) that needs a bigger halo.

Exactly, but even if you only need nghost=1 for the implementation, you could set it to 2 or 10.  It just buffers the arrays.  So why is the indexing used in the remapping using nghost offsets.  That cannot be correct.  You should be able to increase nghost, run CICE, and get bit-for-bit answers with nghost=1 (assuming nghost=1 is all that is needed algorithmically).  Maybe we should start testing this, looks like it would fail as things are currently implemented.  The current remapping only seems to need neighbor values (first level halo).  Why does nghost come into play in the implementation.  And why is dpx only halo updated when nghost=1 if it needs the halo values?

eclare108213:
Those are all good questions @apcraig.  This is ancient code, possibly never tested for anything but nghost=1.  My memory is a bit vague on this (@whlipscomb would know for sure) but I think the incremental remapping was originally coded with a higher order option, which might have needed more ghost cells. 

If there are (hopefully relatively easy) ways to clean it up and make it more robust, then go for it!  Otherwise we can abort if anyone tries to run with nghost /= 1.

cicecore/cicedyn/dynamics/ice_transport_remap.F90
JFLemieux73: It does not overwrite it. The first array is areafac_c the other one is areafac_ce.

apcraig: Good point, areafac_l and areafac_r are recomputed though.

whlipscomb:
@eclare108213, the code is ancient indeed. I have a vague memory of wanting to support nghost /= 1 in case we wanted to have fewer halo calls with a wider halo, but I don't remember the specifics. So in this discussion, I'm afraid I won't be of any help to @JFLemieux73 and @apcraig.

TillRasmussen:
@JFLemieux73 with respect to what to do.

Is it necessary to calculate the areafac_ce on the halo (i or j = 1)?.
If not this will at least fix the out of bound problem.

If it is needed then nghost then either the number of ghost cells needs to be increased or som fix needs to be made

apcraig:
I have created a sandbox with these mods.  I have confirmed that with debug flags, the bounds check fails.  Also, it's not bit-for-bit with different decompositions.  So, before we start to halo update a bunch of grid fields, lets check whether the indexing that is currently in place is correct.  One thing that might be helpful, as I noted above, is to get rid of ib, ie, jb, je and work with ilo, ihi, jlo, jhi directly.  That seems like an easy modification.   I think there is currently confusion about the loop indexing and the array indexing because of this.  I will also have a look at the current code and see whether I can help sort this out.

apcraig:
I did some debugging.  Don't understand the implementation, but this is what I know.

- for "north", areafac_c is the average to the west, areafac_ce is the average to the south.
- for "east", areafac_c is the average to the south, areafac_ce is the average to the west
- areafac_c is valid everywhere it's computed and used, sometimes on the halo, I see no problems with areafac_c.
- areafac_ce references points on the halo for the following cases
  - TL, group1, east and west halo points
  - BL, group1, west halo points
  - TR, group2, east halo points
  - BR, group2, west halo points
- North and south halo points are never used for areafac_ce after they are computed.
- areafac_ce computes values in the gridcell plus on the west, south, and north halo for the "north" points.  areafac_ce computes values in the gridcell plus the west, east, and south halo for the "east" points.
- areafac_ce computation on the west halo is out of bounds for "east" points and no values are computed on the east halo.
- areafac_ce computation is invalid on the south halo for the "north" points.
- The east and west halo points can be computed fine for the "north" points.  The east points can be computed for the "east" points (but they aren't now), but the west halo values cannot be computed for the "east" points.  That would require a second level of halo points.

This is all extremely confusing.  But the are several takeaways in the current implementation
- You are NOT computing areafac_ce everywhere that values are referenced.  Is it correct that some of those values should not be computed and set to 0.  I don't know.
- If you need to compute the west halo values of areafac_ce for the "east" points, you need 2 ghost cells.

I can help fix some of this, but I don't know what the correct implementation is.  Do we need 2 ghost cells to make this work?  Is it correct that some values should be zero?

apcraig:
Also, if we know exactly what gridcells are needed for the areafac_ce and we decide a 2nd layer of ghost cells are needed for dxu and dyu on the west face, we don't necessarily need to set nghost=2.  We could just do an extra halo update for some new dxu and dyu arrays that are shifted east one gridcell.  The new shifted arrays would then contain the information in the halos from the 2nd level of halo cells.  Again, this works fine if it's just one edge and a couple arrays.  If the requirement for a larger halo is more extensive, then we may need to go to nghost=2.

@JFLemieux73, I think the first step is to confirm the use of areafac_ce when areafac is set in the 12 special cases including the ise/jse shifts.  At the same time, we should be able to define for each of those 12 cases what parts of the gridcell and halo need to have a valid computation of areafac_ce and whether any values should be set to zero.  Once we're sure about that, we can revisit the computation of areafac_ce including on each halo edge to make sure we're computing the values everywhere they're needed without going outside the bounds on a standard haloed block.

JFLemieux73:
Thanks @apcraig and @TillRasmussen for your help. I don't have time to look at this today but will do tomorrow. I will keep you informed.

apcraig:
@JFLemieux73, just checking on the status of this.  Is it ready for further review or testing?

JFLemieux73:
Hi Tony,

Sorry for the delay. I was busy with other stuff. The latest change I proposed fixes the negative area issue. There is also no more out of bound problems. However, there is another (rare) case that leads to another negative area. I am trying to find a solution but it is tricky.

JFLemieux73:
Ok here is an explanation of the latest failure. All the individual triangle cases TL, BL, TR and BR were fine. The problem is related to rare cases where there are a combination of TL,BL or TR,BR triangles. In fact, the problematic triangles were TL1, BL2, TR1 and BR2. Here is an example with TR1-BR1:

image

JFLemieux73:
The code uses narea(i+1,j) for the TL triangle (east edge). In the example above, this is a flux in. Related to this, there is a flux out at the north edge. BR1 is (and was) fine: it uses earea(i,j) for areafac. However, the failure was due to the fact that the TR1 triangle (almost the same TL triangle) used earea(i,j+1). The new code fixes that: it uses narea(i+1,j) for the TR1 triangle. Same idea for other similar cases (TL1, BL2 and BR2).

JFLemieux73:
Ohhhhhhhhh yeahhhhhhhh. Both gx3 and gx1 ran fine for 4 years. I will do more testing next week but it looks good.

dabail10:
Great news! Maybe you didn't want to hear this, but I wanted to ask about one gridpoint wide channels. Are we still getting zero transport there?

whlipscomb:
@JFLemieux73, that's a great analysis. Nice that it has a straightforward solution without too many code changes.

@dabail10, I hear you. I'll do what I can to help with the one-gridpoint channels, though it might be down the road a bit.

JFLemieux73:
@dabail10, yes we still get zero transport in one-grid point channels. I suggest we do it in two separate PRs.  Let's finish this one and then we could have a loot at one-grid point channels.

JFLemieux73:
Here are the total ice volume for the old B-grid simulation (the current main code), the new B-grid simulation (this PR code) and the new C-grid simulation (this PR code). Here is gx3 volumes for the Arctic:

volArctic_gx3_B_old_B_new_C_new

JFLemieux73:
The new B-grid sim is on top of the old one...good. As we have seen before with gx3 in the Arctic (see issue #809) the C-grid sim has less volume than B-grid sim (more pronounced in summer). Here is the same for the Antarctic:

volAntarctic_gx3_B_old_B_new_C_new

JFLemieux73:
This looks good: the new B-grid volume is on top of the old B-grid volume and the C-grid volume is very similar to the B-grid volumes for both polar regions.

JFLemieux73:
I juts ran the end-2-end procedure (base suite). As expected there are many runs that are not BFB (especially gx3 and gx1 runs) Here are the results:

427 measured results of 427 total results
270 of 427 tests PASSED
0 of 427 tests PENDING
0 of 427 tests MISSING data
157 of 427 tests FAILED

eclare108213:
@JFLemieux73 this looks great!

dabail10:
This looks great. I can run a QC test on this.

JFLemieux73:
Thanks Dave. I am doing the QC test right now...results in a few minutes!

JFLemieux73:
Here are the results for the B-grid (base: current main code, test: this PR code):

INFO:__main__:  /home/jfl001/data/ppp6/cice/runs/ppp6_intel_smoke_gx1_44x1_medium_qc.qc_base/history                                             
INFO:__main__:  /home/jfl001/data/ppp6/cice/runs/ppp6_intel_smoke_gx1_44x1_medium_qc.qc_test/history                                                

Good!

dabail10:
That is awesome!

TillRasmussen:
Nice:)

apcraig:
Looks good.  I can run a more comprehensive test suite on cheyenne in the next day or two.  @JFLemieux73, are we ready for that?

For the QC, just want to confirm what was done.  These are standard tests with B grids?  One QC test uses the current trunk?  The other uses the current trunk plus this PR?

I guess we expect that the changes will be very small and very infrequent, so it should pass easily and I'm not too worried overall.

Should we try to run a B vs C QC test or other tests like that?

JFLemieux73:
> Looks good. I can run a more comprehensive test suite on cheyenne in the next day or two. @JFLemieux73, are we ready for that? 

YES WE ARE READY.
 
> For the QC, just want to confirm what was done. These are standard tests with B grids? One QC test uses the current trunk? The other uses the current trunk plus this PR?

CORRECT.

> I guess we expect that the changes will be very small and very infrequent, so it should pass easily and I'm not too worried overall.

YES WE SHOULD EXPECT NON BFB RESULTS BUT RESULTS VERY SIMILAR.

> Should we try to run a B vs C QC test or other tests like that?

I AM DOING THAT RIGHT NOW.

JFLemieux73:
QC was applied to the new B-grid solution (this PR) vs the new C-grid (this PR). 

NFO:__main__:Number of files: 1825
INFO:__main__:Quadratic Skill Test Failed for Northern Hemisphere
INFO:__main__:Quadratic Skill Test Failed for Southern Hemisphere

ERROR:__main__:Quality Control Test FAILED

So the total volume for B and C-grid simulations are similar but there are differences in the spatial patterns. I looked at snapshots of hi after the 5 years of simulation. Here is the B-grid solution (0 to 5 m):

B_hi_2009-12-31-00000

and here is the C-grid solution:

C_hi_2009-12-31-00000

JFLemieux73:
Here are hi differences produced by the QC test:

JFLemieux73:
I say we go ahead with this. 

The new code has a very similar B grid solution as the current main code. This is good. 

The C grid code is now robust and runs with l_fixed_area = T (the checkerboard instability is eliminated). 

The total volume for B and C grid solutions are very close. There are however differences in how these volume are spatially distributed.

apcraig (commented):
Thank you for updating the ib,ie,jb,je implementation.  Use of ilo,ihi,jlo,jhi is very helpful and make it easier to understand.  Relative to the previous version, there are differences in some loop start and end indexes, the areafac_c and areafac_ce uses a different value (to roundoff) and some of the outlier triangle cases are different.  Hopefully this fixes the out of bounds, have started comprehensive testing on cheyenne now.  And hopefully the B case is OK in the sense that the loop index changes don't affect answers, the areafac_c/ce is roundoff different, and the triangle fixes are very infrequent in time and space.  QC testing suggest all of that is probably true.  Would be good to just have a final review, especially of the loop indexes.

Will approve code review once testing on cheyenne is complete.

JFLemieux73:
Thanks Tony. The loop start and end indexes should be the same. They have just been replaced by ilo,ihi,jlo,jhi. Also, a difference is that nghost has been replaced by '1'. For example, for the east edge for the areafac_c loop, ib ( which is ilo - nghost) has been replaced by ilo - 1.

apcraig:
Full test suite on cheyenne indicates that with these changes, we have broken bit-for-bit identical with different decompositions.  I suspect this has something to do with the loop indices.  Could it be anything else?  Some part of the solver is now sensitive to the size of the block.  It earlier testing, I indicated that the loop indexing seemed odd in the sense that some values were left at "zero" and not computed when it seemed like maybe they should be.  I think we need to revisit that a bit.  I can try to debug to see what I can find too.

apcraig:
I have created a PR to @JFLemieux73 areafact branch, https://github.com/JFLemieux73/cice/pull/1.  This seems to fix the bit-for-bit reproducibility for different block sizes.  A description of the fix can be found there.  I am carrying out a full test suite now, but I think it's safe to merge that PR so it can be seen here.  We should redo all the QC testing and any other manual testing that was done just to make sure everything is still OK.  I don't expect the fixes to impact science, but it will change answers for all cases including B grids relative to this implementation on this branch currently.

apcraig:
Full test suite of intel on cheyenne with the latest update is looking good.  B, C, and CD results are NOT bit-for-bit with the current main which we expect.  A few box tests are though, so that's probably good.

Will run a full test suite now with pgi and gnu.

JFLemieux73:
New code for B-grid and C-grid ran fine for 4 years for both gx3 and gx1. Time series for volume are the same as before (see figures above in this PR). I am running the base suite right now.

JFLemieux73:
Here are the results for the base suite:

427 measured results of 427 total results
272 of 427 tests PASSED
0 of 427 tests PENDING
0 of 427 tests MISSING data
155 of 427 tests FAILED

Note that we had 157 failures before. I guess there are less failures due to your latest fix Tony.

apcraig (approved):

apcraig:
If you want me to merge this, just let me know.  Not entirely clear if there is still a bit of cleanup to do here or if that will come in future PRs.

eclare108213 (commented):
I request that documentation be updated to be consistent with how we want this code to be run, if that hasn't already been done.  Changes could be included in the next PR, though my preference is to keep code and docs as consistent with each other as possible all the time.

JFLemieux73:
The documentation has been updated. It is mentioned that l_fixed_area = F for B-grid and = T for C-grid to eliminate a checkerboard instability. The current (main) doc does not mention anything about areafact. As most of the modifications in this PR are related to areafact, I don't think we should further modify the doc. I think this PR is ready to be merged.

apcraig:
@eclare108213, I will merge today unless you disagree.

eclare108213:
go for it!

PR #853, opened 2023-08-07
Update conda_macos to fix problems with Github Actions testing

apcraig:
## PR checklist
    Update conda_macos to fix problems with Github Actions testing
- [X] Developer(s): 
    apcraig
    Tested with github actions, fixes problems that developed a couple weeks ago.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Tried many things to debug, nothing worked.  MPI builds were failing.  Switched from mpifort to gfortran with the appropriate include and lib flags added to conda_macos.  Problems went away.  This is not ideal, but works.

The conda_macos was always building fine on my Mac desktop.  Github actions was working one day, not the next.  Could be something with conda too.

The github workflow was modified to build/run a quick helloworld unit test and to build an MPI configuration BEFORE the inputdata is downloaded and before the test suite is run.  This just shortens the time for github actions to fail if there are problems with the build and/or basic run setup.

### DIFF
--- a/.github/workflows/test-cice.yml
+++ b/.github/workflows/test-cice.yml
@@ -87,7 +87,6 @@ jobs:
       - name: clone
         uses: actions/checkout@v2
         with:
-          lfs: true
           submodules: 'recursive'
       - name: link
         run: |
--- a/.github/workflows/test-cice.yml
+++ b/.github/workflows/test-cice.yml
@@ -143,6 +143,13 @@ jobs:
         run: |
           cd $HOME/cice
           ./cice.setup -m conda -e ${{ matrix.envdef }} --test smoke --testid c0
+      - name: run hello world unit test
+        run: |
+          cd $HOME/cice
+          ./cice.setup -m conda -e ${{ matrix.envdef }} --test unittest --pes 2x1 -s helloworld --testid hw01
+          cd *helloworld*hw01
+          ./cice.build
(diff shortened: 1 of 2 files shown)


phil-blain:
Hi Tony, I did not have time to look into the failures. Glad that you found something that works, even if it's not ideal to not use the MPI wrappers.

I like the additions to the workflow, I think it's a good idea to try the build before downloading the input data. 

What I would suggest though is removing the commented lines in both files.

phil-blain:
I also just noticed that the workflow YAML was not updated when we renamed the default branch from `master` to `main`:

~~~yaml
on:
  push:
    branches:
      - master
~~~

this means that the CI does not run on the `main` branch (only on pull requests). 

I think it would be a good idea to also have it run on the `main` branch. This way we could figure out more easily what changes between a working run and a non-working run, without chasing down different PRs.

apcraig:
Thanks @phil-blain, I have fixed the master/main thing.  I'm going to leave the commented lines alone for now.  It's helpful to have them around for debugging the next time.  At some point, maybe the mpi wrappers will be working again.  The commented lines are a reminder to try again in the future.

PR #876, opened 2023-10-02
Add atm_data_version to allow JRA55 forcing filenames to have a unique version string

### LINKED ISSUE #872: JRA55 tx1 forcing is unphysical (JRA55do is OK)
In the process of testing the C grid code with the new areafact computation in a long `tx1` simulation, I found out that the current tx1 JRA55 forcing is unphysical, likely coming from the interpolation procedure used to create it.

For example, here are the x-ward winds at the first time slice of year 2005:

image

Similar patterns can be found for the other years and the rest of the forcing fields. The longitude field in the forcing files is also wrong:

image

compared to the correct tx1 longitude field (from the tx1 bathymetry file):

image

**Fortunately it seems the newer "JRA55do" data is correctly interpolated.** 

So this raises a few points:

- What script was used to create the JRA55 data ? to create the JRA5do data ? I found https://github.com/CICE-Consortium/CICE/blob/main/configuration/tools/jra55_datasets/interp_jra55_ncdf_bilinear.py, which dates back from the original addition of JRA55 in e204fb826412288f8f789b2df3a71c9c78af2099. But it's unclear if this same script was used to create the JRA55do dataset. 
- What should we do with the current JRA55 dataset, which is on Zenodo ? should we create a new version, correctly interpolated, or just deprecate it and use JRA55do for the tx1 grid ?

/cc list based on #435: @daveh150 @rallard77  @dabail10 @apcraig

daveh150:

## PR checklist
    Add ability to read JRA55 forcing file with creation date in filename.
- [x] Developer(s): 
    David Hebert
    ENTER INFORMATION HERE 
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 
With external file forcing updates, we are looking for method to include the file creation date in the file name. This is specific to JRA55 forcing,  

This PR also addresses unphysical JRA55 tx1 forcing, so it Closes #872


daveh150:
This update adds a jra55date variable to allow specification of the file creation date. Intent is to have naming convention

JRA55[do]_jra55date_grid_03h_forcing_year.nc

JRA55_files subroutine will search first for file with date, then continue to search for file without date. On our local system this seemed to work, the only 2 base-suite fails are the ones that use JRA55 tx1 forcing, since the forcing files are different (the baseline does not use the file with dates):

362 measured results of 362 total results
360 of 362 tests PASSED
0 of 362 tests PENDING
0 of 362 tests MISSING data
2 of 362 tests FAILED

FAIL nrlssc_gnu_restart_tx1_40x4_dsectrobin_medium

I copied the new files in Cheyenne for anyone to try testing (that has access to Cheyenne). The are here:

/glade/scratch/dhebert/CICE_data/forcing/tx1/JRA55/8XDAILY/JRA55_20230919_tx1_03hr_forcing_200[5-9].nc

Happy to entertain other methods to include dates in the file names.  Thank you!

apcraig (commented):
Several comments.
- Any new namelist has to be documented in the user guide.
- I suggest we don't call this "jra55_date".  I think it should be more general, "atm_data_date", then it can be used for any kinds of atm forcing data.  In ice_init.F90 (line 2365), atm_data_date can be documented always, not just with jra55.
- I would set the default atm_data_date to something non-functional in the code (ice_init.F90), like "_undef".  I would set atm_data_date to the appropriate string via namelist as is done.
- There is a problem with the "_" in the implementation.  If jra55_date (atm_data_date) is empty, we have two underscores next to each other, this will not be backward compatible.  So we probably need atm_data_date to be "_230919" and for the underscore implementation to be updated.

cicecore/cicedyn/general/ice_forcing.F90
```diff
@@ -2237,30 +2238,42 @@ subroutine JRA55_files(yr)
...
                                     '/8XDAILY/'//trim(atm_data_type_prefix)//                '_03hr_forcing_2005.nc'
 
+
          call file_year(uwind_file,yr)
          INQUIRE(FILE=uwind_file,EXIST=exists)
-!         if (my_task == master_task) then
-!            write(nu_diag,*) subname,cnt,exists,trim(uwind_file)
-!         endif
+         if (my_task == master_task) then
+            write(nu_diag,*) subname,cnt,exists,trim(uwind_file)
```
apcraig, line 2275: I think this should be commented out by default.

cicecore/cicedyn/general/ice_forcing.F90
```diff
@@ -2237,30 +2238,42 @@ subroutine JRA55_files(yr)
...
+         ! search for jra55 file with creation date (jra55date) in filename first
          if (cnt == 1) uwind_file = trim(atm_data_dir)//'/'//trim(atm_data_type_prefix)// &
+                                    '/8XDAILY/'//trim(atm_data_type_prefix)//'_'//trim(atm_data_date)//'_'// &
+                                    trim(grd)//'_03hr_forcing_2005.nc'
+
+         if (cnt == 2) uwind_file = trim(atm_data_dir)//                                  &
+                                    '/8XDAILY/'//trim(atm_data_type_prefix)//'_'//trim(atm_data_date)//'_'// &
+                                    trim(grd)//                                              '_03hr_forcing_2005.nc'
+
+         ! now search for files without jra55date in filename.
```
apcraig, line 2252: I don't think we should ever search for file without "atm_data_date".  The code will read the old files.  That means we need to set atm_data_date='' in ice_in, and it needs to be set to '_20230919' for tx1.

cicecore/cicedyn/general/ice_forcing.F90
```diff
@@ -2237,30 +2238,42 @@ subroutine JRA55_files(yr)
       ! cnt represents the possible file format options and steps thru them until one is found
       exists = .false.
       cnt = 1
-      do while (.not.exists .and. cnt <= 6)
+      do while (.not.exists .and. cnt <= 8)
+
+         ! search for jra55 file with creation date (jra55date) in filename first
          if (cnt == 1) uwind_file = trim(atm_data_dir)//'/'//trim(atm_data_type_prefix)// &
+                                    '/8XDAILY/'//trim(atm_data_type_prefix)//'_'//trim(atm_data_date)//'_'// &
```
apcraig, line 2245: We still have the problem of double underscore if atm_data_date is empty.  This should be

 '/8XDAILY/'//trim(atm_data_type_prefix)//trim(atm_data_date)//'_'// &

and atm_data_date should have a leading _ if that's part of the filename prefix.

configuration/scripts/options/set_nml.jra55
```diff
@@ -1,2 +1,3 @@
 atm_data_format = 'nc'
 atm_data_type   = 'JRA55'
+atm_data_date   = '20230919'
```
apcraig, line 3: I think this setting is incorrect.  jra55 forcing files by default will not be _20230919, only the tx1 file has that setting.

configuration/scripts/ice_in
```diff
@@ -268,6 +268,7 @@
     precip_units    = 'mm_per_month'
     default_season  = 'winter'
     atm_data_type   = 'ncar'
+    atm_data_date   = 'undefined'
```
apcraig, line 271: This should probably be atm_data_date = ''.

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -601,7 +601,8 @@ forcing_nml
    "", "``constant``", "constant-based boundary layer", ""
    "", "``mixed``", "stability-based boundary layer for wind stress, constant-based for sensible+latent heat fluxes", ""
    "``atmiter_conv``", "real", "convergence criteria for ustar", "0.0"
-   "``atm_data_dir``", "string", "path or partial path to atmospheric forcing data directory", ""
+   "``atm_data_dir``", "string", "path or partial path to atmosphericforcing data directory", ""
+   "``atm_data_date``","string", "date of atm data forcing file creation", "``undefined``"
```
apcraig, line 605: These are alphabetical, atm_data_date should go before atm_data_dir.

apcraig:
Thanks @daveh150 for the updates.  I have added several other comments.

daveh150:
I'm testing now with the changes you suggested. I originally was trying to avoid an underscore in the atm_data_date, but I don't think that is possible if we are always going to check for the date in the filename. Sorry I should have waited to push to my fork before the tests were completed.

apcraig:
@daveh150, no worries, thanks for taking this on.

daveh150:
I modified the JRA55_files subroutine to include atm_data_date in all its checks. I put the file creation date just before the year, so the convention will be JRA55_tx1_03hr_forcing_20230919_2005.nc. 

As noted, this will require an underscore before the date. set_nml.gx1, set_nml.gx3, set_nml.tx1 needed to be updated with the proper atm_data_date. 

One note if interest is I added a set_nm.jra55, but this has the atm_data_date specified to _20230919. It can only be used with tests needing the date. This is why set_nml.gx[13] needed to be updated to avoid the '_undef' default.  

Also updated the documentation so it is alphabetical and notes the need for underscore in date. 

The base suite ran successfully. If there are any other suggestions, please let me know. Thanks!

cicecore/cicedyn/general/ice_forcing.F90
apcraig: I'm still concerned this should be commented out by default.

configuration/scripts/tests/base_suite.ts
```diff
@@ -11,7 +11,7 @@ smoke          gx3     4x2        diag1,run5day             smoke_gx3_8x2_diag1_
 smoke          gx3     4x1        diag1,run5day,thread      smoke_gx3_8x2_diag1_run5day
 smoke          gx3     1x8        diag1,run5day,evp1d
 restart        gx1    40x4        droundrobin,medium
-restart        tx1    40x4        dsectrobin,medium
+restart        tx1    40x4        dsectrobin,medium,jra55
```
apcraig, line 14: See comment above.

apcraig:
I see the location of the "atm_data_date" has changed in the filename convention.  I'm OK with this.  But that means the raw data names have to change.  Was this also updated on your test system, on zenodo, and on cheyenne and were the old files removed?

I'm also having second thoughts about atm_data_data as the name for this string.  That name might be confused with the data associated with the forcing data (i.e. 2005).    Also, this unique string doesn't necessarily have to be a date, that's just the convention we've picked for now.  Maybe this should be called "atm_data_string"??  Anyone have thoughts on this?

phil-blain:
> Also, this unique string doesn't necessarily have to be a date, that's just the convention we've picked for now. Maybe this should be called "atm_data_string"?? Anyone have thoughts on this?

I think something like `atm_data_version` would be indeed clearer.

apcraig:
@daveh150, can we get this PR revised and merged this week, including getting the data on zenodo?  Let me know if you need some help with something, I'm happy to contribute.

daveh150:
Hi  @apcraig Thank you for checking. I will start to work on this again now and the rest of the week. I was on leave last week and just getting dug out from end of fiscal year reports. 

I need to read your last comments, but I think we should be able to get the code updated by the end of the week. I'll need help with Zenodo when the time comes. Thank you!

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -601,7 +601,7 @@ forcing_nml
    "", "``constant``", "constant-based boundary layer", ""
    "", "``mixed``", "stability-based boundary layer for wind stress, constant-based for sensible+latent heat fluxes", ""
    "``atmiter_conv``", "real", "convergence criteria for ustar", "0.0"
-   "``atm_data_dir``", "string", "path or partial path to atmospheric forcing data directory", ""
+   "``atm_data_dir``", "string", "path or partial path to atmosphericforcing data directory", ""
```
apcraig, line 604: was this change intentional?

configuration/scripts/options/set_nml.jra55
```diff
@@ -1,2 +0,0 @@
-atm_data_format = 'nc'
-atm_data_type   = 'JRA55'
```
apcraig, line 2: is there a reason we're getting rid of this file?

apcraig:
Thanks @daveh150.  This looks good overall.  There are three comments in the file changes, could you review them and comment/fix.

I don't think we want to upload new gx3/gx1 files that just have the name changed.  The point of this new feature is to make it possible to change forcing versions via namelist.  And we support atm_data_version = '' which is the appropriate value for the current files (except JRA55 tx1).  We do have a small problem that as we update versions of JRA55 and JRA55do, we may have to modify some of the set_nml options to make this work automatically.  But I think we can take that on when it happens.  The main thing is that the gx3/gx1/tx1 cases run with JRA55 or JRA55do out of the box as part of our test suites.  And I think that's true as things are setup now.  I will run a test suite myself to confirm once we feel the code changes are finalized.

apcraig:
Thanks @daveh150, still a couple other things before I test.  ice_forcing.F90, line 2272 still has a comment on it.  Can you look at that.  I think that debugging output needs to be commented out again.  Also, the documentation needs to be updated, it still stays "atm_data_date" in the text, that needs to be changed to "atm_data_version" and the string can be anything, not just yymmdd.  Thanks.

daveh150:
Thank you @apcraig I fixed the developer guide text and comment out the debug check.

apcraig:
Thanks @daveh150, I will start testing.  The documentation still says 

"``atm_data_version`` is a string in form ``_YYYYMMDD`` that allows user to specify the creation date of the atm_forcing_file".

This is not true.  atm_data_version could be any string or even blank string.  Also, at this point, maybe we should recommend a file naming convention a bit more explicitly.  How about if the documentation is updated as follows.

- The namelist ``atm_data_dir`` specifies the path or partial path for the atmosphere input data files, and the namelist ``atm_data_type`` defines the atmospheric forcing mode (see forcing_nml in :ref:`tabnamelist`).  Many of the forcing options are generated internally.  For atmospheric forcing read from files, the directory structure and filenames depend on the grid and ``atm_data_type``.  Many details can be gleaned from the CICE_data directory and filenames as well as from the implementation in **ice_forcing.F90**.  The primary ``atm_data_type`` forcing for gx1, gx3, and tx1 test grids are ``JRA55`` and ``JRA55do``.   For those configurations, the ``atm_data_dir`` should be set to ${CICE_DATA_root}/forcing/${grid}/[JRA55,JRA55do'] and the filenames should be of the form [JRA55,JRA55do]_${grid}_03hr_forcing${atm_data_version}_yyyy.nc where yyyy is the forcing year.  Those files should be placed under ``atm_data_dir``/8XDAILY.  ``atm_data_version`` is a string defined in namelist that supports versioning of the forcing data.  ``atm_data_version`` could be any string including the null string, but it could also be something like _yyyymmdd to indicate the data the forcing data was generated.

apcraig:
I ran a small test suite with gx3, gx1, and tx1 cases and things seems to work fine.  tx1 cases change answers, as expected.  Other grids are bit-for-bit.  I will run a further test once final changes are in.

Also, I noticed that this,

```
         !if (debug_forcing .and. (my_task == master_task)) then
         !   write(nu_diag,*) subname,cnt,exists,trim(uwind_file)
         !endif
```

now has a "debug_forcing" which I think is a good idea, these could be commented out again.  And we just need to update the documentation one more time.  Otherwise, I think this is all fine.

Just to ask, did someone review the data in the forcing files as well as the results from the tx1 model runs.  It all look OK, right?  I haven't done that myself.

daveh150:
@apcraig agree on your point in regard to developer guide docs and atm_data_version can be general. I added your description, thank you for sending.

I also uncomment the debug code you highlighted with the check for debug_forcing. 

I will re-download the forcing files to verify they are good. It would be good for someone else to double check me, my mistake is what got us here in the first place. :-( 

Thank you!

apcraig:
Thanks @daveh150.  There are a couple problems with the documentation.  See https://cice-consortium-cice--876.org.readthedocs.build/en/876/developer_guide/dg_forcing.html

The  :ref:tabnamelist is not being linked thru.  Can you check the syntax/formatting.  Also, I think we should probably change atm_data_dir/8XDAILY to ``atm_data_dir``/8XDAILY.

I also had a look at some random fields in 2005 and 2008 tx1 forcing files.  They seem OK to me.  Then I checked history output from day 8 of a 10 day tx1 CICE run.  And those looked OK to me too, in general.  The initial condition is 'internal' so the ice model isn't producing a great result after 8 days, but the forcing fields seem reasonable in the ice history file.

phil-blain:
This PR addresses https://github.com/CICE-Consortium/CICE/issues/872, so we could add 

~~~
Closes: https://github.com/CICE-Consortium/CICE/issues/872
~~~

in the PR description and the issue will be closed automatically.

daveh150:
Thank you Philippe! I added “Closes #872” to the PR description. Very cool, I was not aware of this feature. 


David

phil-blain:
Another thing: before we merge, we should change the PR title (it will be used as the commit message title) since it references "jra55date" which is from an earlier version of the PR. Maybe something like this would be more accurate:

> ice_forcing: add 'atm_data_version' to allow versioning JRA55 forcing

apcraig:
I updated the title and I did some testing, things seem to be working as expected.  This can be merged then we'll need to migrate the dataset to zenodo.

apcraig (approved):

apcraig:
Anyone else want to review or comment before I merge?

@dabail10, @daveh150 can you migrate the data to zenodo when you have a chance.  Thanks!

eclare108213 (approved):

dabail10:
I am working on the tar files now.

daveh150:
Thank you @apcraig for editing the title and all your time and patience with the commits.

@dabail10 thank you for handling the zenodo upload. I tried once and got on the order of 150-200 kb/s. It was going to take days to upload the files. I hope your connection is better than that.

dabail10:
Yeah, I will do the upload from the office tomorrow. This is the fastest.

dabail10:
The zenodo uploads are done and the input data table has been updated. I put version 6.4.2+, but I guess it will technically be 6.5?

apcraig:
Thanks @dabail10.  I think formally, it's not 6.4.2+ because 6.4.2 doesn't support these new files.  It will be the next version +, I think that should be 6.5, but that has not yet been officially decided.

Issue #897, opened 2023-10-26
NLON, ELAT not computed when TLAT, TLON, ANGLET on grid file


dabail10:
This was raised in the CICE forum. We should fix this before the next release. I can look into it.

Dave

dabail10:
https://bb.cgd.ucar.edu/cesm/threads/nlon-raises-error-in-ice_history_write.8938/#post-51905

apcraig:
How should we fix this?  Should we NOT write the N/E coords with B grid or should we initialize N/E coords with B grid?  I can create a PR unless someone else wants to take it.  Will try to fix today.

apcraig:
I think I'll go with not writing the N/E coords with B grid as suggested by @TillRasmussen.  Let me know if others prefer another solution.

Why aren't our tests catching this??  Do we need to add some tests or some other kind of IC file to our test suite?  @phil-blain?

phil-blain:
> I think I'll go with not writing the N/E coords with B grid as suggested by @TillRasmussen. Let me know if others prefer another solution.

I agree with this approach. It also makes the history files a little less voluminous, which is good.

> Why aren't our tests catching this?? Do we need to add some tests or some other kind of IC file to our test suite? @phil-blain?

As I wrote in the forum, I think this is because all our grids have only `ULON`, `ULAT` and `ANGLE`, and so the corresponding `TLON`, `TLAT` and `ANGLET` are computed by the code (in `ice_grid::Tlatlon`). 

If the grid has `ANGLET`, then the code assumes that it also has `TLON`, `TLAT` and `Tlatlon` is not called. Since the N- and E- locations lat/lon arrays are also computed in that subroutine, then then are not set before the history code writes them to the history files.

apcraig:
What if you have an input grid that has TLON, TLAT, and ANGLET but then you are running on the C-grid.  I think this is not a B grid issue.  I think we have to compute the E/N coords if ANGLET is included on the grid file.  Am I missing something?

phil-blain:
This is my understanding as well. I would retitle the issue

> NLON, ELAT, etc. not computed under l_readCenter=.true.

apcraig:
We should add a grid that has TLON, TLAT, ANGLET to our test suite.  Anyone have a grid we could use?  If someone can provide a file, I'll add some test cases.  We could also create one from the gx3, maybe I'll try to do that.

phil-blain:
I think it should be easy to create one for gx3 from the history output.

[...]

Issue #613, opened 2021-06-22
Write grid variables only once.


dabail10:
One thing that has annoyed me for years is the missing values in land (from land block elimination) in the grid variables. Several users have requested that we fill in the land blocks with valid values. Mostly for analysis after the fact. One way to do this is how they do it in CESM-POP. There is a single file with the string "once" in it, where the time constant variables are written only once. This would be written in ice_grid.F90 when the global arrays are read in. This is still CF-compliant in the sense that this file would always accompany the history files. It also would save some space not writing all the grid information to every history file.

apcraig:
I am looking at this now.  The current implementation

- reads global ULAT and KMT
- creates a decomposition with land block elimination
- reads a handful of global grid fields ULAT, ULON, KMT, HTE, HTN from the grid file
- computes all the CICE grid variables in parallel on the decomposed grid

So, 
- it will be difficult to generate a CICE grid file that includes the land block elimination values.  We would have to compute all those variables on a global grid then scatter them.  That would be a significant refactor of the current implementation.
- we could write out only ULAT, ULON, and KMT to the "grid output file" but we already have that in the grid input file and that probably does not add much value

Some ideas,
- we could still add a separate grid output file with all the CICE grid variables.  they would still have land block elimination (so missing data), but it would allow the users to have either a single grid file or grid variables in all the history files or neither or both.
- users could generate a one-time grid file by turning on the "grid output file" and also setting "distribution_wght='blockall'" in namelist.  That namelist option prevents land block elimination.  This could be done for one run in production (or pre-production) to generate a complete grid file while still allowing land block elimination during the run for performance.

So, by adding an ability to write a separate grid output file, keeping the field-by-field control of grid variables on history files, and the ability to run with and without land block elimination, users should be able to do whatever they want and have the data they need.  Thoughts?

anton-seaice:
I believe the main issue we have with the current setup is that concatenating output files is slow (when using `xarray.concat` or related methods) due to xarray checking for consistency where the same variable exists in multiple files being concatenated. In that way, just writing the grid to a different file solves the problem.

I will endeavor to do some experimenting on how much the data in the eliminated blocks matters also and confirm. I suspect these variables are more important for coordinates (all the lat/lons) and less important for the rest of the grid (lengths/angles/areas).

dabail10:
[...]

PR #899, opened 2023-10-26
Split N/E grid computation out of Tlonlat, create NElonlat subroutine.

apcraig:
## PR checklist
    Split N/E grid computation out of Tlonlat, create NElonlat subroutine.
- [X] Developer(s): 
    apcraig
    early tests show bit-for-bit, running full test suite now
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Closes https://github.com/CICE-Consortium/CICE/issues/897

When TLON, TLAT, ANGLET are on the CICE grid, Tlonlat is NOT called.  This meant N and E grid info was never computed.  This would fail during history writing with invalid values in N and E grid arrays.  And it would also cause problem if the C-grid were run with this type of CICE grid.

There are no test grids that have TLON, TLAT, ANGLET on them, so this error was not found in standard test suites.  This was detected by users.

Also fixes some indentation and write statements.

### DIFF
--- a/cicecore/cicedyn/infrastructure/ice_grid.F90
+++ b/cicecore/cicedyn/infrastructure/ice_grid.F90
@@ -675,36 +675,37 @@
       if (trim(grid_type) == 'cpom_grid') then
          ANGLET(:,:,:) = ANGLE(:,:,:)
       else if (.not. (l_readCenter)) then
-          ANGLET = c0
+         ANGLET = c0
 
-      !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block, &
-      !$OMP                     angle_0,angle_w,angle_s,angle_sw)
-      do iblk = 1, nblocks
-         this_block = get_block(blocks_ice(iblk),iblk)
-         ilo = this_block%ilo
-         ihi = this_block%ihi
-         jlo = this_block%jlo
-         jhi = this_block%jhi
+         !$OMP PARALLEL DO PRIVATE(iblk,i,j,ilo,ihi,jlo,jhi,this_block, &
+         !$OMP                     angle_0,angle_w,angle_s,angle_sw)
+         do iblk = 1, nblocks
+            this_block = get_block(blocks_ice(iblk),iblk)
+            ilo = this_block%ilo
+            ihi = this_block%ihi
+            jlo = this_block%jlo
+            jhi = this_block%jhi
 
-         do j = jlo, jhi
-         do i = ilo, ihi
-            angle_0  = ANGLE(i  ,j  ,iblk) !   w----0
-            angle_w  = ANGLE(i-1,j  ,iblk) !   |    |
-            angle_s  = ANGLE(i,  j-1,iblk) !   |    |
-            angle_sw = ANGLE(i-1,j-1,iblk) !   sw---s
-            ANGLET(i,j,iblk) = atan2(p25*(sin(angle_0)+ &
-                                          sin(angle_w)+ &
-                                          sin(angle_s)+ &
-                                          sin(angle_sw)),&
-                                     p25*(cos(angle_0)+ &
-                                          cos(angle_w)+ &
-                                          cos(angle_s)+ &
-                                          cos(angle_sw)))
+            do j = jlo, jhi
+            do i = ilo, ihi
+               angle_0  = ANGLE(i  ,j  ,iblk) !   w----0
+               angle_w  = ANGLE(i-1,j  ,iblk) !   |    |
+               angle_s  = ANGLE(i,  j-1,iblk) !   |    |
+               angle_sw = ANGLE(i-1,j-1,iblk) !   sw---s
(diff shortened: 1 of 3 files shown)


phil-blain:
OK so we still write NLON et al. to the history files, even when using the B grid. This is tangential to the bug being fixed but would still be worth it, in my opinion. 

I'll take a closer  look at the diff later.

dabail10 (approved):
Oh, I see the problem now. I was focused on the C-grid variables.

apcraig:
I agree, writing the N, E grid info with the B-grid is a separate issue.  I understand the reasons for not writing that info with the B-grid.  But I sort of favor keeping it for now.  It's just a few 2d arrays, it provides some info that might still be useful with the B-grid.  And other variables could depend on those dimensions.  For instance, if we decide to start computing and/or writing data on the N, E locations in the future with the B-grid, we'll need/want that info.  I understand we can add it back later too.  I guess I'm willing to turn off the B-grid writing of N, E.  At the same time, should we turn off the computation of the N, E variables if it's a B-grid?

There is a separate idea hanging out which is to write a history file just for grid variables that could be controlled separately from the standard history files.  For instance, it could be written just once during a long run, reducing the amount of data on history files even further.  I favor focusing on this feature rather than fussing over the N, E grid variables on the B-grid right now.  See https://github.com/CICE-Consortium/CICE/issues/613.

If there is general agreement that we should not compute and/or write the N, E grid variables when running the B-grid, I'll implement that.  Feel free to push for that :)

daveh150:
I favor to not compute/write unused N,E when using B-grid. 

RE not writing grid in history files:  I am hesitant to remove the lat/lon from the history files. I often will look at a history file in Panoply for quick analysis and plots. It is nice to have projected views to include coastline overlays. If the lat/lon is not in the files it would prevent a projected view in Panoply or similar plotting tools. Could this be controlled with namelist options?

apcraig:
Thanks for your input @daveh150.  

Regarding the separate grid file, that would be controlled by namelist.  I imagine there might be two namelist, one to turn on writing to the history file, one to turn on writing to a separate grid file.  Each could be toggled separately to give you exactly the behavior preferred.  And the default would probably be to maintain the current situation.

eclare108213 (approved):

apcraig:
See #897.  I am going to leave the N/E grid initialization and history writing as is.  We can revisit this at some point in the future.

apcraig:
But I did create a gx3 netcdf grid/kmt file with the TLON, TLAT, and ANGLET variables on it and I will add that as a test case.

apcraig:
I just added a gx3 netcdf grid/kmt test.  This is not bit-for-bit with the standard gx3 runs but is definitely just roundoff initially (as expected).  I also verified that all the NE grid variables are defined on the history file and reasonable now.  Prior to this PR, some of those variables were undefined.  This fixes the problem as far as I can tell.

apcraig:
@dabail10, I propose we add the new gx3 netcdf grid and kmt files to zenodo as this is now part of the test suite.  Not sure the best way to do that, but let me know if I can help.  The two files (grid_gx3t.nc and kmt_gx3t.nc) are here

cheyenne:  /glade/p/cesm/pcwg_dev/CICE_data/grid/gx3/

dabail10:
These have been added to Zenodo and the table updated.

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -2581,12 +2642,10 @@ subroutine Tlatlon
...
-!         if (nx_block > 5+2*nghost .and. ny_block > 5+2*nghost) then
-         write(nu_diag,*) 'min/max ULON:', y1*rad_to_deg, y2*rad_to_deg
-         write(nu_diag,*) 'min/max ULAT:', y3*rad_to_deg, y4*rad_to_deg
-!         endif
-         write(nu_diag,*) 'min/max TLON:', x1*rad_to_deg, x2*rad_to_deg
-         write(nu_diag,*) 'min/max TLAT:', x3*rad_to_deg, x4*rad_to_deg
+         write(nu_diag,*) subname,' min/max ULON:', y1*rad_to_deg, y2*rad_to_deg
+         write(nu_diag,*) subname,' min/max ULAT:', y3*rad_to_deg, y4*rad_to_deg
+         write(nu_diag,*) subname,' min/max TLON:', x1*rad_to_deg, x2*rad_to_deg
+         write(nu_diag,*) subname,' min/max TLAT:', x3*rad_to_deg, x4*rad_to_deg
```
phil-blain, line 2648: shouldn't these prints logically be in `Tlatlon`  instead of `NElatlon` ?

phil-blain (commented):
I did not have time to look at the code last week, but took a look this morning and left a few comments.

I'm not sure it warrants a follow-up PR though...

PR #906, opened 2023-11-14
ice_history: refactor CMIP history variables

phil-blain:

## PR checklist
    refactor the CMIP history code by adding two new fields to our `ice_hist_field` type.
- [X] Developer(s): 
    me
  - Base suite is bit for bit (we only check restarts, and this PR modifies the ice_history code).
  - I ran two one-month cases with `f_CMIP= 'm'` and verified that the history outputs are bit for bit between the two runs, if we remove the `history` global NetCDF  attribute (`ncatted -O -h -a history,global,d,, ${test_file}`)
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

I'm working on modifying the history code to write outputs using the correct mask for each variables, instead of hardcoding the use of the `tmask` as is currently done. In doing so I noticed several things that could be improved in that code and thus refactored it. This is the first of two such refactoring PRs.

Details are in each commit message.


cicecore/cicedyn/analysis/ice_history_shared.F90
```diff
@@ -81,6 +81,8 @@ module ice_history_shared
           real (kind=dbl_kind) :: conb    ! additive conversion factor
           character (len=1) :: vhistfreq  ! frequency of history output
           integer (kind=int_kind) :: vhistfreq_n ! number of vhistfreq intervals
+          logical (kind=log_kind) :: avg_ice_present = .false. ! only average where ice is present
+          logical (kind=log_kind) :: mask_ice_free_points  = .false. ! mask ice-free points
```
apcraig, line 85: Is it useful/necessary to initialize the flags here?  No other ice_hist_field variables are initialized here, and they will be initialized in define_hist_field if needed.  Think it might be more consistent and still robust to not initialize here.

apcraig (commented):
This looks like a very nice cleanup!  Just one comment, but otherwise looks good to me.

apcraig:
@dabail10, please have a look if you get a chance and make sure you're happy with this update.

dabail10 (approved):
I like this syntax. A lot cleaner.

cicecore/cicedyn/analysis/ice_history_shared.F90
phil-blain: I agree that since the default values are set using the local variables `l_avg_ice_present` and `L_mask_ice_free_points` in define_hist_field, there is no absolute need to initialize them here.

I can remove that for consistency.

apcraig: Thanks @phil-blain.

phil-blain: Done and force-pushed.

apcraig:
I ran a full test suite on cheyenne and all the tests pass including regression.  This includes the io_suite which tests several features in the history files.

apcraig (approved):

PR #916, opened 2023-12-01
Small change for wave coupling. Wasn't detecting missing values.

dabail10:

## PR checklist
Add check for missing values in wave spectrum.
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#b313d4f82a87bb6e14c26b8b2fe453ddaa70a9df
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
Just a simple check on the wave spectrum.

### DIFF
--- a/cicecore/cicedyn/general/ice_step_mod.F90
+++ b/cicecore/cicedyn/general/ice_step_mod.F90
@@ -676,7 +676,7 @@
          if (tmask(i,j,iblk)) then
 
          ! significant wave height for FSD
-         if (tr_fsd) &
+         if (tr_fsd .and. wave_spectrum(i,j,1,iblk) < 1.0+10_dbl_kind) &
          wave_sig_ht(i,j,iblk) = c4*SQRT(SUM(wave_spectrum(i,j,:,iblk)*dwavefreq(:)))
 
          call icepack_step_therm2(dt=dt, ncat=ncat, &


cicecore/cicedyn/general/ice_step_mod.F90
```diff
@@ -676,7 +676,7 @@ subroutine step_therm2 (dt, iblk)
          if (tmask(i,j,iblk)) then
 
          ! significant wave height for FSD
-         if (tr_fsd) &
+         if (tr_fsd .and. wave_spectrum(i,j,1,iblk) < 1.0+10_dbl_kind) &
```
eclare108213, line 679: Is wave_spectrum always nonnegative?

erinethomas: @eclare108213  - yes, it should be always non-negative.

dabail10: I'm pretty sure. @lettie-roach can confirm or deny this. The main thing is that in CESM at least, it assigns a missing value of 1.0e30 which is different than the typical coupler missing value of 1.0e36. I think we need to fix this eventually either in the WAVEWATCH III cap.

phil-blain: Shouldn't we define a `ww3_missing_value` parameter in the mean time, and use that, instead of hardcoding this number here ?

dabail10: Perhaps. I am mainly just trying to catch any missing value above 1.0e10.

apcraig: Looks like wave_spectrum is always allocated (alloc_arrays_column) and but NOT always set (depends on tr_fsd).  So the above logic could cause problems with a subset of compilers and debug checks.  It would be safer to implement this using two if statements

if (tr_fsd) then
   if (wave_spectrum(i,j,1,iblk) < 1.0+10_dbl_kind) &
      wave_sig_ht(i,j,iblk) = c4*SQRT(SUM(wave_spectrum(i,j,:,iblk)*dwavefreq(:)))
endif

dabail10:
I am rethinking this and will close it for now.

PR #924, opened 2024-01-04
Add vorticity as a diagnostic output

JFLemieux73:

## PR checklist
   Added calculation of vorticity as a diagnostic for B and C grids. 
- [x] Developer(s): 
    @JFLemieux73 
    @eclare108213 
    I did the end-to-end procedure (base_suite):

385 measured results of 385 total results
385 of 385 tests PASSED
0 of 385 tests PENDING
0 of 385 tests MISSING data
0 of 385 tests FAILED

    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

### DIFF
--- a/cicecore/cicedyn/analysis/ice_history.F90
+++ b/cicecore/cicedyn/analysis/ice_history.F90
@@ -3,7 +3,8 @@
 !
 ! The following variables are currently hard-wired as snapshots
 !   (instantaneous rather than time-averages):
-!   divu, shear, sig1, sig2, sigP, trsig, mlt_onset, frz_onset, hisnap, aisnap
+!   divu, shear, vort, sig1, sig2, sigP, trsig, mlt_onset, 
+!   frz_onset, hisnap, aisnap
 !
 ! Options for histfreq: '1','h','d','m','y','x', where x means that
 !   output stream will not be used (recommended for efficiency).
@@ -597,6 +598,7 @@
       call broadcast_scalar (f_strength, master_task)
       call broadcast_scalar (f_divu, master_task)
       call broadcast_scalar (f_shear, master_task)
+      call broadcast_scalar (f_vort, master_task)
       call broadcast_scalar (f_sig1, master_task)
       call broadcast_scalar (f_sig2, master_task)
       call broadcast_scalar (f_sigP, master_task)
@@ -1319,6 +1321,11 @@
              "strain rate (shear)",                                  &
              "none", secday*c100, c0,                                &
              ns1, f_shear)
+
+         call define_hist_field(n_vort,"vort","%/day",tstr2D, tcstr, &
+             "strain rate (vorticity)",                                  &
+             "none", secday*c100, c0,                                &
+             ns1, f_vort)
 
          select case (grid_ice)
          case('B')
@@ -2623,7 +2630,7 @@
          if (f_strength(1:1)/= 'x') &
              call accum_hist_field(n_strength,iblk, strength(:,:,iblk), a2D)
 
-! The following fields (divu, shear, sig1, and sig2) will be smeared
+! The following fields (divu, shear, vort, sig1, and sig2) will be smeared
 !  if averaged over more than a few days.
 ! Snapshots may be more useful (see below).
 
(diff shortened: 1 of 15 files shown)


JFLemieux73:
Because of the different location of u and v, the calculation of vorticity for the B and C grids are not exactly the same. The vorticity fields are however similar. Here is an example for the B-grid:

B_vort_2005-01-02-00000

and the C-grid:
C_vort_2005-01-02-00000

JFLemieux73:
I forgot to mention that I am adding vorticity as an output in order to use Nils Hutter's linear kinematic features (LKFs) detection algorithm. Nils' tool uses vorticity to identify conjugate fault lines.

apcraig (changes requested):
This looks reasonable.  Please add f_vort to set_nml.histall and set_nml.histdbg, set to same values as f_divu there.  Those namelist are used to test the history mechanisms more comprehensively.

Lots of work just to add one history field, thanks.

apcraig:
I will test this on derecho when it's ready.  I can run the full io_suite which will cover binary, netcdf, and pio options.  Please let me know when you'd like me to do that.

JFLemieux73:
Hi @apcraig,

I added f_vort to the two set_nml files. Please go ahead with the testing.

dabail10 (commented):
Do we really need a vorticity variable? Is it used in any of the dynamic calculations? I would perhaps just do this as a temporary in ice_history.F90. This is the approach I've taken with other diagnostic variables that are not used in calculations.

JFLemieux73:
This does not add a lot of computations because it is done after the EVP subcycling loop. I would keep the computation of vorticity together with divergence and shear (as proposed in this PR).  The shear strain rate invariant (a diagnostic) is currently calculated in the deformations subroutine. If we move the vorticity to ice_history.F90 then we should do the same for the shear. I suggest we keep both in the deformations subroutine. @eclare108213 what do you think about this?

apcraig:
I ran a full test suite on derecho with intel, gnu, cray.  Results look good.  If things get updated again, I'm happy to retest.

apcraig (approved):

eclare108213 (approved):
I agree that vort belongs with divu and shear, conceptually.  It's not needed for the dynamics calculation but this is fine.  I also agree that this was a lot of work for a history variable!

JFLemieux73:
Can we merge this PR?

cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -1320,6 +1322,11 @@ subroutine init_hist (dt)
              "none", secday*c100, c0,                                &
              ns1, f_shear)
 
+         call define_hist_field(n_vort,"vort","%/day",tstr2D, tcstr, &
+             "strain rate (vorticity)",                                  &
+             "none", secday*c100, c0,                                &
+             ns1, f_vort)
+
```
phil-blain, line 1329: Should we specify at which location `vort` is defined ? I see shear, divu and strength do not mention it , but most other fields do.

cicecore/cicedyn/dynamics/ice_dyn_eap.F90
```diff
@@ -1357,6 +1366,13 @@ subroutine stress_eap  (nx_block,   ny_block,       &
                  (tensionne + tensionnw + tensionse + tensionsw)**2 &
                 +  (shearne +   shearnw +   shearse +   shearsw)**2)
 
+            ! vorticity
+            dvdxn = dyU(i,j)*vvel(i,j) - dyU(i-1,j)*vvel(i-1,j)
+            dvdxs = dyU(i,j-1)*vvel(i,j-1) - dyU(i-1,j-1)*vvel(i-1,j-1)
+            dudye = dxU(i,j)*uvel(i,j) - dxU(i,j-1)*uvel(i,j-1)
+            dudyw = dxU(i-1,j)*uvel(i-1,j) - dxU(i-1,j-1)*uvel(i-1,j-1)
+            vort(i,j) = p5*tarear(i,j)*(dvdxn + dvdxs - dudye - dudyw)
+
```
phil-blain, line 1375: Here the computation of vorticity is the same as in `ice_dyn_shared::deformations`, right ?

If so, I would prefer we compute the vorticity in a separate subroutine, and call that subroutine from `ice_dyn_eap::stress_eap` and  `ice_dyn_shared::deformations`. This would reduce code duplication.

JFLemieux73: But then it is the same thing for divu and shear...the same computations are done in ice_dyn_eap and ice_dyn_shared. It has been like this for a long time...I say we keep it simple the way it is done right now.

apcraig:
I just resolved some conflicts with the last PR and will run a short test suite to confirm no problems.  Once complete, I'll merge unless there is further discussion.

apcraig:
I ran the quick_suite and io_suite on derecho with cray, gnu, intel.  All tests pass bit-for-bit.  Will merge now.

PR #928, opened 2024-01-26
Update IO formats and add new IO namelist controls

### LINKED ISSUE #862: allow namelist control of PIO options
When using PIO, the non-CESM user is limited to a specified configuration of PIO. Either netcdf or pnetcdf is allowed as the iotype, the rearranger is limited to box and the stride and number of iotasks is limited to 4 and nprocs/4, respectively. 

Testing in UFS has shown that better IO performance at high PE counts can potentially be achieved using either the subset rearranger and specifying PE dependent stride and iotasks but this requires allowing run-time configuration instead of compile-time configuration of PIO.

I have a working feature [branch](https://github.com/DeniseWorthen/CICE/tree/feature/add_pioopts) using an optional new namelist setting ``pio_options``. Currently it allows control of 4 settings (ie, none of the rearranger options) specified as follows:

``pio_options = iotype, rearranger type, iostride, numiotasks``

I am debugging an issue with *reading* a restart file using the subset rearranger, although *writing* a restart using the same options works without issue.

### LINKED ISSUE #914: Support netcdf-4 compression & chunking
Currently support for netcdf-4 features in CICE is pretty limited. Netcdf supports chunking and compression.

When implemented, #862 will let CICE standalone output netcdf-4 files when it is built using the `io_pio` directory. (The nuopc driver also has options to configure netcdf-4)

It would be good to add support for the extra features like chunking and compression (we would use in it in the ACCESS model). This would need a namelist option (e.g. `netcdf4_options`) which would be something like:

`netcdf4_options = deflate_level, chunksize_x, chunksize_y `

This option would only be relevant if the iotype  one of the netcdf4 types (`netcdf4c` or `netcdf4p`), otherwise it would be ignored.

For code changes, in all the places `pio_def_var()` is used we would need to add:
- add `pio_def_var_deflate()` if `deflate_level` is set (https://ncar.github.io/ParallelIO/group___p_i_o__def__var.html)
- `pio_def_var_chunking()` if `chunksize_x` or `chunksize_y` are set (https://ncar.github.io/ParallelIO/group___p_i_o__def__var.html)

This change would mean the io_pio build is ahead of the io_netcdf build. Builds using the io_netcdf files would only be using netcdf-3 / classic.

apcraig:
## PR checklist
    Update IO formats and add new IO namelist controls
- [X] Developer(s): 
    apcraig, anton-s
    Test results look good.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#1619d6ca92028d1c0341be06081d226882a0739a.  There is a problem in the bit-for-bit hdf5 restart files generated from PIO.  The model restarts exactly, but the check that compares restart files fails because there is some non-important data in the PIO hdf5 files that is not bit reproducible.  Have discussed this with PIO folks and they say it's OK.  I don't fully agree, netCDF hdf5 does not have this problem.  But it means the PIO hdf5 restart tests will fail the "cmp" validation even though they restart the model exactly.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Closes #914
Closes #862 

This provides new features for CICE IO both thru netCDF and PIO.  New namelist are added to control history and restart format, hdf5 compression and chunking, the PIO rearranger, and PIO IO task control.  Separate controls are provided for history and restart files.  The namelist changes are for

  _history_format, restart_format
  history_rearranger, restart_rearranger
  history_iotasks, history_root, history_stride, restart_iotasks, restart_root, and restart_stride
  history_chunksize, history_deflate, restart_chunksize, restart_deflate._

In particular,

- Update restart_format and history_format options to 'cdf1', 'cdf2', 'cdf5', 'hdf5', 'pnetcdf1', 'pnetcdf2', 'pnetcdf5', 'default'.  The old options, 'default', 'pio_netcdf', and 'pio_pnetcdf' are still supported and backwards compatible with lcdf64, but are deprecated and no longer documented.  The old options and old namelist lcdf64 are covered by the new options.  Support of the old options should be removed in the future.  Note that some problems were discovered when opening files with hdf5 format but reading non-hdf5 files with a spack built PIO/netCDF.  As a result, the format specified for the restart read is always 'cdf1' which provides flexibility and robustness across software installs, although it may result in serial reads of hdf5 files when a parallel read could be done.
- Deprecate lcdf64 namelist.  This namelist is no longer needed and is covered by the new restart_format and history_format options.  The namelist still exists and is backwards compatible with the old 'default', 'pio_netcdf', and 'pio_pnetcdf' format options, but is no longer documented.  This should be removed in the future.
- Add new namelist to control PIO pe/task setup (iotasks, root, stride) for history and restart.  These settings control the PIO IO tasks.  The root, stride, and iotasks are consistent with the MPI communicator.  root=0 is the first MPI task.  These control PIO IO performance and are usually a function of things like the IO and node hardware.  See PIO for more information.  CICE computes PIO iotask, root, and stride defaults for cases where -99 is passed in for some or all of these namelist.  Those defaults are somewhat constrained by a bug in PIO, https://github.com/NCAR/ParallelIO/issues/1986.  The current implementation avoids the bug by limiting the iotasks for some MPI task counts.  This is noted in ice_pio.F90.
- Add new namelist to control PIO rearranger (rearranger) for history and restart.  Supports 'box', 'subset', and 'default'.  These control how PIO rearrangment is carried out.  default is equivalent to box and the box generally performs better.  See PIO for more information.
- Add new namelist to support hdf5 compression and chunking (deflate, chunksize) for history and restart.  The deflate controls file compression and is an integer between 0 and 9 where 0 means no compression and 9 is maximum compression.   Generally, the higher the number, the slower the IO and the smaller the file, but the optimal setting depends on the contents of the file.  Chunksize provides a performance control for the hdf5 parallel writes.  It is a 2d array and is associated with the size of the piece of the array written by hdf5.  hdf5 can be read and written in parallel, but that depends on how netCDF and PIO are built.  Note that prior version of PIO, including PIO1, do not support the hdf5 compression and chunking thru the PIO interface.
- Add new namelist settings (set_nml files) and update the io_suite to cover the new IO options.  Remove old namelist settings associated with the deprecated format options and the lcdf64 namelist.  These deprecated feature are no longer tested.
- Update documentation to add new namelist and IO features.
- Update the nuopc/cmeps driver code to support the new features.
- Update the default ice_in to add the new namelist.
- Update the derecho netcdf module to a version that supports hdf5.
- Clean up some code formatting (indentation)


apcraig:
This is a draft.  There are still some changes to come.  Creating a draft PR to facilitate development, support early review, and allow for quick review of documentation changes.

anton-seaice:
Thanks Tony - I will go through in detail next week.

One thing I have been thinking about:
- I think we should only be calling pio_init once, and then using that set of mpi threads for all restart/history files. I think that is how the library is intended to be used, but haven't found a good way to confirm that.
- At the moment, we call pio_init once per file written/read, so as a minimum we should call pio_finalize when we close the file.

anton-seaice:
I think 

https://github.com/apcraig/CICE/blob/d2799fc643a8113f20b8bf9b5daa8239807c9006/configuration/scripts/options/set_nml.iopio2p#L2

history_format = 'pio_netcdf'

needs updating / has a typo and should be `pio_pnetcdf`

anton-seaice (commented):
Hi Craig. Looks good. A few suggestions

anton-seaice:
I'm still working on the chunking ... I realised the module that I was going to get `pio_def_var_chunking` from only added the call in very recent versions (https://github.com/NCAR/ParallelIO/blame/6539ef05ae7584ec570a56fdab9f7dfb336c2b80/src/flib/pio.F90#L96), even though the functionality has existed for a long time. So ill need to figure out a better way to call the chunking definition.

apcraig:
> I think
> 
> https://github.com/apcraig/CICE/blob/d2799fc643a8113f20b8bf9b5daa8239807c9006/configuration/scripts/options/set_nml.iopio2p#L2
> 
> history_format = 'pio_netcdf'
> 
> needs updating / has a typo and should be `pio_pnetcdf`

I noticed this issue too.  Our tests weren't quite doing what I expected, but we were still testing pio_pnetcdf via restarts for pio2.  Anyway, this file will be removed before this PR is finalized.  I'm keeping a few older files temporarily to do regression testing against the older options, but it's just while the PR is evolving.

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -230,8 +244,22 @@ setup_nml
...
+   "``restart_format``", "``cdf1``", "read/write restart files with netcdf cdf1 (netcdf3-classic) format", "``cdf2``"
+   "", "``cdf2``", "read/write restart files with netcdf cdf2 (netcdf3-64bit-offset) format", ""
+   "", "``cdf5``", "read/write restart files with netcdf cdf5 (netcdf3-64bit-data) format", ""
+   "", "``default``", "read/write history files in default format", ""
+   "", "``hdf5``", "read/write restart files with netcdf hdf5 (netcdf4) format", ""
+   "", "``pio_pnetcdf``", "read/write restart files with pnetcdf in pio, deprecated", ""
+   "", "``pio_netcdf``", "read/write restart files with netcdf in pio, deprecated", ""
+   "", "``pnetcdf1``", "read/write restart files with pnetcdf cdf1 (netcdf3-classic) format", ""
+   "", "``pnetcdf2``", "read/write restart files with pnetcdf cdf2 (netcdf3-64bit-offset) format", ""
+   "", "``pnetcdf5``", "read/write restart files with pnetcdf cdf5 (netcdf3-64bit-data) format", ""
```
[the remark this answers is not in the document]
apcraig: Good point, I'll fix this for restart and history documentation.

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -230,8 +244,22 @@ setup_nml
...
+   "``restart_iotasks``", "integer", "pe io tasks for restart output with restart_root and restart_stride, -99=internal default", "-99"
+   "``restart_rearranger``", "box", "box io rearranger option for restart output", "default"
+   "", "default", "internal default io rearranger option for restart output", ""
+   "", "subset", "subset io rearranger option for restart output", ""
+   "``restart_root``", "integer", "pe root task for restart output with restart_iotasks and restart_stride, -99=internal default", "-99"
+   "``restart_stride``", "integer", "pe stride for restart output with restart_iotasks and restart_root, -99=internal default", "-99"
```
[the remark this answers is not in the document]
apcraig: I thought about that, but then I thought we might reuse these namelist for some other IO options and didn't necessarily want to say "pio".  But maybe I should, we have no plans to extend usage at this point.

cicecore/cicedyn/general/ice_init.F90
```diff
@@ -1232,6 +1259,111 @@ subroutine input_data
...
+         elseif (restart_format == 'pio_pnetcdf') then
+            if (my_task == master_task) then
+               write(nu_diag,*) subname//' WARNING: restart_format='//trim(restart_format)// &
+                                ' is deprecated, please update namelist settings'
+            endif
+            restart_format = 'pnetcdf1'
+         endif
+      endif
+
       if (ktransport <= 0) then
```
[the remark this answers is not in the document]
apcraig: It's happening later in the code.  If a format is selected that isn't compatible with the ICE_IOTYPE, it will abort when it's used in the io code, not during initialization.  We could add something in the initialization, but that would require adding some new CPPs that trigger from the ICE_IOTYPE, and I'm reluctant to do that just to move an abort earlier.

apcraig:
> I'm still working on the chunking ... I realised the module that I was going to get `pio_def_var_chunking` from only added the call in very recent versions (https://github.com/NCAR/ParallelIO/blame/6539ef05ae7584ec570a56fdab9f7dfb336c2b80/src/flib/pio.F90#L96), even though the functionality has existed for a long time. So ill need to figure out a better way to call the chunking definition.

OK, so the concern is the chunking stuff will abort in PIO if someone is linking to an older version of PIO?  A quick workaround might be to have some default chunking size (say -99).  If that value is set, then don't use the chunking implementation, use the non-chunking one.  And when the chunking call returns with an error, add an extra warning message that this could be caused by an older version of PIO and to avoid the error, set the chunking value to -99.  Just an idea.  And in that case, I'd set the default to -99 in code and namelist.  Make folks have to turn it explicitly.

anton-seaice:
> > I'm still working on the chunking ... I realised the module that I was going to get `pio_def_var_chunking` from only added the call in very recent versions (https://github.com/NCAR/ParallelIO/blame/6539ef05ae7584ec570a56fdab9f7dfb336c2b80/src/flib/pio.F90#L96), even though the functionality has existed for a long time. So ill need to figure out a better way to call the chunking definition.
> 
> OK, so the concern is the chunking stuff will abort in PIO if someone is linking to an older version of PIO? 

It doesn't compile because the routine is not defined.

My suggestion is to import it like `use pio_nf, only: pio_def_var_chunking` which seems programmatically the same as the new versions where its included through `use pio`.

I will wrap the chunking and compression with `#ifndef USE_PIO1` too.

> A quick workaround might be to have some default chunking size (say -99). If that value is set, then don't use the chunking implementation, use the non-chunking one. And when the chunking call returns with an error, add an extra warning message that this could be caused by an older version of PIO and to avoid the error, set the chunking value to -99. Just an idea. And in that case, I'd set the default to -99 in code and namelist. Make folks have to turn it explicitly.

At the moment I've just wrapped the chunking and compression in if `_format='hdf5'` checks, not quite as comprehensive as setting to -99.

apcraig:
@anton-seaice, do you think we should apply the chunking only if chunksize>0 in code like

```
      if (restart_format=='hdf5' .and. size(dims)>1) then
         if (dims(1)==dimid_ni .and. dims(2)==dimid_nj) then
            chunks(1)=restart_chunksize(1)
            chunks(2)=restart_chunksize(2)
            do i = 3, size(dims)
               chunks(i) = 0
            enddo
            status = nf90_def_var_chunking(ncid, varid, NF90_CHUNKED, chunksizes=chunks)
            call ice_check_nc(status, subname//' ERROR: chunking var '//trim(vname), file=__FILE__, line=__LINE__)
         endif
      endif
```

Should that first line be

`      if (restart_format=='hdf5' .and. size(dims)>1 .and. maxval(restart_chunksize)>0) then
`

What happens when the chunks are all zeros and you call nf90_def_var_chunking?
Is there a cost to calling nf90_def_var_chunking when we don't need to?

anton-seaice:
> What happens when the chunks are all zeros and you call nf90_def_var_chunking? Is there a cost to calling nf90_def_var_chunking when we don't need to?

If chunks are all zeros then it will set the chunksize to the maximum (i.e. the size of the array) unless the variable is more than 4MB, at which point some '[Default](https://docs.unidata.ucar.edu/nug/current/netcdf_perf_chunking.html#default_chunking_4_1)' chunksize is used. 

If we made the change suggested: 

- For vars without unlimited variables (i.e coordinates), the storage type will default to _NC_CONTIGUOUS_ (which I guess is the practically the same as the current behaviour of setting storage type to _NC_CHUNKED_ with the chunksize equal to the array size).          
- For data vars with unlimited variables (i.e. data vars with a time dimension), the storage type will default to _NC_CHUNKED_ anyway. 

There might be some edge cases, where 'hdf5' and contiguous storage are needed, or where the default chunking scheme doesn't work appropriately and its not desired to set the chunksizes in the configuration, but these seem obscure enough that we can look at them if needed? Otherwise we would need to make the time variable length 1 (rather than unlimited) or add another namelist option for hdf5 and _NC_CONTIGUOUS_ (or use `_chunksize=-99` for contiguous, and `_chunksize=0` for default chunksizes) but it would only be meaningful for variables >4MB.

anton-seaice:
I guess that was a long way of saying, I think what we doing now is OK, but we might get feedback suggesting something different is needed in bespoke configurations. Noting we are not changing any current behaviour, it only affects configurations which move to 'hdf5' output.

apcraig:
> I guess that was a long way of saying, I think what we doing now is OK, but we might get feedback suggesting something different is needed in bespoke configurations. Noting we are not changing any current behaviour, it only affects configurations which move to 'hdf5' output.

Sounds good.  Will leave as is.  Thanks.

doc/source/user_guide/ug_implementation.rst
```diff
@@ -1136,28 +1136,90 @@ relaxation parameter ``arlx1i`` effectively sets the damping timescale in
...
+data via a number of different methods.  The IO implementation is specified
+both at build-time (via selection of specific source code) and run-time (via namelist).
+Three different IO packages are available in CICE under the directory
+**cicecore/cicedyn/infrastructure/io**.  Those are io_binary, io_netcdf, and
+io_pio2, and those support IO thru binary, netCDF, and PIO interfaces respectively.
+The io_pio2 directory supports both PIO1 and PIO2 and can write data thru the
+netCDF or parallel netCDF (pnetCDF) interface.  The netCDF history files are CF-compliant, and
+header information for data contained in the netCDF files is displayed with 
+the command ``ncdump -h filename.nc``.  To select the io source code, set ``ICE_IOTYPE``
+in **cice.settings** to ``binary``, ``netcdf``, ``pio1``, or ``pio2``.
```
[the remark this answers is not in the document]
apcraig: I pointed 4.3.6 to this section now so they are sort of linked together.

doc/source/cice_index.rst
```diff
@@ -323,7 +323,11 @@ section :ref:`tabnamelist`.
    "history_dir", "path to history output files", ""
    "history_file", "history output file prefix", ""
    "history_format", "history file format", ""
+   "history_iotasks", "history output total number of tasks used", ""
    "history_precision", "history output precision: 4 or 8 byte", "4"
+   "history_rearranger", "history output io rearranger method", ""
+   "history_root", "history output io root task id", ""
+   "history_stride", "history output io task stride", ""
```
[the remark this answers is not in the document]
apcraig: Done in the namelist documentation.

apcraig:
I think this is ready for final review and merging.  I am running a full test suite now.

@anton-seaice, let me know if I've missed anything or if you think we still have a few things to do.  Thanks!

@eclare108213 @dabail10 @DeniseWorthen @phil-blain, feel free to have a look, let me know if you have an suggested changes.

This adds support for many new netCDF formats and some new controls for PIO.  It also adds compression for hdf5 netcdf output.  I've tested compression and it's pretty awesome.  Thanks @anton-seaice!

anton-seaice:
All looks good to me Tony :) After any more review comments I will retest a few combinations.

@dabail10 - I snuck some related changes to the nuopc driver in. It gets a little bit messy because the io_type comes from nuopc, but the chunking and compression settings come from the cice namelist. (If its better to look at the nuopc driver as a seperate change let me know).

cicecore/cicedyn/infrastructure/io/io_pio2/ice_history_write.F90
```diff
@@ -232,64 +237,40 @@ subroutine ice_write_hist (ns)
...
       elseif (use_leap_years) then
-         call ice_pio_check(pio_put_att(File,varid,'calendar','Gregorian'), &
-              subname//' ERROR: defining att calendar Gregorian',file=__FILE__,line=__LINE__)
+         cal_att='Gregorian'
       else
          call abort_ice(subname//' ERROR: invalid calendar settings')
       endif
 
+      time_coord = coord_attributes('time', 'time', trim(cal_units))
+      call ice_hist_coord_def(File, time_coord, pio_double, (/timid/), varid) !why is pio_double this not lprecision ?
```
anton-seaice, line 256: @apcraig Do you know why the time coord is always hardcoded to be type double? (i.e. 1 second as a fraction of a day is still much bigger than the minimum a type single would support?)

apcraig: Ahh... Time.....

The problem is not the precision per se, it's having the digits to support accuracy overall.  There are a lot of seconds in 2024 years, for instance, 11 digits worth to be exact.  Single precision just isn't enough.  And it's not just when we count in seconds, even if you count days, you still need the precision to have 1/86400th of a day if we want to resolve time to the second.  We're sort of lucky we only need doubles to count to 1000s of years.  If we started to run millions of years (or needed to resolve milliseconds), we might be in trouble.  There are ways around this, like creating offsets and not counting from zero, and we do that.  But in general, you don't want to end up in a situation where time accuracy is lost, so double is safer.

anton-seaice: Ok - maybe just remove that comment in the code next time you are doing a tidy up then :)

dabail10:
> All looks good to me Tony :) After any more review comments I will retest a few combinations.
> 
> @dabail10 - I snuck some related changes to the nuopc driver in. It gets a little bit messy because the io_type comes from nuopc, but the chunking and compression settings come from the cice namelist. (If its better to look at the nuopc driver as a seperate change let me know).

We would definitely prefer to get all PIO related settings from the driver. Since we are specifying the pe layout for the whole system, it knows how many io tasks and so on to assign each component.  I would override the namelist settings here.

cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
```diff
@@ -645,6 +646,36 @@ subroutine InitializeAdvertise(gcomp, importState, exportState, clock, rc)
        call abort_ice(trim(errmsg))
     endif
 
+    ! Netcdf output created by PIO
```
DeniseWorthen, line 649: Should this section be inside or outside the CESM ifdef?

apcraig:
I have run a full test suite on derecho and results are as expected.  I updated the PR,

Test results look good. https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#1619d6ca92028d1c0341be06081d226882a0739a. There is a problem in the bit-for-bit hdf5 restart files generated from PIO. The model restarts exactly, but the check that compares restart files fails because there is some non-important data in the PIO hdf5 files that is not bit reproducible. Have discussed this with PIO folks and they say it's OK. I don't fully agree, netCDF hdf5 does not have this problem. But it means the PIO hdf5 restart tests will fail the "cmp" validation even though they restart the model exactly.

cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
anton-seaice: Probably outside would be better? I am not sure. We (ACCESS) compile with the CESMCOUPLED flag on currently.

anton-seaice:
> > All looks good to me Tony :) After any more review comments I will retest a few combinations.
> > @dabail10 - I snuck some related changes to the nuopc driver in. It gets a little bit messy because the io_type comes from nuopc, but the chunking and compression settings come from the cice namelist. (If its better to look at the nuopc driver as a seperate change let me know).
> 
> We would definitely prefer to get all PIO related settings from the driver. Since we are specifying the pe layout for the whole system, it knows how many io tasks and so on to assign each component. I would override the namelist settings here.

OK - The IO tasks etc are unchanged for the nuopc driver.

There are new CICE namelist parameters - history_chunksize, history_compression, restart_chunksize and restart_compression added in this PR. We could add corresponding variables through nuopc to override them if that is neater?

anton-seaice:
> I have run a full test suite on derecho and results are as expected. I updated the PR,
> 
> Test results look good. https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#1619d6ca92028d1c0341be06081d226882a0739a. There is a problem in the bit-for-bit hdf5 restart files generated from PIO. The model restarts exactly, but the check that compares restart files fails because there is some non-important data in the PIO hdf5 files that is not bit reproducible. Have discussed this with PIO folks and they say it's OK. I don't fully agree, netCDF hdf5 does not have this problem. But it means the PIO hdf5 restart tests will fail the "cmp" validation even though they restart the model exactly.

Ok - sounds like its not something that is feasible/worthwhile for us to address? Is the failure repeatable or a bit random?

cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
DeniseWorthen: Right, and we (UWM) do not have CESMCOUPLED set so I think we'll need them outside.

apcraig:
> > I have run a full test suite on derecho and results are as expected. I updated the PR,
> > Test results look good. https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#1619d6ca92028d1c0341be06081d226882a0739a. There is a problem in the bit-for-bit hdf5 restart files generated from PIO. The model restarts exactly, but the check that compares restart files fails because there is some non-important data in the PIO hdf5 files that is not bit reproducible. Have discussed this with PIO folks and they say it's OK. I don't fully agree, netCDF hdf5 does not have this problem. But it means the PIO hdf5 restart tests will fail the "cmp" validation even though they restart the model exactly.
> 
> Ok - sounds like its not something that is feasible/worthwhile for us to address? Is the failure repeatable or a bit random?

The failures are repeatable.  The differences in the files are random.  So some "junk" ends up in the file that does not impact results but is random.  You can only see the difference with a cmp or od.  If you do ncdiff or cprnc, the data looks fine.  I'm reluctant to move away from our simple "cmp" just for the pio hdf5 results at this point.  netcdf hdf5 does NOT behave this way, PIO folks are aware of this, and are not concerned.  I'm pretty sure this is a PIO thing which we have little control over.  I could explore further.  I do know that CICE PIO and netcdf handle fill data slightly differently. but I don't think it's related to that.  I think we should just live with this for now.  I don't expect it will cause problems for users, but we should keep our ears open.

cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
anton-seaice: Ok - I will update

DeniseWorthen: @anton-seaice Let me test before you move it outside. I'm just now looking at the code changes. Generally in UWM we are going to be setting these options via namelist.

phil-blain (commented):
I went through the documentation changes and they look good.

I'll take a look at the `io_netcdf` code later.

I'd like to mention that it is really hard and demands a lot of energy to review big PRs like this, especially because we do not put a lot of emphasis on making small, atomic commits. 

I think we could discuss this in the monthly meetings, but I would be really glad if we could make smaller commits, and make sure that each commit message explains why the changes are made, and make sure that each commits builds, etc. This would make it much more enjoyable (at least for me) to review PRs. 

It would also open the possibility of using true merges instead of squash merges, which would also be very useful (in my opinion) because our history would be easier to read and understand. Just my 2 cents, as usual... :)

Here is a blog post discussing this subject that I like: https://github.blog/2022-06-30-write-better-commits-build-better-projects/

cicecore/cicedyn/infrastructure/io/io_netcdf/ice_history_write.F90
```diff
@@ -742,6 +648,12 @@ subroutine ice_write_hist (ns)
          call ice_check_nc(status, subname// ' ERROR: global attribute io_flavor', &
                            file=__FILE__, line=__LINE__)
 
+         if (history_format == 'hdf5') then
+            status = nf90_put_att(ncid,nf90_global,'deflate',history_deflate)
+            call ice_check_nc(status, subname//' ERROR: defining attribte deflate', &
+                              file=__FILE__,line=__LINE__)
+         endif
+
```
anton-seaice, line 656: @apcraig I am not sure about adding this as an attribute. It makes is seem like its a global feature, when its actually at the variable level. You can already see the deflate level through ncdump (and probably other tools) without add it as an attribute.

apcraig: I'm happy to remove this again, it's not a critical thing.  Will do that now.

DeniseWorthen:
@anton-seaice I'm trying to test the new features using the ``hdf5`` option. I'm getting an error, which I believe is the result of the spack-stack not having built PIO w/ netcdf4/hdf5 enabled. The error message I get is

```
74:  (abort_ice)ABORTED:
174:  (abort_ice) called from ice_pio.F90
174:  (abort_ice) line number          238
174:  (abort_ice) error =
174:  (ice_pio_check)NetCDF: Attempt to use feature that was not turned on when netCD
174:  F was built., (ice_pio_init) ERROR: Failed to open file cice_model.res.nc
```

The message part that says ```Attempt to use feature that was not turned on when netCDF was built``` is coming from PIO itself I believe. But we use netcdf4/hdf5 for the ATM model so we should have netCDF built correctly. Does this message really indicate that PIO itself was not built w/ netcdf4/hdf5 support?

apcraig:
> @anton-seaice I'm trying to test the new features using the `hdf5` option. I'm getting an error, which I believe is the result of the spack-stack not having built PIO w/ netcdf4/hdf5 enabled. The error message I get is
> 
> ```
> 74:  (abort_ice)ABORTED:
> 174:  (abort_ice) called from ice_pio.F90
> 174:  (abort_ice) line number          238
> 174:  (abort_ice) error =
> 174:  (ice_pio_check)NetCDF: Attempt to use feature that was not turned on when netCD
> 174:  F was built., (ice_pio_init) ERROR: Failed to open file cice_model.res.nc
> ```
> 
> The message part that says `Attempt to use feature that was not turned on when netCDF was built` is coming from PIO itself I believe. But we use netcdf4/hdf5 for the ATM model so we should have netCDF built correctly. Does this message really indicate that PIO itself was not built w/ netcdf4/hdf5 support?

So this is happening on a read?  And I assume cice_model.res.nc is not hdf5?  We shouldn't have to tell PIO the netcdf format when we're reading, I'll have a look at the code again on that.  But this would just fail later on a PIO hdf5 write, so it doesn't really solve the problem.

My guess is that the atm and PIO are compiled with and writing thru different version of the netcdf lib.  I suspect this is possible, but am not 100% sure.  Some questions/ideas.  Is PIO being built on the fly?  How consistent are the builds for each component?  Can you verify what version of netcdf is compiled with the atm and pio libs?  If you are using a "netcdf" module, you might try loading the equivalent "netcdf-mpi" module.  To get the hdf5 to work on Derecho, I had to do

```
  module unload netcdf
  module load netcdf-mpi/4.9.2
```

You can see that in the PR.  Just some thoughts.  Keep us posted.

DeniseWorthen:
The file it is reading is netcdf-classic. But, like you, I thought the read should be fine. 

We build UWM with a spack-stack on each platform---and the basic build should be the same (but the exact intel or gnu compiler might be different). I'm still learning my way around spack-stack, this is the package dependencies (Jim is the package maintainer).  

https://github.com/JCSDA/spack/blob/03ecae6b77b974c7f364fbbca7ad06e2bc1ba8ab/var/spack/repos/builtin/packages/parallelio/package.py#L33-L41

DeniseWorthen:
Indeed, with

```
    restart_format = 'hdf5'
    history_format = 'hdf5'
```

I can write a history file (ncdump -k = netCDF-4), but not read a netcdf-classic restart file (ncdump -k = classic).

anton-seaice:
> The file it is reading is netcdf-classic. But, like you, I thought the read should be fine.
> 
> We build UWM with a spack-stack on each platform---and the basic build should be the same (but the exact intel or gnu compiler might be different). I'm still learning my way around spack-stack, this is the package dependencies (Jim is the package maintainer).
> 
> https://github.com/JCSDA/spack/blob/03ecae6b77b974c7f364fbbca7ad06e2bc1ba8ab/var/spack/repos/builtin/packages/parallelio/package.py#L33-L41

Hi Denise. I haven't seen this before. I also thought that all the netcdf formats should be backwards compatible for read.

Where is your spack environment file? The "variant" may be changed in the environment file too. Maybe you would need parallel-netcdf turned on for the netcdf-c package? (https://github.com/JCSDA/spack/blob/03ecae6b77b974c7f364fbbca7ad06e2bc1ba8ab/var/spack/repos/builtin/packages/netcdf-c/package.py#L124)

apcraig:
This is all a bit surprising.  How about changing this in cicecore/cicedyn/infrastructure/io/io_pio2/ice_restart.F90 as a sensitivity test?  At about line 81, change

```
      call ice_pio_init(mode='read', filename=trim(filename), File=File, &
           fformat=trim(restart_format), rearr=trim(restart_rearranger), &
           iotasks=restart_iotasks, root=restart_root, stride=restart_stride, &
           debug=first_call)
```
to

```
      call ice_pio_init(mode='read', filename=trim(filename), File=File, &
           rearr=trim(restart_rearranger), &
           iotasks=restart_iotasks, root=restart_root, stride=restart_stride, &
           debug=first_call)
```

remove the "fformat" argument.  This should not be needed for the read anyway in general.  Then try to set restart_format='hdf5' and try reading both the classic and hdf5 version of the grid file.  Does that work?

anton-seaice:
> Oh, good point. What I see from the JCDSA issue for netcdf-c is (this happens to be on Dom's OSX system) is
> 
> ```
> [+]      ^netcdf-c@4.9.2%apple-clang@13.1.6+blosc~byterange+dap~fsync~hdf4~jna+mpi~nczarr_zip+optimize~parallel-netcdf+pic+shared~szip+zstd build_system=autotools patches=0161eb8 arch=darwin-monterey-m1
> ```
> 
> I'm sure the + is "with" but I'm not entirely sure that ~ is "without"?

That all looks normal I think. ~ is indeed "without"

apcraig:
Can someone explain what's going on.  I don't know spack.  As I understand it from @DeniseWorthen, if restart_format is set to hdf5 and the restart file is hdf5, then the code reads and writes hdf5 just fine.  But if the restart file is not hdf5, then it fails.  netcdf should be able to convert formats automatically on read, but it doesn't in the case.  I'm pretty sure this does work fine on derecho, it reads a classic restart and writes hdf5.  So why does this seem to only be a problem with a spack build when even the spack build supports hdf5 read and write fine (just not mixing).

In general, the code supports reading and writing different restart formats (with some limits).  But this should be allowed, right?

anton-seaice:
> Can someone explain what's going on. I don't know spack. As I understand it from @DeniseWorthen, if restart_format is set to hdf5 and the restart file is hdf5, then the code reads and writes hdf5 just fine. But if the restart file is not hdf5, then it fails. netcdf should be able to convert formats automatically on read, but it doesn't in the case. I'm pretty sure this does work fine on derecho, it reads a classic restart and writes hdf5. So why does this seem to only be a problem with a spack build when even the spack build supports hdf5 read and write fine (just not mixing).
> 
> In general, the code supports reading and writing different restart formats (with some limits). But this should be allowed, right?

You are on it Tony :)

The only difference I can think of is possibly the derecho configuration of netcdf-c includes 'parallel_netcdf' support ? Which maybe is needed for parallel I/O of netcdf classic files but not netcdf4 files.

apcraig:
OK.  I'd be curious to see whether my suggestion of changing the ice_pio_init call for the restart read changes anything.  Maybe if we don't force the read to be hdf5 (if we're even doing that), that will read the classic and hdf5 while still writing hdf5 or other formats.  I'm testing that change on derecho and it works fine both ways on derecho.  Just wondering whether it'll work around the spack problems.

anton-seaice:
> OK. I'd be curious to see whether my suggestion of changing the ice_pio_init call for the restart read changes anything. Maybe if we don't force the read to be hdf5 (if we're even doing that), that will read the classic and hdf5 while still writing hdf5 or other formats. I'm testing that change on derecho and it works fine both ways on derecho. Just wondering whether it'll work around the spack problems.

I hardcoded line 236 in ice_pio to use `PIO_IOTYPE_NETCDF` for read operations, and it will then read both netcdf classic and netcdf4 initial conditions files. But this means all reads of netcdf files will be single PE only.

apcraig:
> I hardcoded line 236 in ice_pio to use `PIO_IOTYPE_NETCDF` for read operations, and it will then read both netcdf classic and netcdf4 initial conditions files. But this means all reads of netcdf files will be single PE only.

But that means you only wrote classic files?  Can you try changing just the ice_pio_init call on the restart read as suggested above.  I'm curious if that supports reading different file types and writing hdf5 (in parallel?).  If so, that might be a good solution.  I think if we can only read serially, that's probably an OK trade off for flexibility.  As long as the writes are in parallel using hdf5 or pnetcdf.  Performance is generally more of an issue for output as there is usually more output than input.

anton-seaice:
> > I hardcoded line 236 in ice_pio to use `PIO_IOTYPE_NETCDF` for read operations, and it will then read both netcdf classic and netcdf4 initial conditions files. But this means all reads of netcdf files will be single PE only.
> 
> But that means you only wrote classic files? Can you try changing just the ice_pio_init call on the restart read as suggested above. I'm curious if that supports reading different file types and writing hdf5 (in parallel?). If so, that might be a good solution. I think if we can only read serially, that's probably an OK trade off for flexibility. As long as the writes are in parallel using hdf5 or pnetcdf. Performance is generally more of an issue for output as there is usually more output than input.

Yeah - I only hardcoded the type for read operations. The output was still netcdf4 type.

Most likely parallel reads are not that important ... I don't have a good config to test that on at the moment. The other option would be try the parallel read, and if that fails, try a read for netcdf classic. 

I will try and build the dependencies with pnetcdf enabled, well see!

apcraig:
I think the most important thing is that it works robustly at this point.  I think we should defer read performance, make sure it's not aborting when it probably shouldn't, and make sure the write performance is working as expected with netcdf, hdf5, pnetcdf and in parallel when we expect it should.  It might be even that is too high a bar, keep me posted.

DeniseWorthen:
Catching up here....I've now switched over to Derecho using the spack-stack built for UWM there. I have the same issue: using hdf5 as a restart_format fails when using a classic restart file. Converting it to a netcdf4-classic works for restart_format=hdf5. I'm now trying the code change that Tony suggested above for our build.

In terms of general solutions, I think yes, read performance is not as critical. I think suggestion 3 above seems a good compromise. Are we sure that would work in all cases?

> - Retry opening in ice_pio.F90. If the read fails with the requested iotype, retry with iotype="cdf1".

DeniseWorthen:
I've been wondering--Is there any way to use the netcdf inquire format function (nc_inq_format) to determine what to do for restart reads?

anton-seaice:
Sorry for the Friday afternoon explanation ! ... Here goes. I will revisit Monday if needed :)

> If you open a file hdf5 thru PIO, it expects to read/write in parallel. If the file to be read is NOT an hdf5 file, then it can't read in parallel and an abort is triggered. .

So the interesting thing here is, if the parallel read fails, the pio library should retry with a serial read and netcdf classic (aka PIO_TYPE_NETCDF). [see link](https://github.com/NCAR/ParallelIO/blob/6539ef05ae7584ec570a56fdab9f7dfb336c2b80/src/clib/pio_file.c#L41)

I messed around for long enough to get extra PIO logging turned on (see [cice.runlog.240209-153008.txt](https://github.com/CICE-Consortium/CICE/files/14217425/cice.runlog.240209-153008.txt)), and saw that the iotype when it retries to open the file using a serial open hadn't been updated. e.g.: logs like this:

`1 retry nc_open(/g/data/ik11/inputs/CICE_data/ic/gx3/iced_gx3_v5.nc) : fd = -1, iotype = 4, do_io = 1, ierr = -128`

Where `nc_open` is the serial read that is the retry, but the iotype = 4 (should be PIO_IOTYPE_NETCDF=2)

The parallelio changes this iotype before retrying:

                /* reset file markers for NETCDF on all tasks */
                file->iotype = PIO_IOTYPE_NETCDF;

I suspected that the variable `file` which is shared from fortran to the c binding, wasn't getting updated. So I made a local 'lFile' within ice_pio_init, and it fixed the problem! (See https://github.com/CICE-Consortium/CICE/commit/3ef71dd0c5f9cb7cac9983fd746c085b585bcbc3). I don't really understand what is going on here, but maybe you will Tony?

> @DeniseWorthen, do you think it's worth trying to set up standalone CICE testing using the spack builds? Would that be difficult? Also, maybe I'll try to generate different formats for the restart input files, so we add some test coverage for that.

I have done this on Gadi - it was straightforward enough. Spack produces modulefiles, so as Denise says, you just need to `module use` and `module load` the modules produced via spack, rather than the default ones.

apcraig:
Thanks @anton-seaice for all this work, very good sleuthing.   So how should we handle things?  Do you want to create a PR to the branch with something like https://github.com/CICE-Consortium/CICE/commit/3ef71dd0c5f9cb7cac9983fd746c085b585bcbc3 or should I update the branch based on your changes?  Should we go with my solution which just sets the restart read to cdf1?  I don't understand the fix, is whatever is going on here something we need to worry about elsewhere in PIO?  We haven't run into many c/fortran issues with PIO, so I assume we're OK, but this is weird.

Separately, I'll try to get a spack thing going on Derecho, but maybe as a separate PR at a later time.  It would be good to get this PR sorted out and merged if we can.

anton-seaice:
Ok. There’s someone here I can ask about this sort of stuff, I’ll try that on Monday and see if it makes sense to him. Otherwise maybe to be safe we should just hardcode it to always do the serial read as you suggest.

DeniseWorthen:
I also don't fully understand whether the issue is how we're using the PIO interface vs an issue in PIO itself.

I think it would be fine to commit this with forcing restart read to be serial and creating a followup issue (?). The PR already represents a substantial body of work.

apcraig:
> I think it would be fine to commit this with forcing restart read to be serial and creating a followup issue (?). The PR already represents a substantial body of work.

I think this is a good plan.  I will update, test, and push some changes onto the branch today, nothing particularly groundbreaking.  I will also create an issue.  If we decide to pursue other options next week on this branch, that's great.  But it could also be reconsidered for a PR in the future if we decide to do that.

apcraig:
I just updated the branch with a couple changes, https://github.com/CICE-Consortium/CICE/pull/928/commits/4f09d1f2c0abfa0c1ad499aed8d55695bdce1f06.  Will run a full test suite this weekend and report/fix any problems found (I don't expect any).

I also added #932 and #933 in case we defer those issues.

Unless we have some quick resolution as to why or what is happening wrt #932, I recommend we merge what we have now (after final review) early next week and address #932 and #933 separately.

anton-seaice:
> I just updated the branch with a couple changes, [4f09d1f](https://github.com/CICE-Consortium/CICE/commit/4f09d1f2c0abfa0c1ad499aed8d55695bdce1f06). Will run a full test suite this weekend and report/fix any problems found (I don't expect any).
> 
> I also added #932 and #933 in case we defer those issues.

Thankyou - sounds good.

> Unless we have some quick resolution as to why or what is happening wrt #932, I recommend we merge what we have now (after final review) early next week and address #932 and #933 separately.

Yes ok. I raised https://github.com/NCAR/ParallelIO/issues/1985 and will updated in #932 as needed.

apcraig:
Testing is complete and everything looks good.  Anyone have any fine review comments before merging?  Could a couple folks approve the PR?

cicecore/cicedyn/infrastructure/ice_read_write.F90
```diff
@@ -1241,7 +1241,7 @@ subroutine ice_read_nc_xy(fid,  nrec,  varname, work,  diag, &
          status = nf90_get_att(fid, varid, "_FillValue", missingvalue)
 !          call ice_check_nc(status, subname//' ERROR: Missing _FillValue', &
 !                            file=__FILE__, line=__LINE__)
-         write(nu_diag,*) subname,' missingvalue= ',missingvalue
+!         write(nu_diag,*) subname,' missingvalue= ',missingvalue
```
phil-blain, line 1244: I did not read the full discussion, but is there any reason why we are commenting these lines instead of removing them ?

phil-blain (commented):
I think the refactoring done in `io_netcdf/ice_history_write` does makes the code cleaner and with less repetition, so that's nice, thanks.

I went through the rest of the PR it looks good to me.

cicecore/cicedyn/infrastructure/ice_read_write.F90
apcraig: There a bit of useful debugging code that is part of the optional file read diagnostics that is turned off by default.  It's sort of helpful to have this code around in case it's needed.

dabail10:
This is pretty extensive and I am not sure when I can properly review this. I am at a meeting in Santa Fe this week. Doing some scouting for the CICE Consortium workshop in May. :)

DeniseWorthen:
I've been testing this, and I'm wondering if I understand this correctly. I see in my ice_diag.d file

```
 (ice_pio_init) nprocs     =          720
 (ice_pio_init) pio_iotype =            1
 (ice_pio_init) iotasks    =          179
 (ice_pio_init) baseroot   =            1
 (ice_pio_init) stride     =            4
 (ice_pio_init) nmode      =           32
 (ice_pio_init) create file ./RESTART/iced.2013-04-01-10800.nc
```

Is the number of iotasks actually 179? Or does "1" get added downstream in the PIO interface to make it 180?

apcraig:
> I've been testing this, and I'm wondering if I understand this correctly. I see in my ice_diag.d file
> 
> ```
>  (ice_pio_init) nprocs     =          720
>  (ice_pio_init) pio_iotype =            1
>  (ice_pio_init) iotasks    =          179
>  (ice_pio_init) baseroot   =            1
>  (ice_pio_init) stride     =            4
>  (ice_pio_init) nmode      =           32
>  (ice_pio_init) create file ./RESTART/iced.2013-04-01-10800.nc
> ```
> 
> Is the number of iotasks actually 179? Or does "1" get added downstream in the PIO interface to make it 180? That is, iotasks is set here as if the counting starts at 0, so 0:179.

iotasks is the number of tasks.  So it is 179.  What did you set in namelist?  If you set iotasks to 180, I think it should use 180.  But there also  seems to be a problem in PIO that causes an abort with certain iotask counts in some cases when the iotasks is the "maximum", so I have coded around it.  There is a whole section in icecore/cicedyn/infrastructure/io/io_pio2/ice_pio.F90 that sets the pio pe setup, lines 141-168.  You can see at line 163 a comment about this.  I end up reducing the iotasks to avoid this problem.  The right equation is line 164 and that could be applied at line 144 as well.  Again, had some aborts when I used line 164 even though it should work.

If you formally set iotasks to 180, does it work OK.  Would it be better to use the right equation, giving the full iotasks and then sometimes finding the model aborts?

DeniseWorthen:
@apcraig I am using all default settings so far. In this case, I am specifying (nprocs=720):

```
    restart_format      = 'pnetcdf5'
    restart_iotasks     = -99
    restart_rearranger  = 'default'
    restart_root        = -99
    restart_stride      = -99
    restart_chunksize   = 0,0
    restart_deflate     = 0
```

apcraig:
@anton-seaice @DeniseWorthen it would be great to get a formal review from you both if possible.  Please let me know if you think anything else needs to change.  @eclare108213 @dabail10 I think I'd like one of you to review if possible as well.  Would be good to get this merged.  I'm happy to make additional changes as requested.

apcraig:
@DeniseWorthen I think if you set iotasks to 180 manually, you should get 180.  Let me know if that's not the case.

DeniseWorthen:
@apcraig I tested this PR fairly extensively last week, with an eye primarily towards the impact on performance of the various options. I ran w/ the PIO serial, pnetcdf, hdf5 and with using the default settings vs manually specifying. Everything worked as expected. As a point of interest, I found pnetcdf w/ subset rearranger gave by far the best performance. And, secondly that there seemed to be a small but consistent performance hit (~4%) when using 179 io tasks (nprocs=720, stride=4 which the default calculations provide) vs manually setting iotasks = 180. I didn't run into any unexpected issues, so I'm happy to approve. Thanks to both you and @anton-seaice for this work.

DeniseWorthen (approved):

anton-seaice:
Thanks Tony

I built CICE standalone in the conda environment and tried various combinations of netcdf/pio, _format, _deflate and _chunksizes and it all looked ok. I built cice coupled with nuopc/cmeps and tested netcdf4 with a smaller subset of _chunksize/_deflate etc and that looked ok too.

1. Using io_netcdf and _type='cdf5' I get this error when viewing the "virtual attributes" in the output files using ncdump:

```
Attempting netcdf-4 operation on netcdf-3 file
Location: file ?; fcn ? line 1116
```

I think that might be an issue with ncdump rather than our output? But I didn't investigate in detail. The data appears to still be there

2. Do we need both 'set_env.iopio2' and 'set_env.iopio2p' as options? They look the same. Similar for iopio1 and iopio1p

anton-seaice (approved):

apcraig:
I am going to retest the iotasks with a "180" implementation and see if I can understand better when and why it fails on Derecho.

I will remove the set_nml.iopio1p and iopio2p, thanks for reminding me!  Those are no longer used.

Thanks @DeniseWorthen and @anton-seaice for all your development and testing efforts!

dabail10:
Just checked this out an trying to do the testsuite in CESM. More later.

apcraig:
Did more testing with respect to the PIO error.  PIO error reported here, https://github.com/NCAR/ParallelIO/issues/1986.

Am updating the iotask calculation now to further limit the iotask reduction based on the extra testing.

dabail10:
This compiles and runs fine in my test suite. Thanks for updating the nuopc/cmeps driver. Let me be clear on the logic.

1. When pio_typename from the driver is netcdf and history_format / restart_format are anything other than cdf2, these get reset to netcdf. That is to say 'cdf2' == 'netcdf'?

2. If pio_typename is 'netcdf4p' or 'netcdf4c', then these are set to hdf5.

3. If pio_typename is 'pnetcdf' then these are left as cdf2? I'm less certain on this one.

Dave

apcraig:
I just updated the pio default iotask computation to be a little less conservative.  @DeniseWorthen, you should now get 180 iotasks with stride=4 and nprocs=720 with root=[0,1,2,3].  The iotasks are still reduced to 179 with stride=4 if nprocs=719 with root=[0,1,2] even though 180 iotasks should be OK in that case.  I ran a number of tests which suggest the current implementation is robust with the current bug in PIO2, https://github.com/NCAR/ParallelIO/issues/1986.

I also removed set_env iopio1p and iopio2p, no longer used.

I am running additional tests now just to make sure everything is working.  Will also update the comments in the PR as needed.

apcraig:
I will merge this by the end of the week unless there are any additional comments.  Happy to delay (and/or update other things) if anyone thinks we should.  Thanks!

anton-seaice:
> This compiles and runs fine in my test suite. Thanks for updating the nuopc/cmeps driver. Let me be clear on the logic.
> 
> 1. When pio_typename from the driver is netcdf and history_format / restart_format are anything other than cdf2, these get reset to netcdf. That is to say 'cdf2' == 'netcdf'?
> 
> 2. If pio_typename is 'netcdf4p' or 'netcdf4c', then these are set to hdf5.
> 
> 3. If pio_typename is 'pnetcdf' then these are left as cdf2? I'm less certain on this one.
> 
> 
> Dave

pio_typename is always the type used for IO (if CESMCOUPLED is set). This is not changed.

If pio_typename is netcdf or pnetcdf, current behaviour is unchanged (except the driver attempts to give a warning if the _format is set in _ice_in_)

In _ice_history_write.F90_ and _ice_restart.F90_ chunking and compression are set for variables if the _format is 'hdf5' (for all drivers). So in the nuopc/cmeps driver, if pio_typename is a netcdf4 type, then _format is overridden to 'hdf5'. This is only done so that the chunking and compression can be set. It seemed an ok compromise ... otherwise we would have to add a few `#ifdef CESMCOUPLED` statements to check for `pio_typename=='netcdf4p'` instead of `_format=='hdf5' `. Its definitely messy, because also the type comes from pio_typename but the chunking and compression come from _ice_in_.

dabail10:
This is very helpful. We don't use it currently, but there is another option in PIO for the some of the finer grain options. We might need to support it at some point. 

> > This compiles and runs fine in my test suite. Thanks for updating the nuopc/cmeps driver. Let me be clear on the logic.
> > 
> > 1. When pio_typename from the driver is netcdf and history_format / restart_format are anything other than cdf2, these get reset to netcdf. That is to say 'cdf2' == 'netcdf'?
> > 2. If pio_typename is 'netcdf4p' or 'netcdf4c', then these are set to hdf5.
> > 3. If pio_typename is 'pnetcdf' then these are left as cdf2? I'm less certain on this one.
> > 
> > Dave
> 
> pio_typename is always the type used for IO (if CESMCOUPLED is set). This is not changed.
> 
> If pio_typename is netcdf or pnetcdf, current behaviour is unchanged (except the driver attempts to give a warning if the _format is set in _ice_in_)
> 
> In _ice_history_write.F90_ and _ice_restart.F90_ chunking and compression are set for variables if the _format is 'hdf5' (for all drivers). So in the nuopc/cmeps driver, if pio_typename is a netcdf4 type, then _format is overridden to 'hdf5'. This is only done so that the chunking and compression can be set. It seemed an ok compromise ... otherwise we would have to add a few `#ifdef CESMCOUPLED` statements to check for `pio_typename=='netcdf4p'` instead of `_format=='hdf5' `. Its definitely messy, because also the type comes from pio_typename but the chunking and compression come from _ice_in_.

apcraig:
Full test suite run on Derecho with latest code is as expected, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#25467263df7c5b61e94d55b7c73208fa08a67127

dabail10 (approved):
I have tested this within CESM and it does not change anything with the way we do things.

doc/source/user_guide/ug_implementation.rst
```diff
@@ -1136,28 +1136,104 @@ relaxation parameter ``arlx1i`` effectively sets the damping timescale in
...
+and PIO (https://github.com/NCAR/ParallelIO) interfaces respectively.
+The io_pio2 directory supports both PIO1 and PIO2 and can write data thru the
+netCDF or parallel netCDF (pnetCDF) interface.  The netCDF history files are CF-compliant, and
+header information for data contained in the netCDF files is displayed with 
+the command ``ncdump -h filename.nc``.  To select the io source code, set ``ICE_IOTYPE``
+in **cice.settings** to ``binary``, ``netcdf``, ``pio1``, or ``pio2``.
+
+At run-time, more detailed IO settings are available.  ``restart_format`` and
+``history_format`` namelist options specify the method and format further.  Valid options
+are listed in :ref:`formats`.  These format specify the format of new files created
```
eclare108213, line 1166: These options specify...

doc/source/user_guide/ug_implementation.rst
```diff
@@ -1136,28 +1136,104 @@ relaxation parameter ``arlx1i`` effectively sets the damping timescale in
...
+provides information about the specific netCDF file format.
+In general, the detailed format is not enforced for input files, so any netCDF format 
+can be read in CICE regardless of CICE namelist settings.
+
+.. _formats:
+
+.. table:: CICE IO formats
+
+   +--------------+----------------------+---------+---------------------+
+   | **Namelist** | **Format**           | **By**  |  **Valid With**     |
```
eclare108213, line 1180: "By" is not immediately clear.  Maybe "Source library" or something like that?

doc/source/user_guide/ug_implementation.rst
```diff
@@ -1404,18 +1475,16 @@ The timers use *MPI\_WTIME* for parallel runs and the F90 intrinsic
...
+CICE reads and writes restart data in binary unformatted or netCDF formats via
+separate implementations of binary, netCDF, and PIO interfaces as described
+above.  In addition, ``restart_format`` as well as other restart namelist
+options control the specific file format as well as features related to 
+IO performance, see :ref:`iooverview`.
 
 The restart files created by CICE contain all of the variables needed
 for a full, exact restart. The filename begins with the character string
-‘iced.’, and the restart dump frequency is given by the namelist
+define by the ``restart_file`` namelist input, and the restart dump frequency
```
eclare108213, line 1486: defined

eclare108213 (approved):
Just a few minor comments on the documentation. Approving based on extensive testing by the development team.  Thanks all!

apcraig:
Thanks @eclare108213, I updated the documentation as suggested.  I changed "By" to "Written With" which I think is clearer.  Let me know if you see anything else.

eclare108213:
Looks good, thank you.

dabail10:
Also, any chance we can remove support for PIO1?

apcraig:
> Also, any chance we can remove support for PIO1?

I'm open to some discussion.  But why would we want to do that?  The coding and maintenance effort to support PIO1 is very small.  We still use PIO1 in RASM.  There are some nice things about it, it's all in Fortran :)

dabail10:
I know it is minimal, but Jim Edwards is not willing to support PIO1 anymore. Any chance you can move RASM to PIO2? Anyhow, something to be thinking about. Also, I the flag is CESM1_PIO. I think this should just be PIO1 now.

apcraig:
I understand PIO1 is kind of a dead end.  But I'm trying to avoid having to update RASM to PIO2 because the PIO build would have to change, and I prefer to defer that effort indefinitely if I can.  CESM1_PIO is a different thing than PIO1.  We do not need CESM1_PIO1 to run with PIO1.  We already have a USE_PIO1 CPP that is used with PIO1.  CESM1_PIO is needed to use an older version of PIO1 in RASM that is not compatible with the latest version of PIO1.  This is what happens when libraries do not maintain backwards compatibility.

Issue #842, opened 2023-07-11
msec not allowed to be set to non-zero values for initial runs


zarzycki:
Code reference: https://github.com/CICE-Consortium/CICE/blob/766ff8d9606ae08bdd34ac2b36b6b068464c7e71/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90#L783C1-L786C1

In this block of code, is there a reason why msec is not initialized (well, set to something other than the default of 0)?

In CESM, I have found ESMF chokes with async clocks if START_TOD is not = 0 (e.g., a 12Z initialization of CAM-MPAS). I have gotten around this by modifying ice_comp_nuopc.F90:

```
          call abort_ice(subname//' :: ERROR idate lt zero')
       endif
       myear = (idate/10000)                     ! integer year of basedate
       mmonth= (idate-myear*10000)/100           ! integer month of basedate
       mday  =  idate-myear*10000-mmonth*100     ! day of month of basedate
!++CMZ
       msec = start_tod
!--CMZ
```

which alleviates the issue and synchronizes the clocks and permits the run to complete, but I'm not sure if there's a particular design choice to omit setting msec I am ignoring.

dabail10:
Thanks Colin. I guess we have not needed a time of day start time. This is a good suggestion and I will look to implement it.

dabail10:
Adding @DeniseWorthen to this one. How is UFS handling the curr_tod and start_tod?

anton-seaice:
The suggested change looks correct, ACCESS don't use part days for anything much at this point, so we haven't had this issue.

dabail10:
Finally circling back to this. I just had another request from the CAM developers. I think we want to listen to the driver on this and not the namelist. I will implement and test Colin's suggested fix.

dabail10:
@zarzycki does this change also work with RUN_REFTOD?

zarzycki:
I cannot say I recall trying a mid-day hybrid/branch run, which I think is needed to exercise RUN_REFTOD.

I would assume it should behave similarly since I think CAM is initialized in the same way, but I cannot 100% confirm it.

apcraig:
Fixed in #991

PR #936, opened 2024-02-20
Set 'idate0' and 'use_leap_years' in nuopc cap

anton-seaice:
## PR checklist
   Set 'idate0' and 'use_leap_years' in nuopc cap. The nuopc cap sets the calendar for CICE (rather than the normal namelist options), but these variables were missing. This means that the attributes of the time variable history output are incorrect in two ways:

- _time:units_ start from year 0000, which is not allowed if Gregorian calendar is used strictly (it starts in 1582 or so)
- _time:calendar_ is always 'noleap', even when it should be 'Gregorian'

- [x] Developer(s): 
    @anton-seaice 
@DeniseWorthen @dabail10 
I did adhoc testing, with the defaults in `ice_in`, a gregorian calendar in nuopc/cmeps and current CICE, on initial run, the time var looks like this:

```
        double time(time) ;
                time:long_name = "time" ;
                time:units = "days since 0000-01-01 00:00:00" ;
                time:calendar = "noleap" ;
                time:bounds = "time_bounds" ;

                time = 715146 ;
```

After the change:
runtype = 'initial'

```
ncdump -v time archive/output000/GMOM_JRA.cice.h.1958-01-01.nc 

        double time(time) ;
                time:long_name = "time" ;
                time:units = "days since 1958-01-01 00:00:00" ;
                time:calendar = "Gregorian" ;
                time:bounds = "time_bounds" ;

time = 1 ;
```
runtype = 'continue'
```
ncdump -v time archive/output001/GMOM_JRA.cice.h.1958-01-02.nc 

        double time(time) ;
                time:long_name = "time" ;
                time:units = "days since 1958-01-01 00:00:00" ;
                time:calendar = "Gregorian" ;
                time:bounds = "time_bounds" ;

                data:

 time = 2 ;
```
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No (NUOPC driver changes only)

Set run start date (idate0) in nuopc driver, so that the history "time:units" attribute in netcdf output is consistent with the date set through nuopc.

Set use_leap_years in nuopc cap, so that netcdf output "time:calendar" is set correctly in history output (i.e. is correctly set as noleap / gregorian) for calendars set through the driver.

### DIFF
--- a/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
+++ b/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
@@ -23,10 +23,10 @@ module ice_comp_nuopc
   use ice_domain_size    , only : nx_global, ny_global
   use ice_grid           , only : grid_format, init_grid2
   use ice_communicate    , only : init_communicate, my_task, master_task, mpi_comm_ice
-  use ice_calendar       , only : force_restart_now, write_ic, init_calendar
-  use ice_calendar       , only : idate, mday, mmonth, myear, year_init
+  use ice_calendar       , only : force_restart_now, write_ic
+  use ice_calendar       , only : idate, idate0,  mday, mmonth, myear, year_init, month_init, day_init
   use ice_calendar       , only : msec, dt, calendar, calendar_type, nextsw_cday, istep
-  use ice_calendar       , only : ice_calendar_noleap, ice_calendar_gregorian
+  use ice_calendar       , only : ice_calendar_noleap, ice_calendar_gregorian, use_leap_years
   use ice_kinds_mod      , only : dbl_kind, int_kind, char_len, char_len_long
   use ice_fileunits      , only : nu_diag, nu_diag_set, inst_index, inst_name
   use ice_fileunits      , only : inst_suffix, release_all_fileunits, flush_fileunit
@@ -675,6 +675,10 @@ contains
     else
       if(mastertask) write(nu_diag,*) trim(subname)//'WARNING: pio_typename from driver needs to be set for netcdf output to work'
     end if
+    ! Set use_leap_years as some CICE calls use this instead of the calendar type
+    call ESMF_TimeGet( currTime, calkindflag=esmf_caltype, rc=rc )
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+    if (esmf_caltype == ESMF_CALKIND_GREGORIAN) use_leap_years = .true.
 
 #else
 
@@ -805,7 +809,7 @@ contains
        if (ref_ymd /= start_ymd .or. ref_tod /= start_tod) then
           if (my_task == master_task) then
              write(nu_diag,*) trim(subname),': ref_ymd ',ref_ymd, ' must equal start_ymd ',start_ymd
-             write(nu_diag,*) trim(subname),': ref_ymd ',ref_tod, ' must equal start_ymd ',start_tod
+             write(nu_diag,*) trim(subname),': ref_ymd ',ref_tod, ' must equal start_tod ',start_tod
           end if
        end if
 
@@ -836,6 +840,12 @@ contains
        endif
 
     end if
+
+    !  - start time from ESMF clock. Used to set history time units
+    idate0    = start_ymd
+    year_init = (idate0/10000)
+    month_init= (idate0-year_init*10000)/100           ! integer month of basedate
+    day_init  = idate0-year_init*10000-month_init*100 
 
     call calendar()     ! update calendar info
 
--- a/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
+++ b/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
@@ -675,10 +675,8 @@ contains
     else
       if(mastertask) write(nu_diag,*) trim(subname)//'WARNING: pio_typename from driver needs to be set for netcdf output to work'
     end if
-    ! Set use_leap_years as some CICE calls use this instead of the calendar type
-    call ESMF_TimeGet( currTime, calkindflag=esmf_caltype, rc=rc )
-    if (ChkErr(rc,__LINE__,u_FILE_u)) return
-    if (esmf_caltype == ESMF_CALKIND_GREGORIAN) use_leap_years = .true.
+
+    
 
 #else
 
@@ -793,7 +791,7 @@ contains
     call cice_init2()
     call t_stopf ('cice_init2')
     !---------------------------------------------------------------------------
-    ! use EClock to reset calendar information on initial start
+    ! use EClock to reset calendar information
     !---------------------------------------------------------------------------
 
     ! - on initial run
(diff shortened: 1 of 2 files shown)


cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90
```diff
@@ -774,7 +779,7 @@ subroutine InitializeRealize(gcomp, importState, exportState, clock, rc)
        if (ref_ymd /= start_ymd .or. ref_tod /= start_tod) then
           if (my_task == master_task) then
              write(nu_diag,*) trim(subname),': ref_ymd ',ref_ymd, ' must equal start_ymd ',start_ymd
-             write(nu_diag,*) trim(subname),': ref_ymd ',ref_tod, ' must equal start_ymd ',start_tod
+             write(nu_diag,*) trim(subname),': ref_ymd ',ref_tod, ' must equal start_tod ',start_tod
```
DeniseWorthen, line 782:
```suggestion
             write(nu_diag,*) trim(subname),': ref_tod',ref_tod, ' must equal start_tod ',start_tod
```

DeniseWorthen:
Do I understand correctly?  In the CESM case, the ``_init`` values for year,month,day and sec were either not being provided via the namelist or the default values were present in ``ice_in``. Either case resulting in erroneous time_origin and time values in the history files. 

In our case, we do set the appropriate ``_init`` values in the namelist, so this change has no impact.

dabail10:
Ah, I see part of the problem. In ice_history_write.F90 we do the following. So, we do need to see use_leap_years = .true. This seems silly to me as we already have the variable calendar_type which we could use here.

```
        if (days_per_year == 360) then
           status = pio_put_att(File,varid,'calendar','360_day')
        elseif (days_per_year == 365 .and. .not.use_leap_years ) then
           status = pio_put_att(File,varid,'calendar','noleap')
        elseif (use_leap_years) then
           status = pio_put_att(File,varid,'calendar','Gregorian')
        else
           call abort_ice(subname//'ERROR: invalid calendar settings')
        endif

```

dabail10:
I guess I am also confused when init_calendar is called. I guess myear, mmonth, mday are only set when runtype == 'initial'. I believe we intended it this way. However, it looks like init_calendar is called in cice_init2 (InitializeRealize). Then for an initial run only, it checks the year, month, day against the values from the coupler. However, it calls calendar and not init_calendar here, so it does not look like idate0 is reset after checking with the coupler. So, instead of adding this code to ice_comp_nuopc.F90, I think there should be a call to init_calendar in this code block. We should also add something in init_calendar that sets use_leap_years to true when calendar_type = 'GREGORIAN'.

anton-seaice:
> Do I understand correctly? In the CESM case, the `_init` values for year,month,day and sec were either not being provided via the namelist or the default values were present in `ice_in`. Either case resulting in erroneous time_origin and time values in the history files.

This is what is happening for ACCESS - getting the right time_origin time_units relied on configuring the same information in two places (cmeps and cice namelist).

> In our case, we do set the appropriate `_init` values in the namelist, so this change has no impact.

Ok good, do we need these lines within a CESM_COUPLED then?
```
    !  - start time from ESMF clock. Used to set history time units
    idate0    = start_ymd
    year_init = (idate0/10000)
    month_init= (idate0-year_init*10000)/100           ! integer month of basedate
    day_init  = idate0-year_init*10000-month_init*100 
```

This change would overwrite the values set in the namelist with the values from ESMF (UFS probbably has then set the same already so it would have no impact?).

anton-seaice:
> I guess I don't understand these changes. We have code in init_calendar that sets idate0 from myear, mmonth, and mday. These are set in ice_comp_nuopc.F90. Similarly, we set the calendar_type in ice_comp_nuopc.F90, I guess technically we should set use_leap_years as well, but this is not used here.
> 
> Dave
> 
> ```
>       idate0 = (myear)*10000 + mmonth*100 + mday ! date (yyyymmdd)
>       stop_now = 0      ! end program execution if stop_now=1
>       dt_dyn = dt/real(ndtd,kind=dbl_kind) ! dynamics et al timestep
>       force_restart_now = .false.
> 
>       ! initialize nstreams to zero (will be initialized from namelist in 'init_hist')
>       ! this avoids using it uninitialzed in 'calendar' below
>       nstreams = 0
> 
> #ifdef CESMCOUPLED
>       ! calendar_type set by coupling 
>       ...
> ```

In this bit of code (from init_calendar) myear, mmonth and mday are set from year_init, month_init, day_init etc which are set from the cice namelist. init_calendar is called in cice_init2, which is before myear/mmonth/mday are updated in InitializeRealize.

> I guess I am also confused when init_calendar is called. I guess myear, mmonth, mday are only set when runtype == 'initial'. I believe we intended it this way. However, it looks like init_calendar is called in cice_init2 (InitializeRealize). Then for an initial run only, it checks the year, month, day against the values from the coupler. However, it calls calendar and not init_calendar here, so it does not look like idate0 is reset after checking with the coupler. So, instead of adding this code to ice_comp_nuopc.F90, I think there should be a call to init_calendar in this code block. 

I don't think this fixes the problem, _init_calendar_ sets myear/mmonth/mday and idate0 from year_init/month_init/day_init and would therefore overwrite the coupler date with the defaults from the cice namelist. (Unless we set year_init/month_init/day_init in the driver before init_calendar is called.)

> We should also add something in init_calendar that sets use_leap_years to true when calendar_type = 'GREGORIAN'.

We can do this, but it feels neater to keep cmeps things in the cmeps driver?

dabail10:
Perhaps. However, there is already an #ifdef CESMCOUPLED in ice_calendar.F90.

anton-seaice:
Ok - i moved the setting of `use_leap_years` to ice_calendar. Let me know if you want to address setting idate0 differently.

anton-seaice:
Also - we could probably address #842 at the same time as well if that is useful :)

dabail10:
> Ok - i moved the setting of `use_leap_years` to ice_calendar. Let me know if you want to address setting idate0 differently.

Actually, we should look at the sequence here. Is calendar_type set before calling init_calendar? If not, we should do it all in ice_comp_nuopc.F90. Thanks for finding this.

anton-seaice:
> > Ok - i moved the setting of `use_leap_years` to ice_calendar. Let me know if you want to address setting idate0 differently.
> 
> Actually, we should look at the sequence here. Is calendar_type set before calling init_calendar? If not, we should do it all in ice_comp_nuopc.F90. Thanks for finding this.

Yeah the calendar is set before the namelist is read and before any init calls:

https://github.com/CICE-Consortium/CICE/blob/aca835755aa82ead50040ea7e43ec63619667054/cicecore/drivers/nuopc/cmeps/ice_comp_nuopc.F90#L488

We can't set use_leap_years at the same line in code because it would be overwritten by ice_init.

anton-seaice:
@dabail10 What should we do with this?

dabail10:
I think we set everything here in ice_comp_nuopc.F90 after the call to cice_init2. Basically as you had it.

anton-seaice:
> I think we set everything here in ice_comp_nuopc.F90 after the call to cice_init2. Basically as you had it.

Ok done. Can you review please? :)

PR #940, opened 2024-03-02
Add a new diagnostic variable dvsdtd.

dabail10:

## PR checklist
This adds a new diagnostic variable dvsdtd so that the SIMIP variable sndmassdyn can be used.
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#1db0d9c13c0885c2da5c498a3fee0daa60788434
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
We finally added the SIMIP variable sndmassdyn. This required the computation of dvsdtd. Also, there were some unit issues with some of the sndmass variables.


cicecore/cicedyn/general/ice_flux.F90
```diff
@@ -419,6 +420,7 @@ subroutine alloc_flux
          strintyU   (nx_block,ny_block,max_blocks), & ! divergence of internal ice stress, y (N/m^2)
          daidtd     (nx_block,ny_block,max_blocks), & ! ice area tendency due to transport   (1/s)
          dvidtd     (nx_block,ny_block,max_blocks), & ! ice volume tendency due to transport (m/s)
+         dvsdtd     (nx_block,ny_block,max_blocks), & ! snow volume tendency due to transport (m/s)
```
eclare108213, line 423: When I first looked at the PR, I thought this was salt, not snow.  Do we need to distinguish the names of those now, in case we want to add a salt tendency term?

eclare108213 (commented):
Looks good.  Is there not also a dvsdtt?
These changes will need to be made to all of the drivers, eventually.

apcraig (commented):
In update state, I suggest
- Add "if present" to dagedt calculation
- With "if present", I think this syntax still sometimes creates run time problems with debug flags (it shouldn't)

```
   if (present(xxx)) xxx = yyy
```
   would recommend


```
   if (present(xxx)) then
     xxx = yyy
   endif
```
- Can we update the indentation in update_state?  Seems it may be overdue.

cicecore/cicedyn/general/ice_flux.F90
dabail10: I guess it would not be hard to add a dvsdtt term. I was doing it specifically since there is the need for an SIMIP variable sndmassdyn. There is not a similar term for the thermodynamics. The snow melt and snow ice formation are called out in separate terms. I was using dvsdtd to be consistent with dvidtd.

apcraig:
> Looks good. Is there not also a dvsdtt? These changes will need to be made to all of the drivers, eventually.

Since update_state is not backwards compatible, could you update all the call to update_state in all the other drivers.  I would NOT add dvsdtd to those calls, but just update the arguments to use keywords= syntax.  That should be adequate.  Thanks.

dabail10:
Ah. Good point. I will do that. I will also work on dvsdtt at the same time.

dabail10:
I have added dvsdtt and fixed all of the calls to update_state in the drivers. The updated test results for the standalone driver are here:

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#e68dd144aa1577111f5a87f93378afd15847b62e

apcraig:
I ran a test suite on Derecho and this all looks good.  @dabail10, anything left to do?  Any other comments?  If not, I'll merge in the next day or two.

apcraig (approved):

dabail10:
I believe this is everything here.

PR #943, opened 2024-03-25
Add checks that individual history streams have unique filenames

### LINKED ISSUE #915: hist_avg on multiple streams writes the same filenames when .false.
This was introduced when the hist_avg variable was modified to set the behaviour on each stream.

apcraig:
## PR checklist
    Add checks that individual history streams have unique filenames
- [X] Developer(s): 
    apcraig
    Ran full test suite on derecho with intel, gnu, cray.  Tests pass as expected.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes, modifies histinst tests
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Add checks that individual history streams have unique filenames.  If they
do not, then streams will overwrite each other.  With this change, abort in that case.  Added in ice_history_shared.F90, subroutine construct_filename. The implementation tracks the latest filenames for each stream and checks versus those names.  Because the file naming convention relies heavily on the current model date/time, this should be adequate (versus keeping track of all history filenames ever used).

Updated the cstream string in subroutine construct_filename.  It was hardwired to len=1 which probably was an error.  Made it len=char_len to support longer hist_suffix character strings in filenames.

Updated the ncfile variable implementation in ice_write_hist in io_binary, io_netcdf, and io_pio2.  It was defined as an array of length max_nstrm, and was changed to a non-array character string.  The array implementation served no purpose.

Modified the set_nml.histinst to add hist_suffix values for each stream.  The latest code modifications cause The current test suite to fail with "histall,histinst" because it creates multiple streams with the same filename.  Setting hist_suffix for histinst fixes this (and tests hist_suffix).

Clean up abort calls in ice_history_shared.F90, add space before "ERROR:".

Update the documentation describing history streams.  Some "\_" formatting was changed to simply "_" where the backslash wasn't needed in ug_implementation.rst.

Several namelist settings were tested to make sure the model would abort with identical filenames including the case where a time averaged file with output at each timestep conflicts with an instantaneous history stream of lower output frequency.

Closes #915


dabail10 (approved):
Looks good.

Issue #945, opened 2024-04-18
Can we remove `nprocs` from `ice_in` ?


anton-seaice:
In this line, all that happens is we check `nprocs` from the namelist matches the results from `get_num_procs()`:

https://github.com/CICE-Consortium/CICE/blob/29c7bcf839bc3ce48e4d6128d6f29ba73839222e/cicecore/cicedyn/infrastructure/ice_domain.F90#L252

Can we just assign ` nprocs = get_num_procs() ` ?

We would have to move it a little bit earlier in the code, before: 
https://github.com/CICE-Consortium/CICE/blob/29c7bcf839bc3ce48e4d6128d6f29ba73839222e/cicecore/cicedyn/infrastructure/ice_domain.F90#L207

This would make model configuration a little easier (i.e. one less configuration to have to match up in multiple places).

apcraig:
I've considered this a number of times, but have put off a change.  I guess there are a few issues.  First, is there ever a case where nprocs (MPI_COMM_SIZE) is not the number of process you want to run on?  Could the CICE model be part of a multi-model system where somehow the MPI communicator doesn't match the number of processors the ice model should run on?  Second, this feature provides a check.  Even in a standalone mode, the size of the communicator is going to depend on how the model was launched (via mpiexec or similar).  Is it helpful to have an internal check that the number of processors specified by the user in namelist actually matches the number of processors in the communicator, especially since block size and many other things will depend on the number of processors in play?  Does the fact that MPI and OpenMP are supported together affect the usefulness of this feature?

Having said all that, I'm open to changing this.  In that case, I assume you're setting the block size and the max blocks will accommodate a range of processor counts.  And if the max blocks is too big, you just live with it (or maybe max blocks is set internally?).  

Maybe I'll suggest adding support for setting nprocs=-1, which would mean ignore it and use get_num_procs instead.  I think that provides backwards compatibility and support for cases where it may be useful, but allows users to rely on get_num_procs if they want.

anton-seaice:
There's many possibilities - but the code aborts if nprocs != get_num_procs, so any possibility that uses nprocs from the namelist as different to the system number of processors would need code modifications anyway.

We think we can improve the automated max_blocks estimate too, but it requires calling `proc_decomposition` twice, which is a bit messy.

apcraig:
[...]

PR #949, opened 2024-05-02
Autoset nprocs and improve max_blocks estimate

anton-seaice:
## PR checklist
    Contributes to #945  
This improves setting nprocs and max_blocks automatically. 

- [ ] Developer(s): 
   @anton-seaice @minghangli-uni
   @apcraig 
    Needs doing
    - [ ] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [ ] No
    - [ ] Yes
        - [ ] Yes
        - [ ] No 

This change allows `nprocs` to be set to `-1` in 'ice_in' and then the number of processors will be automatically detected.

This change improves the automatic calculation of `max_blocks` to give a better (but still not foolproof) estimate of max_blocks if it is not set in `ice_in`.


apcraig:
This looks pretty good.  A couple thoughts.

- Setting nprocs=get_num_procs does not happen all the time for the cpp CESMCOUPLED.  I think we'll want backwards compatibility on that for now.
- The max_blocks estimate is improved, but it is still focused on even distribution of the 2d blocks.  There are decompositions where it still will not work in general, decompositions that are more equal weighted in terms of estimated work vs number of blocks.

The modifications I'm working on handle the nprocs thing in much the same way, but also refactor the max_blocks calculation so each proc (mpi task) has a locally defined max_blocks that matches the decomposition exactly.  I'm still developing and testing to make sure it'll work properly.

I'm happy to merge this as a first step (if the CESMCOUPLED thing is fixed).  I could then update the implementation if I make additional progress.  Thoughts?

anton-seaice:
> I'm happy to merge this as a first step (if the CESMCOUPLED thing is fixed). I could then update the implementation if I make additional progress. Thoughts?

Lets go with this. I think I made the related changes to do this.

I guess we should add some tests?

cicecore/cicedyn/infrastructure/ice_domain.F90
```diff
@@ -242,16 +257,6 @@ subroutine init_domain_blocks
       !*** domain size zero or negative
       !***
       call abort_ice(subname//' ERROR: Invalid domain: size < 1', file=__FILE__, line=__LINE__) ! no domain
-   else if (nprocs /= get_num_procs()) then
-      !***
-      !*** input nprocs does not match system (eg MPI) request
```
phil-blain, line 247: we could keep this comment, I think

cicecore/cicedyn/infrastructure/ice_domain.F90
```diff
@@ -216,21 +216,36 @@ subroutine init_domain_blocks
...
+      max_blocks=(((nx_global-1)/block_size_x + 1) *         &
+                  ((ny_global-1)/block_size_y + 1) - 1) / nprocs + 1
+      max_blocks=max(1,max_blocks)
+      if (my_task == master_task) then
+         write(nu_diag,'(/,a52,i6,/)') &
+            '(ice_domain): max_block < 1: max_block estimated to ',max_blocks
+      endif
+   endif
+   
+
```
phil-blain, line 248: can we avoid whitespace-only lines here ? Also I would think a single blank line is sufficient.

doc/source/user_guide/ug_case_settings.rst
```diff
@@ -370,6 +370,7 @@ domain_nml
    "``maskhalo_bound``", "logical", "mask unused halo cells for boundary updates", "``.false.``"
    "``max_blocks``", "integer", "maximum number of blocks per MPI task for memory allocation", "-1"
    "``nprocs``", "integer", "number of processors to use", "-1"
+   "", "``-1``", "find number of processors automatically", ""
```
phil-blain, line 373: I think while we're in here, "MPI ranks" would be more accurate than "processors" (for both the new line and the existing one).

phil-blain (commented):
Thanks Anton.

> This change allows nprocs to be set to -1 in 'ice_in' and then the number of processors will be automatically detected.
> 
> This change re-orders the calculation of max_blocks to ensure nprocs is set before the calculation.

This description reflects the latest state of the PR, super.

cicecore/cicedyn/infrastructure/ice_domain.F90
```diff
@@ -216,21 +216,36 @@ subroutine init_domain_blocks
...
    call broadcast_scalar(block_size_x,      master_task)
    call broadcast_scalar(block_size_y,      master_task)
    call broadcast_scalar(nx_global,         master_task)
    call broadcast_scalar(ny_global,         master_task)
 
+   ! Set nprocs if not set in namelist
+#ifdef CESMCOUPLED
+   nprocs = get_num_procs()
+#else
+   if (nprocs == -1) then
```
apcraig, line 229: Lets make this (nprocs < 0)?

cicecore/cicedyn/infrastructure/ice_domain.F90
```diff
@@ -216,21 +216,36 @@ subroutine init_domain_blocks
...
-         '(ice_domain): max_block < 1: max_block estimated to ',max_blocks
-     endif
-   endif
    call broadcast_scalar(max_blocks,        master_task)
    call broadcast_scalar(block_size_x,      master_task)
    call broadcast_scalar(block_size_y,      master_task)
    call broadcast_scalar(nx_global,         master_task)
    call broadcast_scalar(ny_global,         master_task)
 
+   ! Set nprocs if not set in namelist
```
apcraig, line 225: Is it better to compute this on all procs or to compute it on master proc then broadcast?  I'm not sure one is better than the other, but we certainly want to make sure each proc has the same value, so maybe we should do it before the broadcast?

Also, why did you revert the improved max_blocks computation back to the original version?  If your implementation is more robust, let upgrade, even if it might change again.

apcraig:
I will run a full test suite once comments are addressed and the PR is no longer draft.  Thanks!

anton-seaice:
> I will run a full test suite once comments are addressed and the PR is no longer draft. Thanks!

My question here is whether we should add some test cases for `nprocs=-1`. It may be a bit of work though, because most of the scripts are written around setting this explicitly.

cicecore/cicedyn/infrastructure/ice_domain.F90
anton-seaice: Isn't this clear from the ERROR message?

cicecore/cicedyn/infrastructure/ice_domain.F90
anton-seaice: Part A: This was just a stylistic decision. It seems easier to read / follow / modify if all the broadcast scalar calls together. The result is the same.

Part B: Oh I misread / I misunderstood. Ok the max_blocks using proc_decomposition is back in :)

cicecore/cicedyn/infrastructure/ice_domain.F90
anton-seaice: Ok done

cicecore/cicedyn/infrastructure/ice_domain.F90
anton-seaice: I made it one line instead of two :)

doc/source/user_guide/ug_case_settings.rst
anton-seaice: We seem to use "processors" throughout the rest of the docs, although they seem light on for details about MPI in general. 

I definitely out of my depth here .. but google implies the rank is just the unique name for each process ? So this should be number of "processes" as it could be possible to have multiple processes per processor ! (And by this logic, we should change all the other "processors" to "processes" in the docs)

apcraig:
> > I will run a full test suite once comments are addressed and the PR is no longer draft. Thanks!
> 
> My question here is whether we should add some test cases for `nprocs=-1`. It may be a bit of work though, because most of the scripts are written around setting this explicitly.

Please run a test manually with nprocs=-1 to make sure it works.  I think the question is maybe whether we should have nprocs=-1 be the default setup for all tests.  You could try that and see if everything runs and is bit-for-bit.  I think you'd need to set nprocs=-1 in configuration/scripts/ice_in and then remove "nprocs            = ${task}" in cice.setup.  Is that the direction we want to go?

doc/source/user_guide/ug_case_settings.rst
apcraig: I think "MPI tasks" is technically correct here.  Processors and processes are both wrong.  "MPI ranks" is probably the same thing, but not terminology we generally use in the document.  Lets go with "MPI tasks"?

anton-seaice: Is "Number of MPI tasks (typically 1 per processor) to use" good?

There are lots of uses of "processor" in the docs, I guess these could be changed to task? Should I make a new issue?

apcraig: "typically 1 per processor" would not be a good thing to add.  The way I think about this is that processors are the hardware in use while tasks and threads are how those processors are used.  So nprocs is specifically associated with the number of MPI tasks only.  It has nothing to do with threads.  If you are running 16 mpi tasks threaded 2 ways each, you are using 32 processors and have 16 MPI tasks.  nprocs is MPI tasks and has nothing to do with the "processors" or "processes" used/set.  To me, processes is a word separate from hardware.  If you could oversubscribe threads on a processors, you might have 1 processor but 4 processes on that processor.  

Again, this is how I think about things and probably how most of the documentation uses that terminology.  I fully acknowledge that my ideas about this may not be the norm.  In that case, we should update the documentation using more standard language.  But before we do, we should carefully think about the terminology.  At this point, I would just update this one line of documentation unless you see other glaring errors.  We could open an issue, but before we do, I'd like for someone to review the documentation (at least a few sections here and there) to see whether there seems to be a bunch of incorrect terminology.  I prefer to have an issue that says "several things need to be fixed such as...." rather than "review documentation for correct/consistent use of tasks, threads, processors, processes".

anton-seaice: Thanks Tony. Ill just make it say "Number of MPI tasks". I am a bit out of my depth re the best terminology for the documentation, so ill leave it for now.

apcraig: I think you still need to update the documentation here.  Will start some testing on the changes.

anton-seaice: I made the update, I wonder if it no longer makes sense if compiling with the 'serial' folder instead of the 'mpi' folder ?

anton-seaice:
Our intermittent failure is back !

https://github.com/CICE-Consortium/CICE/actions/runs/9024916341/job/24799676376 failed 
https://github.com/ACCESS-NRI/CICE/actions/runs/9024916168/job/24799675998 passed

But they are the same commit

apcraig:
I restarted the github action failure and it passed.  Ugh....

apcraig (approved):
Full test suite run on derecho with intel, all passed.

PR #964, opened 2024-08-02
Enable prescribed ice for UFS

NickSzapiro-NOAA:

## PR checklist
    Enable prescribed ice CICE for UFS in nuopc/cmeps cap
- [x] Developer(s): 
    Nicholas Szapiro & Denise Worthen
   Regression test atm_ds2s_docn_pcice added to ufs-weather-model (https://github.com/ufs-community/ufs-weather-model/pull/2186) 
    - [x] bit for bit (when not in prescribed ice)
    - [x] more substantial (when in prescribed ice #ifndef cesmcoupled) 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 
Prescribed ice CICE was enclosed by #ifdef CESMCOUPLED in CMEPS driver. This allows use of prescribed ice CICE in ufs-weather-model with CDEPS calls to input (sea ice concentration) from file. Starting as cherry-picked as in UFS but dependency on CDEPS likely needs to be optional at compile time. This is needed for https://github.com/NOAA-EMC/CICE/issues/84

### DIFF
--- a/cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90
+++ b/cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90
@@ -7,33 +7,15 @@ module ice_prescribed_mod
   ! Ice/ocean fluxes are set to zero, and ice dynamics are not calculated.
   ! Regridding and data cycling capabilities are included.
 
-  use ESMF, only : ESMF_Clock, ESMF_Mesh, ESMF_SUCCESS, ESMF_FAILURE
+  use ESMF, only : ESMF_GridComp, ESMF_Clock, ESMF_Mesh, ESMF_SUCCESS, ESMF_FAILURE
   use ESMF, only : ESMF_LogFoundError, ESMF_LOGERR_PASSTHRU, ESMF_Finalize, ESMF_END_ABORT
-
-#ifndef CESMCOUPLED
-
-  use ice_kinds_mod
-  implicit none
-  private ! except
-  public  :: ice_prescribed_init      ! initialize input data stream
-  logical(kind=log_kind), parameter, public :: prescribed_ice = .false.     ! true if prescribed ice
-contains
-  ! This is a stub routine for now
-  subroutine ice_prescribed_init(clock, mesh, rc)
-    type(ESMF_Clock)       , intent(in)  :: clock
-    type(ESMF_Mesh)        , intent(in)  :: mesh
-    integer                , intent(out) :: rc
-    ! do nothing
-    rc = ESMF_SUCCESS
-  end subroutine ice_prescribed_init
-
-#else
 
   use ice_kinds_mod
   use shr_nl_mod       , only : shr_nl_find_group_name
   use dshr_strdata_mod , only : shr_strdata_type, shr_strdata_print
   use dshr_strdata_mod , only : shr_strdata_init_from_inline, shr_strdata_advance
   use dshr_methods_mod , only : dshr_fldbun_getfldptr
+  use dshr_mod         , only : dshr_pio_init
   use ice_broadcast
   use ice_communicate   , only : my_task, master_task, MPI_COMM_ICE
   use ice_fileunits
@@ -43,7 +25,7 @@ contains
   use ice_blocks        , only : nx_block, ny_block, block, get_block
   use ice_domain        , only : nblocks, distrb_info, blocks_ice
   use ice_grid          , only : TLAT, TLON, hm, tmask, tarea, grid_type, ocn_gridcell_frac
-  use ice_calendar      , only : idate, calendar_type
+  use ice_calendar      , only : idate, msec, calendar_type
   use ice_arrays_column , only : hin_max
   use ice_read_write
   use ice_exit          , only: abort_ice
@@ -74,13 +56,17 @@ contains
 contains
 !===============================================================================
 
-  subroutine ice_prescribed_init(clock, mesh, rc)
+  subroutine ice_prescribed_init(gcomp, clock, mesh, rc)
 
     ! Prescribed ice initialization
 
-    include 'mpif.h'
+#ifndef SERIAL_REMOVE_MPI
+    !TODO: add 1d character array to cicecore/cicedyn/infrastructure/comm/mpi/ice_broadcast.F90
+    use mpi   ! MPI Fortran module
+#endif
 
     ! input/output parameters
+    type(ESMF_GridComp)    , intent(in)  :: gcomp
     type(ESMF_Clock)       , intent(in)  :: clock
     type(ESMF_Mesh)        , intent(in)  :: mesh
     integer                , intent(out) :: rc
@@ -93,6 +79,7 @@ contains
     character(len=char_len_long)     :: stream_dataFiles(nFilesMaximum)
     character(len=char_len_long)     :: stream_varname
     character(len=char_len_long)     :: stream_mapalgo
+    character(len=char_len_long)     :: stream_taxmode
     integer(kind=int_kind)           :: stream_yearfirst   ! first year in stream to use
     integer(kind=int_kind)           :: stream_yearlast    ! last year in stream to use
     integer(kind=int_kind)           :: stream_yearalign   ! align stream_year_first
@@ -110,6 +97,7 @@ contains
          stream_varname ,               &
          stream_datafiles,              &
          stream_mapalgo,                &
+         stream_taxmode,                &
          stream_yearalign,              &
          stream_yearfirst ,             &
          stream_yearlast
@@ -125,6 +113,7 @@ contains
     stream_meshfile     = ' '
     stream_datafiles(:) = ' '
     stream_mapalgo      = 'bilinear'
+    stream_taxmode      = 'cycle'
 
     ! read namelist on master task
     if (my_task == master_task) then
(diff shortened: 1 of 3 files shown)


NickSzapiro-NOAA:
The dependency on CDEPS is not a problem in UFS, but I'm not sure the best way to handle it here. Maybe a different #ifdef is needed in cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90 ?

cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90
```diff
@@ -7,33 +7,15 @@ module ice_prescribed_mod
...
-  end subroutine ice_prescribed_init
-
-#else
-
   use ice_kinds_mod
   use shr_nl_mod       , only : shr_nl_find_group_name
   use dshr_strdata_mod , only : shr_strdata_type, shr_strdata_print
   use dshr_strdata_mod , only : shr_strdata_init_from_inline, shr_strdata_advance
   use dshr_methods_mod , only : dshr_fldbun_getfldptr
+  use dshr_mod         , only : dshr_pio_init
```
dabail10, line 18: These come from the CIME shr stuff I believe. This is how we do the reading from stream files.

dabail10 (commented):
This looks fine as far as I can tell. I will run with it on Monday to make sure.

NickSzapiro-NOAA:
Thanks Dave! So most of the code in cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90 was enclosed in an #ifndef CESMCOUPLED else. There are the functions now "uncovered" to use the CDEPS inline functionality:
 https://github.com/ufs-community/ufs-weather-model/pull/2186/files#diff-db20ac41c986cab4cffeeaf178a6a4abf1c092a137af4cbe274e19deb7326954

anton-seaice:
If I am reading this correctly, there were already some depedencies on CDEPS (i.e. other `use dshr_` statements) in this file and no-one has raised it as an issue? So I think we can just leave it be until someone wants to build without CDEPS? I have a vague recollection of someone saying there are other unexpected dependencies between CMEPS & CDEPS in places, so I don't think this would break anything particularly.

dabail10:
There was only a dependency on CDEPS when CESMCOUPLED was defined. I guess my concern is that we have code duplication here. Will we need two copies of the CDEPS library? What happens when they get out of sync?

cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90
```diff
@@ -211,6 +207,11 @@ subroutine ice_prescribed_init(clock, mesh, rc)
           hin_max(1) = 999._dbl_kind
        end if
 
+#ifndef CESMCOUPLED
+       ! If need initial cice values for coupling
+       call ice_prescribed_run(idate, msec)
+#endif
```
NickSzapiro-NOAA, line 213: UFS needed this call ice_prescribed_run at the end of ice_prescribed_init. I don't know if it's a bug for CESM not to. Presumably, the CICE state should be initialized in init, but it may depend on the components' run sequence. We left the #ifndef CESMCOUPLED so as not to impact CESM @dabail10

NickSzapiro-NOAA:
The ice prescribed code is only useful if CICE is linked with CDEPS. I didn't think including CDEPS with CICE was an option. I'm happy to block it all off if not...maybe #ifdef CICE_HAS_CDEPS (?)

dabail10:
I guess I am confused. I understand that you want to use the prescribed ice and hence CDEPS. What I think you are saying is that you would have a separate set of modules / subroutines from CDEPS that would come with CICE. Kind of like what is done for the shr_orb routines. When we use CICE in CESM, then CDEPS is part of the CESM checkout. One thing you might be able to do is have CDEPS as a submodule / external for CICE. This is only checked out when using UFS. Perhaps I am missing something still?

NickSzapiro-NOAA:
Sorry for being confusing.

UFS already has CDEPS as a subcomponent. This code builds and runs fine in UFS (with appropriate linking). It shouldn't change anything for CESM either.

I was worried if anyone else uses this driver and does not have CDEPS. If that is not a concern, I agree with @anton-seaice that this code is fine as is until further notice. No stubs (or CDEPS submodule) are needed here then

cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90
```diff
@@ -7,33 +7,15 @@ module ice_prescribed_mod
   ! Ice/ocean fluxes are set to zero, and ice dynamics are not calculated.
   ! Regridding and data cycling capabilities are included.
 
-  use ESMF, only : ESMF_Clock, ESMF_Mesh, ESMF_SUCCESS, ESMF_FAILURE
+  use ESMF, only : ESMF_GridComp, ESMF_Clock, ESMF_Mesh, ESMF_SUCCESS, ESMF_FAILURE
   use ESMF, only : ESMF_LogFoundError, ESMF_LOGERR_PASSTHRU, ESMF_Finalize, ESMF_END_ABORT
 
-#ifndef CESMCOUPLED
```
anton-seaice, line 13: I think we should leave this section, and add a UFSCOUPLED or a NO_CDEPS FLAG .

anton-seaice:
> There was only a dependency on CDEPS when CESMCOUPLED was defined. I guess my concern is that we have code duplication here. 

Sorry - you are right. I didn't realise most of the ice_prescribed_mod file was within the CESM coupled ifdef. 

I think we shouldn't do something that might require others to add a compiler flag if there existing setup works ... ? I added a comment to this effect - https://github.com/CICE-Consortium/CICE/pull/964/files#r1708080117

dabail10:
I like this strategy of adding a NOCDEPS flag or something similar. However, how many groups use the nuopc/cmeps driver? If it is only the three of us and we all use CDEPS, then I would advocate for removing the CESMCOUPLED flag here.

DeniseWorthen:
> I like this strategy of adding a NOCDEPS flag or something similar. However, how many groups use the nuopc/cmeps driver? If it is only the three of us and we all use CDEPS, then I would advocate for removing the CESMCOUPLED flag here.

I don't think any ifdef is required in the ``ice_prescribed_mod`` at this point. If someone were to attempt to use this mode w/o CDEPS, they would probably either a) fail compile or b) fail at run time when the ``ice_prescribed_init ``routine tries to call the ``dshr_pio_init`` or ``shr_strdata_init_from_inline``.

dabail10:
We should probably add an informative message here to let people know that CDEPS is required.

NickSzapiro-NOAA:
> We should probably add an informative message here to let people know that CDEPS is required.

I don't see any docs for drivers/nuopc/cmeps . May I ask where I should inform of this dependency on CDEPS?

dabail10:
There is no documentation for the driver per se. I think we just need a graceful abort that checks the first call to CDEPS and prints a message saying that CDEPS is required for this module or something like that.

NickSzapiro-NOAA:
Using this code as is without any "substitute" ifdefs, the nuopc/cmeps driver code should fail to compile without CDEPS. This isn't a practical problem as CDEPS comes with CMEPS in coupled models we know of. @anton-seaice, do y'all share this driver too?

It is useful to enable this prescribed ice code more generally beyond CESMCOUPLED. 

Maybe the position here is: Why "clutter" the code with ifdefs, stub routines, etc. unless needed? Especially if just to abort. This is of course a Consortium decision and open to other solutions.

dabail10:
Excellent point. So, I agree with removing the ifdefs and it is just buyer beware for those using the nuopc/cmeps driver.  Maybe something in the comments about the dependency on CDEPS? That way if it fails to compile, someone could look in ice_prescribed_mod.F90 and see this note.

apcraig:
My feeling is that if the nuopc/cmeps implementation is always used with cdeps as far as we know, then get rid of the CPPs.  Especially if they are not tested.  If someone comes along in the future that uses nuopc/cmeps and doesn't have cdeps (and doesn't need prescribed ice mode), they can add the CPPs and test it at that time.

cicecore/drivers/nuopc/cmeps/ice_prescribed_mod.F90
NickSzapiro-NOAA: It seems cleaner to admit the dependency on CDEPS. Is this ok @anton-seaice?

anton-seaice: Consensus seems to be to go with the PR as currently written

anton-seaice (approved):
I checked this builds and starts, we don't use prescribed_ice so I didn't test that.

apcraig:
Is this ready to merge?  @dabail10, can you review/confirm when you have a chance.  It's fine by me but have no stake in and have not tested these changes.  Thanks.

dabail10:
I have not had time to test this. I think it is fine.

apcraig:
@dabail10.  I'll wait for you to approve.  I'm not in a rush.  Will leave it up to you whether to test or not before merging.

dabail10:
Just tested this in CESM and it passed a simple debug restart test.

dabail10 (approved):

apcraig (approved):

PR #985, opened 2024-10-18
Allow kmt variable name of "mask" and different size netcdf 2d reads

anton-seaice:
## PR checklist
    This is two bits of tidy up in the grid infrastructure, to make further extensions easier. I'm doing this PR now so the changes are easier to review.
- [ ] Developer(s): 
    @anton-seaice
@apcraig
    Still running, will update next week
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 

This change allows for a single processor read of netcdf variables that are not of size nx_global by ny_global. This is to make it easier to extent the ice_grid routines.

This changes allows the mask variable in the kmt netcdf file to be called "mask" instead of "kmc". Both options now work.

This change splits reading the mask and grid into seperate subroutines, so the routine can be reused independently later.


apcraig:
This all looks pretty reasonable.  When you have test results and feel it's ready, let me know.  I will run some independent testing, review, and merge.  Thanks!

anton-seaice:
I ran the whole base_suite, i needed to extend the run-time for several tests ? And sometimes by quite a big increase in time.

The results were bit for bit / no regressions were found.

This is ready to merge from my perspective

eclare108213:
> I ran the whole base_suite, i needed to extend the run-time for several tests ? And sometimes by quite a big increase in time.

@anton-seaice Were there corresponding runs with debugging flags on for any of the slow tests?  Could there be NaNs in those cases?

anton-seaice:
Sorry - my comment wasn't clear!

I needed to increase the run-time for the tests both before and after the change. It was the tests with small core counts that seemed to be more affected, so I wonder if maybe our job scheduler is different or compiler options or something need adjusting. It's not related to this change.



> @anton-seaice Were there corresponding runs with debugging flags on for any of the slow tests? Could there be NaNs in those cases?

It doesn't look like base_suite has tests where the only different between two tests is just debug.

apcraig:
I know some of the tests were timing out.  If we add "short" and so forth to the test, we lose the regression testing because the case name changes.  I can live with that but want to avoid doing that often.  Lets go with it this time, but maybe we can think about whether there are other ways to set the requested queue time without it changing the casename in the future.  I'll think about it too.

apcraig:
Ran the quick suite on derecho with 6 compilers and all tests pass, tested done with #986 . Will merge this week, trying to merge fsd mods first.

apcraig (approved):

Issue #977, opened 2024-10-01
Add tripolar netcdf grid to input data


anton-seaice:
I am looking at writing tests for the grid load routines, and there doesn't seem to be any test coverage for a tripolar grid in netcdf format? Have I missed something?

p.s. We should add one.

apcraig:
That's probably correct.  The grid files used in Consortium testing were generated in the user community.  It's possible we don't have an example of a tripole netcdf grid file.

anton-seaice:
Shall we add a tx1nc ? Is this a CESM grid? Someone make a bathymetry netcdf for it already, so we could create a netcdf if needed.

dabail10:
The tx1 is never used. It is just meant to be a test for tripole grids. I think that we should stick with standard tripole grids that are actually used with MOM. We have a 2/3 degree tripole. NOAA/EMC uses a 0.5-degree and 0.25 degree. Can we also couple this with reading MOM supergrids?

anton-seaice:
I suggest we add both a "normal/CICE/POP" grid now, and a supergrid when that gets implemented ? The lower the resolution the better in my view - they are to demonstrate the tripole works and tests run faster on a lower resolution

(ill also suggest we just drop the binary grid formats, but thats a different discussion :-))

anton-seaice:
I realised that the forcing data in our test cases is offline regridded to the grid format used. As we already have forcing data in "tx1" format, its a strong argument to re-use "tx1" in a netcdf and a mom-mosaic-netcdf.

anton-seaice:
Does tx1 exist already in a tripole netcdf and mom supergrid formats ?

Or should we just create a new tripole grid (at say 2 or 4 degrees nominal) + a new set of forcing data and use that instead?

dabail10:
Definitely not in MOM supergrid format. I'm not sure we even have a netcdf version. Though, this could be created easily enough.

anton-seaice:
Back on thinking about this for the mom supergrid format - is there a preference for 1 degree or 0.25 degree grid ?

Does forcing data need to be generated using scripts in this repository, or any valid method? (the forcing data if not generated through this repo would have enough metadata to capture how it was made still)

apcraig:
[...]

PR #993, opened 2024-11-07
Add support for mom (supergrid) grid format

anton-seaice:
This changes adds `grid_format = 'mom_nc'` support to directly load the grid from a netcdf file which uses the mom (supergrid) definition, and calculates all fields to follow the method MOM6 uses. i.e.
- all lats & lons are read from the file
- all cell x/y lengths are read from the file
- A grid and B grid areas are read from the file, C grid areas are calculated internally (per https://github.com/NOAA-GFDL/MOM6/issues/740)
- angles are calculated internally (even though they exist in the mosaic file)

There are a few limitations: 
- If the CHANNEL_CONFIG option is used in MOM6, this can change some cell sizes at runtime, however we are loading these from the file.
- MOM6 generates a land-sea mask by making land where the bathymetry<0.5m , however we are still loading from a kmt file which needs to be generated "offline".

I haven't tested every permutation of boundary conditions and grid type, I focussed on on tripolar grids and did best-effots checks on the other boundary conditions.

I found exact matches (to single point precision) between the MOM6 & CICE6 saved grids, see https://github.com/anton-seaice/sandbox/blob/main/new_cice_grid/mom_mosaic_in_cice.ipynb

The NUOPC driver checks for consistency in areas, and lon/lats between the model and the mediator. It applies a correction for differences in area between the model and the mediator. Even if the cell area field is populated in the mediator mesh file, this is not used and the area calculated by ESMF is used instead. Therefore depending on how the mosaic file was created (and how the areas are calculated for that process), the correction between model and mediator areas may not be 1 everywhere.

This change doesn't include any new tests, and I suggest deferring adding tests as it requires new forcing data

I also added _kmt_type = 'none'_ - This is only for testing / idealised configurations, so it's not in the docs. It's not essential for this PR.

## PR checklist
    Add support for creating the CICE grid from a MOM mosaic (supergrid) format
- [x] Developer(s): 
   @anton-seaice 
   @dabail10 
    To-do, if the change & scope looks ok, I will run the cice standalone tests, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#07bb8a004f0dff968441bf1f5852d79680d3d514
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No - blocked on needing new test data, see https://github.com/CICE-Consortium/CICE/issues/977#issuecomment-2406033051
    - [x] Yes
        - [ ] Yes
        - [ ] No 
This changes adds `grid_format = 'mom_nc'` support to directly load the grid from a netcdf file which uses the mom mosaic (supergrid) definition, and calculates all fields to follow the method MOM6 uses. Contributes to #807


eclare108213:
@anton-seaice, it looks like 'mosaic' in this context has nothing to do with the field campaign.  Is that correct?  If so, it could be confusing for other users interested in the MOSAiC data.  Let's figure out how to document both to reduce confusion.  Certainly explain what they are in the readthedocs, but also perhaps add some comments to the diagnostic output?

dabail10:
> @anton-seaice, it looks like 'mosaic' in this context has nothing to do with the field campaign. Is that correct? If so, it could be confusing for other users interested in the MOSAiC data. Let's figure out how to document both to reduce confusion. Certainly explain what they are in the readthedocs, but also perhaps add some comments to the diagnostic output?

I agree. Maybe mosaic is not the right term to use here. Maybe mom_supergrid or just 'mom'. Also, I would add that grid_type has the following options currently: 'displaced_pole', 'tripole', 'default', 'regional', and 'latlon'. So, perhaps we should do something like 'mom_tripole'?

doc/source/user_guide/ug_implementation.rst
```diff
@@ -149,6 +151,15 @@ and ANGLE variables. If the variables exist, ANGLET, TLON and TLAT will also be
 read from a netcdf grid file. The kmt (mask) netcdf file needs a variable named 
 kmt or mask, set to 0 for land and 1 for ocean.
 
+For the *mom_mosaic* formats (``grid_type='mom_mosaic'``), a grid file following
```
anton-seaice, line 154:
```suggestion
For the *mom_mosaic* formats (``grid_format='mom_mosaic'``), a grid file following
```

anton-seaice:
Thanks both

Yes definitely unrelated to MOSAiC voyages.

The term mosaic and supergrid are used a little bit interchangeably, although most of the MOM6 source code and the ESMF docs refer to a mosaic specification, so I used that. Happy to change to calling it supergrid if that is clearer. I didn't want to only refer to it as the MOM format because there are is an earlier mom format which is different.

@dabail10 - apologies - there was an error in the description, which i've just amended. This changes adds `grid_format = mom_mosaic` and it supports the existing `grid_type` options of `regional`, `tripole` and `displaced_pole`. The only one it doesn't support is `tripoleT` and aborts if that configuration is set.

dabail10:
> Thanks both
> 
> Yes definitely unrelated to MOSAiC voyages.
> 
> The term mosaic and supergrid are used a little bit interchangeably, although most of the MOM6 source code and the ESMF docs refer to a mosaic specification, so I used that. Happy to change to calling it supergrid if that is clearer. I didn't want to only refer to it as the MOM format because there are is an earlier mom format which is different.
> 
> @dabail10 - apologies - there was an error in the description, which i've just amended. This changes adds `grid_format = mom_mosaic` and it supports the existing `grid_type` options of `regional`, `tripole` and `displaced_pole`. The only one it doesn't support is `tripoleT` and aborts if that configuration is set.

I understand now. However isn't grid format either nc or bin? Maybe this should be mom_nc?

cicecore/cicedyn/general/ice_init.F90
```diff
@@ -2692,14 +2692,16 @@ subroutine input_data
...
+          kmt_type  /=  'none' ) then
          if (my_task == master_task) write(nu_diag,*) subname//' ERROR: unknown kmt_type=',trim(kmt_type)
          abort_list = trim(abort_list)//":27"
       endif
 
       if (grid_type  /=  'column'      .and. &
           grid_type  /=  'rectangular' .and. &
-          kmt_type   /=  'file') then
+          kmt_type   /=  'file' .and. &
+          kmt_type   /= 'none') then
```
dabail10, line 2704: Also, the supergrid does not have depth. There is a separate bathymetry file. "KMT" is the old style pop depth file which just lists the lowest depth level which is water. We use KMT to derive the mask, but also the basal stress formulation will use the actual depth.

anton-seaice:
> I understand now. However isn't grid format either nc or bin? Maybe this should be mom_nc?

I figured there would never be a mom_mosaic grid not in a netcdf format, so i didn't think we needed to included `nc`. 

If it is clearer we could use `mom_nc` or leave grid_type alone and add a new namelist variable (e.g. `grid_specification` and it could be type `pop` or `mom_mosaic`)

dabail10:
> > I understand now. However isn't grid format either nc or bin? Maybe this should be mom_nc?
> 
> I figured there would never be a mom_mosaic grid not in a netcdf format, so i didn't think we needed to included `nc`.
> 
> If it is clearer we could use `mom_nc` or leave grid_type alone and add a new namelist variable (e.g. `grid_specification` and it could be type `pop` or `mom_mosaic`)

This might be better. However, I think this is overkill. I think grid_format = 'pop_nc' or 'mom_nc' is fine. I also really think you should *not* have "mosaic" in the name. This is going to be confusing.

cicecore/cicedyn/general/ice_init.F90
anton-seaice: I was just using kmt_type of 'none' for testing, for now in runs we would still use kmt_type = file

In some ways, it might be better to rename kmt_type to mask_type with options of `kmt_file`, `mask_file` and in the future `bathymetry_file` ?

anton-seaice:
> This might be better. However, I think this is overkill. I think grid_format = 'pop_nc' or 'mom_nc' is fine. I also really think you should _not_ have "mosaic" in the name. This is going to be confusing.

I guess is we just cover that a mom_grid refers to a mom mosaic / supergrid format in the docs then its sufficiently clear that is the format being referred to ? Its unlikely that configs using the latest CICE will want to use an old MOM grid format ?

dabail10:
> > > I understand now. However isn't grid format either nc or bin? Maybe this should be mom_nc?
> > 
> > 
> > I figured there would never be a mom_mosaic grid not in a netcdf format, so i didn't think we needed to included `nc`.
> > If it is clearer we could use `mom_nc` or leave grid_type alone and add a new namelist variable (e.g. `grid_specification` and it could be type `pop` or `mom_mosaic`)
> 
> This might be better. However, I think this is overkill. I think grid_format = 'pop_nc' or 'mom_nc' is fine. I also really think you should _not_ have "mosaic" in the name. This is going to be confusing.

Before I get more into this, I am so grateful that you've done the work on this!

I just talked to one of our MOM6 developers here. Here is what I think we should do:

For POP binary and netcdf files: grid_format = 'bin','nc' , kmt_type = 'file'. This maintains backward compatibility.

For MOM6 supergrid files: grid_format = 'mom6_nc' and kmt_type = 'none'. Then key off grid_format to see if we need to read in the MOM6 bathymetry file.

The "mosaic" file was an older MOM file. We will never have to support this. Our MOM developer has a script called mosaic_to_supergrid.py. I agree that we should stop using 'kmt' but again for backward compatibility, I think we leave it as kmt for now.

anton-seaice:
Hopefully it saves us generating many CICE grid files offline :)

Ok great - those namelist options are ok with me.

I haven't tried loading a bathymetry file yet, is the format of the CICE bathymetry file different to the MOM6 format ?

I was planning to,  in a later PR, generate the CICE land/sea mask from the bathymetry file. In the interim, we would keep using the current `kmt_type=file` for the land/sea mask - is that consistent with your thinking ?

dabail10:
> Hopefully it saves us generating many CICE grid files offline :)
> 
> Ok great - those namelist options are ok with me.
> 
> I haven't tried loading a bathymetry file yet, is the format of the CICE bathymetry file different to the MOM6 format ?
> 
> I was planning to, in a later PR, generate the CICE land/sea mask from the bathymetry file. In the interim, we would keep using the current `kmt_type=file` for the land/sea mask - is that consistent with your thinking ?

I think the MOM6 bathymetry is very different from the one CICE expects. So we will also have to use grid_format here. I definitely agree with keeping kmt_file around for the near future.

anton-seaice:
> For POP binary and netcdf files: grid_format = 'bin','nc' , kmt_type = 'file'. This maintains backward compatibility.
> 
> For MOM6 supergrid files: grid_format = 'mom6_nc' and kmt_type = 'none'. Then key off grid_format to see if we need to read in the MOM6 bathymetry file.

@apcraig - are these option names ok ? implementing the MOM6 bathymetry would be a future PR

apcraig:
I think the current plan is fine.  We should maintain the current namelist and backwards compatibility whenever possible.  I also agree that using "mosaic" could be confusing.

Would having a kmt_type='grid' make sense?  That would mean the kmt info is on the grid file?  Or maybe even kmt_type='grid' and kmt_type='grid_bathy'.  The former means the mask is read off the grid file, the latter means the bathymetry is read off the grid file then a kmt mask is created.  And how does that interact with the bathymetry namelist values?  I haven't looked at all the current and possible permutations of grid/kmt/bathy options, but maybe we should clarify all the existing and potential options and make sure we have a clear set of namelist values that make implementation of them clear (without breaking backwards compatibility).

anton-seaice:
Thanks - I change `mom_mosaic` to `mom_nc` and updated the comments / names etc as needed

anton-seaice:
Just marking this as a draft until i've resolved why this causes vertical truncations in the ocean/MOM

https://github.com/COSIMA/access-om3/issues/275

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -328,75 +329,120 @@ subroutine init_grid1
...
 
-      if (grid_type == 'tripole' .and. (mod(nx_global,2)/=0)) then
+      if (trim(grid_type) == 'tripole' .and. (mod(nx_global,2)/=0)) then
          call abort_ice(subname//' ERROR: grid_type tripole requires even nx_global number', &
                         file=__FILE__, line=__LINE__)
       endif
 
+      if (trim(grid_format) == 'mom_nc' .and. ns_boundary_type == 'tripoleT') then
+         call abort_ice(subname//" ERROR: ns_boundary_type='tripoleT' not implemented "&
+                        "for grid_format='mom_nc' tripole needs tripole ns_boundary_type", &
```
eclare108213, line 356: This error message might be difficult to understand.  Suggestion:
" ERROR: ns_boundary_type='tripoleT' not implemented "&
                        "for grid_format='mom_nc'. Use 'tripole' instead."

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -1344,6 +1434,701 @@ subroutine latlongrid
...
+      if (ierr/=0) call abort_ice(subname//' ERROR: Dealloc', file=__FILE__, line=__LINE__)
+
+      !-----------------------------------------------------------------
+      ! fin
+      !-----------------------------------------------------------------
+      call ice_close_nc(fid_grid)
+      l_readCenter = .true.       ! we have read t quantities 
+
+      end subroutine mom_grid
+
```
eclare108213, line 1591: The convention in CICE and Icepack is to separate subroutines using
`!=======================================================================`

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -1344,6 +1434,701 @@ subroutine latlongrid
...
+      !-----------------------------------------------------------------
+
+      if (my_task == master_task) then
+         allocate( &
+            work_mom(nx_global*2+1, ny_global*2+1), &
+            work_gE(nx_global+1,ny_global+1)      , &
+            work_gN(nx_global+1,ny_global+1)      , &
+            G_ULAT(nx_global+1,ny_global+1)       , & !1 bigger to include left and bottom
+            G_TLAT(nx_global+1,ny_global+1)       , & !1 bigger to include top and right
+            G_TLON(nx_global+1,ny_global+1)       , & !1 bigger to include left and bottom
```
eclare108213, line 1478: suggest removing '1 bigger to' from these comments.

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -1344,6 +1434,701 @@ subroutine latlongrid
...
+                         field_loc_NEcorner, field_type_scalar)
+      deallocate(G_tarea, G_uarea, stat=ierr)
+      if (ierr/=0) call abort_ice(subname//' ERROR: Dealloc error', file=__FILE__, line=__LINE__)
+
+      end subroutine mom_area
+
+
+      subroutine grid_rotation_angle(lon_cnr, lat_cnr, lon_cen, angle)
+      !  create angles in the same way mom6 creates the angle
+      !  based on https://github.com/mom-ocean/MOM6/blob/129e1bda02d454fb280819d1d87ae16347fd044c/src/initialization/MOM_shared_initialization.F90#L535
```
eclare108213, line 2091: Is there a normal reference for this, instead of a web page?  Put the web link in the documentation instead of the code itself?  Web pages aren't necessarily permanent (especially particular line numbers).

eclare108213 (changes requested):
This looks good to me, overall, with some minor suggestions.  Please remove Icepack from the PR.

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -328,75 +329,120 @@ subroutine init_grid1
...
+         stat=ierr &
+      )
+      if (ierr/=0) call abort_ice(subname//' ERROR: Out of memory', file=__FILE__, line=__LINE__)
 
       ! check tripole flags here
       ! can't check in init_data because ns_boundary_type is not yet read
       ! can't check in init_domain_blocks because grid_type is not accessible due to circular logic
 
-      if (grid_type == 'tripole' .and. ns_boundary_type /= 'tripole' .and. &
+      if (trim(grid_type) == 'tripole' .and. ns_boundary_type /= 'tripole' .and. &
```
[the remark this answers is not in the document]
anton-seaice: I am not sure why I changed this line, so ill revert this. 

It looks like initialization would fail before this point if there is extra whitespace in the namelist arguments (e.g. https://github.com/access-nri/cice/blob/d7ff38bd0232257ab83940cac37ea3b533eac5af/cicecore/cicedyn/general/ice_init.F90#L2672-L2673) 

I suspect we should do something like `grid_type = trim(grid_type)` during initialisation ? (possibly for all string namelist variables?) Although i also suspect that its highly unlikely people will set 

`  grid_type = "  tripole "`
when they mean 
`  grid_type = "tripole"`
?

cicecore/cicedyn/infrastructure/ice_grid.F90
anton-seaice: Thanks - done

cicecore/cicedyn/infrastructure/ice_grid.F90
anton-seaice: Thanks - done

cicecore/cicedyn/infrastructure/ice_grid.F90
anton-seaice: Thanks - done

apcraig:
@anton-seaice, is the PR ready for formal review and testing?

cicecore/cicedyn/infrastructure/ice_grid.F90
apcraig: Is a "break" still needed before subroutine mom_corners_global?

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -347,57 +351,98 @@ subroutine init_grid1
...
-            call ice_read_global(nu_kmt, 1,work_g2,'ida4',.true.)  ! KMT
+            case default 
 
-            if (my_task == master_task) then
-               close (nu_grid)
-               close (nu_kmt)
-            endif
+               call ice_open(nu_grid,grid_file,64) 
+               call ice_read_global(nu_grid,1,work_g1,'rda8',.true.) 
+               if (my_task == master_task) close (nu_grid)
```
apcraig, line 403: I am curious about use of 

    if (my_task == master_task) close (nu_grid)

vs

  call ice_close_nc(nu_grid)

Seems both approaches are used in this file and some ice_close_nc were changed to close (noted below).

anton-seaice: The first is for unformatted binary files and the second is for netcdf.

cicecore/cicedyn/general/ice_init.F90
```diff
@@ -2703,14 +2704,16 @@ subroutine input_data
...
```
anton-seaice, line 2716:
```suggestion
          kmt_type   /=  'none') then
```

anton-seaice:
Thanks @apcraig - this is ready for review. The issue I mentioned in the last meeting has been fixed (was using cm instead of m for a few fields).

cicecore/cicedyn/general/ice_init.F90
```diff
@@ -2703,14 +2704,16 @@ subroutine input_data
           kmt_type  /=  'channel_onenorth' .and. &
           kmt_type  /=  'wall'    .and. &
           kmt_type  /=  'default' .and. &
-          kmt_type  /=  'boxislands') then
+          kmt_type  /=  'boxislands'.and. &
+          kmt_type  /=  'none' ) then
```
apcraig, line 2708: I think kmt_type = 'none' should be documented.  Can note it's only for idealized testing in the namelist info.

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -347,57 +351,98 @@ subroutine init_grid1
...
+                  im = im + 2 
+               enddo
 
-            call ice_read_global_nc(fid_kmt,1,mask_fieldname,work_g2,.true.)
-            call ice_close_nc(fid_kmt)
+               deallocate(work_mom, stat=ierr)
+               if (ierr/=0) call abort_ice(subname//' ERROR: Dealloc error', file=__FILE__, line=__LINE__)
 
-         else
+            case('nc')
```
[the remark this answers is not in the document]
apcraig: I agree.  Lets add pop_nc and move to it.  We can leave nc as an undocumented option for backwards compatibility.  I could make this change as a separate PR if that would be preferred, but like the idea of doing it now.

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -534,6 +593,26 @@ subroutine init_grid2
...
+
+            do j = 1,ny_block
+            do i = 1,nx_block
+               tarea(i,j,iblk) = dxT(i,j,iblk)*dyT(i,j,iblk)
+               uarea(i,j,iblk) = dxU(i,j,iblk)*dyU(i,j,iblk)
+               narea(i,j,iblk) = dxN(i,j,iblk)*dyN(i,j,iblk)
+               earea(i,j,iblk) = dxE(i,j,iblk)*dyE(i,j,iblk)
+            enddo
+            enddo
+         enddo
```
apcraig, line 613: Looks like there might be a

!$OMP END PARALLEL DO

missing here.

cicecore/cicedyn/infrastructure/ice_grid.F90
apcraig: just a reminder to add this.

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -1346,6 +1435,708 @@ subroutine latlongrid
...
+         jlo = this_block%jlo
+         jhi = this_block%jhi
+
+         do j = 1,ny_block
+         do i = 1,nx_block
+            narea(i,j,iblk) = dxN(i,j,iblk)*dyN(i,j,iblk)
+            earea(i,j,iblk) = dxE(i,j,iblk)*dyE(i,j,iblk)
+         enddo
+         enddo
+      enddo
```
apcraig, line 1994: Is an !$OMP END PARALLEL DO needed here?

apcraig (commented):
Thanks for adding all the memory allocation/deallocation checks.  I added a few comments for further review.  Then I will run a large test suite on derecho with multiple compilers.  I hope we can add a mom grid testcase soon to the standalone model.

cicecore/cicedyn/infrastructure/ice_grid.F90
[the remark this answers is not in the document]
anton-seaice: Are we sure about the new name? I guess lots of users will have grids in this format, even though they don't use and never used POP? Lots of people moving from cice4 or 5 will have grids in this format and might find it confusing that it's now known as 'pop_nc' ?

anton-seaice:
Thanks for the feedback @apcraig - I will implement and update this PR soon

cicecore/cicedyn/infrastructure/ice_grid.F90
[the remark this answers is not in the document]
anton-seaice: Was there a decision about this? I can't tell from the meeting minutes ?

cicecore/cicedyn/infrastructure/ice_grid.F90
anton-seaice: I don't have a better reference - I think the link is useful but could remove it ?

cicecore/cicedyn/general/ice_init.F90
anton-seaice: Ok - i've done this. It needs another read still :)

cicecore/cicedyn/infrastructure/ice_grid.F90
anton-seaice: Thanks - done

cicecore/cicedyn/infrastructure/ice_grid.F90
anton-seaice: actually done this time :)

cicecore/cicedyn/infrastructure/ice_grid.F90
anton-seaice: thanks done

anton-seaice:
I think i've addressed the comments @apcraig & @eclare108213

cicecore/cicedyn/infrastructure/ice_grid.F90
eclare108213: It's okay in this case, thanks.

cicecore/cicedyn/infrastructure/ice_grid.F90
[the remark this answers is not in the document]
eclare108213: In the meeting I don't think there were strong feelings about using pop_nc (or not), except that we keep nc around for backward compatibility.  My personal preference is to add pop_nc here, since it is a model-specific format like mom_nc, and put it in the namelist files but still keep nc available in case someone doesn't change their ice_in.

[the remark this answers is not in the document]
anton-seaice: Ok - I am on leave next week but will implement this change.

Shall we change `grid_format = 'bin'` also (it could become `pop_bin`)

apcraig:
I am carrying out a full test suite on derecho.  If all looks good, I'll approve and merge tomorrow unless there are any other concerns.

cicecore/cicedyn/infrastructure/ice_grid.F90
```diff
@@ -347,57 +351,98 @@ subroutine init_grid1
                         file=__FILE__, line=__LINE__)
       endif
 
+      if (grid_format == 'mom_nc' .and. ns_boundary_type == 'tripoleT') then
+         call abort_ice(subname//" ERROR: ns_boundary_type='tripoleT' not implemented "&
```
apcraig, line 355:
```suggestion
         call abort_ice(subname//" ERROR: ns_boundary_type='tripoleT' not implemented "// &
```

cray compiler fails here.

apcraig:
With the one line fix above at ice_grid.F90, line 355, these changes pass the full test suite on derecho with 6 compilers.  Once the one line fix is added to the PR, I will approve and merge the PR.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#07bb8a004f0dff968441bf1f5852d79680d3d514

anton-seaice:
Thanks @apcraig - sounds good :)

apcraig (approved):

Issue #994, opened 2024-11-08
Synchronizing of fluxes especially when a point goes from ice free to ice covered.


dabail10:
In ice_flux.F90 (and icedrv_flux.F90 in icepack) we set flwout to zero so that it can be accumulated in merge_fluxes. However, there is a timestepping bug here such that if you have a point that goes from ice free to ice covered in a single timestep, then flwout will not be computed in merge_fluxes (because aicen_init is all zero) and stays zero. However, the coupler fluxes are passed based on aice, so for a single timestep flwout is passed to the coupler with a zero value. This causes a crash in CAM and likely is a problem in other atmospheric models. I'm not sure why this was never a problem before.

The solution is to remove the line:

flwout = c0

in ice_flux.F90 and icedrv_flux.F90 and only do this just before the call to merge_fluxes based on aice_init. Working on it now.

dabail10:
I have a solution. Let me know if you think this is ok. Basically, after merge_fluxes where lwout is computed, I make sure it never goes above the open ocean value (negative).

```
      ! Make sure flwout is not zero.

      TsfK = sst + Tffresh
      flwout = min(flwout,-stefan_boltzmann * TsfK**4)

```

anton-seaice:
@kieranricardo - this might be relevant to you ?

dabail10:
@peverwhee

dabail10:
Actually I have this backwards. The reasonable range of temperatures in sea ice covered regions would be:

-80C to 0C or 193.15 to 273.15K.

So, this means a range of 79 W/m^2 to 316 W/m2. So, this should be:

flwout = min(flwout, -79)

since flwout is negative.

dabail10:
Actually a better option is to add a check in scale_fluxes something like:

```
if (aice > puny .and. abs(flwout) < puny) then
   flwout = flwout_ocn
endif
```

This makes sure to catch the one special case of a point going from ice-free to ice-covered and the flwout is zero.

Dave

dabail10:
I have a PR in for this now, but it occurs to me that there may be similar problems with fswabs, fsens, flat ... while a zero value is valid in these fields, we should technically be setting points that go from ice free to ice covered to have an open ocean value for these. For example, you can look in init_coupler_flux for some of these initial settings for different seasons.

anton-seaice:
This doesn't solve you problem Dave, but I am slightly confused about the current behavior here.

[...]

PR #999, opened 2024-11-21
Fix outgoing lw for history and coupling.

dabail10:

## PR checklist
A bug was identified when a grid cell goes from ice free to ice covered. Issue #994
- [X] Developer(s): 
dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#6aa677ec49b415298b6368a19e5122f9c9d49f8e
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 
This is a two-line change in ice_flux.F90 (scale_fluxes) to check for the condition where aice > ), but flwout > -puny. This means the cell was ice free at the beginning of the step, but non-zero at the end. It sets the outgoing longwave to the open ocean value based on the freezing temperature. This is bfb, but will change the history output of flwup and flwup_ai. Test results are coming.

### DIFF
--- a/cicecore/cicedyn/general/ice_flux.F90
+++ b/cicecore/cicedyn/general/ice_flux.F90
@@ -1209,7 +1209,7 @@
       real (kind=dbl_kind) :: &
           ar, &   ! 1/aice
           stefan_boltzmann, &
-          Tffresh
+          Tffresh, puny
 
       integer (kind=int_kind) :: &
           i, j    ! horizontal indices
@@ -1217,7 +1217,7 @@
       character(len=*), parameter :: subname = '(scale_fluxes)'
 
       call icepack_query_parameters(stefan_boltzmann_out=stefan_boltzmann, &
-         Tffresh_out=Tffresh)
+         Tffresh_out=Tffresh, puny_out=puny)
       call icepack_warnings_flush(nu_diag)
       if (icepack_warnings_aborted()) call abort_ice(error_message=subname, &
          file=__FILE__, line=__LINE__)
@@ -1231,7 +1231,10 @@
             fsens   (i,j) = fsens   (i,j) * ar
             flat    (i,j) = flat    (i,j) * ar
             fswabs  (i,j) = fswabs  (i,j) * ar
+            if (flwout(i,j) > -puny) & 
+               flwout  (i,j) = -stefan_boltzmann *(Tf(i,j) + Tffresh)**4
             flwout  (i,j) = flwout  (i,j) * ar
+            ! Special case where aice_init was zero and aice > 0.
             evap    (i,j) = evap    (i,j) * ar
             Tref    (i,j) = Tref    (i,j) * ar
             Qref    (i,j) = Qref    (i,j) * ar
--- a/cicecore/cicedyn/general/ice_flux.F90
+++ b/cicecore/cicedyn/general/ice_flux.F90
@@ -1231,10 +1231,10 @@
             fsens   (i,j) = fsens   (i,j) * ar
             flat    (i,j) = flat    (i,j) * ar
             fswabs  (i,j) = fswabs  (i,j) * ar
+            ! Special case where aice_init was zero and aice > 0.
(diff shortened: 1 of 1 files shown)


dabail10:
Here are the test results:

https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#6aa677ec49b415298b6368a19e5122f9c9d49f8e

cicecore/cicedyn/general/ice_flux.F90
```diff
@@ -1231,7 +1231,10 @@ subroutine scale_fluxes (nx_block, ny_block, &
             fsens   (i,j) = fsens   (i,j) * ar
             flat    (i,j) = flat    (i,j) * ar
             fswabs  (i,j) = fswabs  (i,j) * ar
+            if (flwout(i,j) > -puny) & 
+               flwout  (i,j) = -stefan_boltzmann *(Tf(i,j) + Tffresh)**4
             flwout  (i,j) = flwout  (i,j) * ar
+            ! Special case where aice_init was zero and aice > 0.
```
eclare108213, line 1237: why is the comment 2 lines below the if statement?

dabail10: Whoops. Fixed now.

anton-seaice:
If its ok, ill defer to Tony and Elizabeth on this. ACCESS don't use this field.

dabail10:
> If its ok, ill defer to Tony and Elizabeth on this. ACCESS don't use this field.

Really? Was does the atmosphere component use for the upward longwave computation?

dabail10:
> Its calculated internally in the atmosphere - presumably in the same way CICE is calculating it

This probably is fine as long as the surface temperature is correct for the sea ice.

dabail10:
The shortwave fluxes and albedo are computed at the end of the step with the updated ice fraction. This is to be synchronized with the shortwave coming from the atmosphere on the next step. However, we could potentially revisit this. I'm doing it in scale_fluxes because it is where flwout is either divided by aice or it is set to the freezing point longwave. I agree there is a larger issue an I am happy to open one for this.

eclare108213:
flwout is initialized in suboutine init_coupler_fluxes [here](https://github.com/CICE-Consortium/CICE/blob/749e0742c0323cafe1723974b9e5ea4a6cc31bbe/cicecore/cicedyn/general/ice_flux.F90#L767). This is currently the freshwater value, intended over ice. Would it work to alternatively initialize it here for seawater Tf in an if block based on aice, rather than in scale_fluxes?

dabail10:
This is the whole issue with the bug. It is indeed initialized in init_coupler_fluxes, but then it is overwritten in init_flux_atm to be zero!

eclare108213:
Then why not initialize it in init_flux_atm?  Sorry I'm missing something here, maybe in the order that things are done in the various drivers.

dabail10:
It's an interesting question. The reason things are initialized in init_flux_atm as they are category merged quantities. So, flwout = flwout + sum(aicen(n)*flwoutn(n)). True also of fsens, flat, etc. However, if there is no ice at the beginning of the timestep then it is never computed and hence the issue with the bug. I tried this originally with something like:

where (aice(:,:,:) > 0) then
    flwout(:,:,:) = 0
endwhere

Then flwout would only be zeroed out when there was ice at the beginning of the step. This did not quite work as expected though. I don't understand why this was done in both init_flux_atm and init_flux_coupler. My main goal is to fix an actual bug where flwout = c0 when aice > 0. I am open to adding something more general later on.

eclare108213 (approved):
I'm approving this as a bug fix, but please create an issue referencing this PR, to thoroughly examine and document the coupling fields and steps.

apcraig (approved):

PR #1001, opened 2024-12-17
Enable triggering restart writes at any forecast time ("restart_fh") via UFS configuration as for other components

NickSzapiro-NOAA:

## PR checklist
    Enable writing forecast hour defined restarts ("restart_fh") as delimited floating point forecast hours in UFS model_configure component attributes
- [x] Developer(s): 
    Nick Szapiro, Denise Worthen, & Jun Wang
    Testing in UFS with cpld_control_gfsv17 confirms that restarts are bit-for-bit when (1) sharing common times with RESTART_N and (2) instead of RESTART_N. This feature is exercised in ufs-weather-model regression testing (see https://github.com/ufs-community/ufs-weather-model/pull/2419)
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No Docs for caps remain in caps

This PR enables writing forecast hour defined restarts ("restart_fh") in the same way as space-delimited floating point forecast hours in model_configure attributes, like `restart_fh: 0.25 2.5 6 17 24`. The implementation builds on [current restart_fh in MOM6](https://github.com/ufs-community/ufs-weather-model/pull/2099) for CMEPS, MOM6, CICE, and WW3 to have option of forecast hour restarts in addition to existing functionality. Restart writes are triggered when input restart_fh forecast hours are evenly divisible by a component's timestep (internally compared in units of integer seconds) and skipped otherwise. These restarts may be particularly helpful for data assimilation.

Enabling the functionality depends on UFS shared code (hence the #ifndef blocks), but this only depends on ESMF and could be adapted and shared with CESM as well, if CESM would like:
https://github.com/ufs-community/ufs-weather-model/blob/develop/CDEPS-interface/ufs/cdeps_share/shr_is_restart_fh_mod.F90


apcraig:
@dabail10, @anton-seaice, please have a quick look and provide feedback/approve as needed.

anton-seaice (approved):
No concerns from me.

Where is `shr_is_restart_fh_mod` defined ? Is it more adaptable to add an `#ifdef UFSCOUPLED` instead of relying on `#ifndef CESMCOUPLED` having this meaning?

NickSzapiro-NOAA:
Thanks @anton-seaice. Maybe the main goal is not to have many of these #ifdefs...certainly open to being more adaptable if it helps.

`shr_is_restart_fh_mod` is in UFS so that components don't each need to maintain separate copies (https://github.com/ufs-community/ufs-weather-model/blob/develop/CICE-interface/cice_files.cmake#L173)

dabail10 (approved):
This looks fine, but I think this is a bit of an issue where we have all of these different methods of reading / writing restarts depending on the coupled system.

DeniseWorthen:
Two comments here, just my .02....

The ``ifdef CESMCOUPLED`` is the ifdef we've tried to use across all component models which CESM and UFS share (WW3, CMEPS etc), so I think it makes more sense than creating a new UFSCOUPLED ifdef.

The issue of restart writing frequency is probably just going to be fundamentally different in a weather-forecasting setting than in a climate/research setting. The more flexible solution is to have this capability in defined by ESMF (through config), but for now that capability doesn't exist.

apcraig (approved):

PR #1005, opened 2025-01-31
Add ability to write out one-time grid history file

### LINKED ISSUE #613: Write grid variables only once.

apcraig:
## PR checklist
    Add ability to write out one-time grid history file
- [X] Developer(s): 
    apcraig
    Tests pass as expected, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#792aac351375090653303efe6280f8f8fc141db8
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Add ability to write one-time grid file in netCDf and PIO history features
    - Add grid_outfile logical namelist
    - Add extra history stream to be used just for grid output file, that stream is designated with the histfreq='g' label.  It's turned on with the logical grid_outfile namelist and then automatically turned off after it's written the first time.  This happens in ice_history.F90.
    - Add logic in ice_history_write.F90 to explicitly write the grid fields to the grid history file.
    - The one time grid filename is iceh_grid.nc
    - Add grid_outfile=.true. to set_nml.histall
    - Add external_variables global attribute to history files to note grid fields that are expected to be on other files
    - Add date_created global attribute
    
Fix bug in PIO where 5d variables had their two "z" axis coordinates reversed when writing the field even though the variable had the correct axes defined in the history file.  Detected by comparing netCDF and PIO history files.
    
Set history netcdf attribute Conventions to CF-1.8.  Ran history files through a CF checker.  This resulted in a couple of warning messages but no errors.  If grid variables associated with variable coordinate attributes are removed from the history file, this will result in CF errors even though external_variables are defined.  This is sort of a shortcoming in the current CF conventions, but we decided to move ahead anyway.  CF compliance will only be broken if grid variables are turned off in history files, and they are defined as part of other variable's coordinate (or similar) attributes.

Add axis to the history coord_attributes datatype to provide another attribute on the netcdf file.  Set axis to 'X', 'Y', or 'T' to a small subset of history file coordinate variables.
    
Change netcdf "unitless" unit to "1".
    
Update some abort calls in ice_history.F90
    
Update version_name namelist implementation, was not being broadcast or written to the log file.  Affected history file output.
    
Remove hist namelist settings in set_nml.run* files to minimize conflicts with other namelist options groups.
    
Add a test case that allows manual comparison of netCDF and PIO history files.  This needs to be done with cprnc.  cmp of restart files and diff of log files produce differences even when results are bit-for-bit identical.

Update documentation

Closes #613


eclare108213 (approved):
Nicely done.  My question is whether fixing the PIO axis bug will mess up anyone's post-processing tools.  Or have they not been plotting what they thought they were plotting?

apcraig:
It's hard to know how many users are looking at 5d output fields.  And this bug was only with PIO output which would probably impact only CESM users and a few other groups.   Hard to know how they were interpreting the output, but it was something that had to be fixed.  This fix probably has the smallest impact.  We can try to highlight this bug in our Release Notes.

apcraig:
Does anyone have a comment about the namelist name "grid_outfile".  Would "grid_histout" or "hist_gridout" or something else be better?

dabail10:
Sorry I haven't had a chance to look at this. I will try next week.

anton-seaice (commented):
Thanks @apcraig !

`grid_file` could be renamed to `grid_file_in` and then the output could be `grid_file_out` ?

My preferences for this change would be:
- grid output file name is configurable. It could be that `grid_file_out` is none for no output instead of `.false.` and otherwise the filename.
- no time dimension / variable in the grid output. Having a time variable in data that doesn't change is confusing for post processing tools. The model time can still be in the attributes. 
- raise an error if the user tries to use `grid_outfile = .true.` with binary io ?

Some other cf related stuff that might help:
- adding all the grid variables to the `external_variables` global attribute would clean a bunch or warnings in the history output when it's in a different file. see https://cfconventions.org/cf-conventions/cf-conventions.html#external-variables
- it looks like we could add the `standard_name` attribute to our output variables too - its unrelated to this change though. And that `date_created` as a global attribute would be more conventional than `history` 
- it also looks like CF prefers for the time bounds variable not to have 'long_name','units' or 'calendar' attributes. they are inherited from the 'time' variable. See http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#cell-boundaries

apcraig:
Thanks @anton-seaice, lots of good ideas.

I'm not sure we should have a string for grid_file_out.  That's not how the history mechanism works in general.  The namelists hist_file and hist_suffix with the frequency generates the history filename.  The grid history file does the same.

I will remove the time axis from the grid history file, that's a good idea.

I considered a number of other updates to the history files for CF including removing the bounds attributes.  I will also look into the external_variables attribute.  I'm not familiar with that one.  Should that only be set if the grid variables are NOT in the file?  Agree standard_name would be useful.  I'm a little afraid that I'll add names that are not standard though :).  I was inclined to leave the history files the same as much as possible, as long as CF errors weren't there.  But I'm happy to do some more work to clean up the CF warning messages and to add other attributes if that makes sense to others.  I worry a little about backwards compatibility (ie. if someone is looking for a certain attribute that we remove or rename), but maybe we should just move forward.   Thoughts?

cicecore/cicedyn/analysis/ice_history_shared.F90
```diff
@@ -822,12 +822,22 @@ subroutine construct_filename(ncfile,suffix,ns)
                  write(ncfile,'(a,a,i4.4,a,a)') &
                        history_file(1:lenstr(history_file))//trim(cstream),'.', &
                        iyear,'.',trim(suffix)
+              elseif (histfreq(ns) == 'g') then
+                 write(ncfile,'(a,a,a,a)')  &
+                       history_file(1:lenstr(history_file))//trim(cstream),'_grid', &
```
anton-seaice, line 827: I can't see how cstream in this case would be anything other than empty ?

anton-seaice:
> I'm not sure we should have a string for grid_file_out. That's not how the history mechanism works in general. The namelists hist_file and hist_suffix with the frequency generates the history filename. The grid history file does the same.

I guess i bring this up because I know that we will end up wanting to change the filename. There's configurable suffixes
for all the namelist frequencies.


> I worry a little about backwards compatibility (ie. if someone is looking for a certain attribute that we remove or rename), but maybe we should just move forward. Thoughts?

Maybe create a new issue to discuss and we can make changes in a later PR ?

cicecore/cicedyn/analysis/ice_history_shared.F90
apcraig: In theory, a user could set the 6th element of hist_suffix in namelist and it would be used here, but that feature is certainly not advertised nor expected to be used.  I will remove the "cstream" string from the grid history file so there is no chance of something odd happening here.  Good catch.

anton-seaice:
Apologies - I misunderstood the cf-docs about external_variables. (It looks like there is some similar discussion [here](https://github.com/cf-convention/cf-conventions/issues/357))

> * Do we proceed with a separate grid file, ability to turn off the grid variables in history files, and fail CF compliance formally?

My feeling is progress with:

https://github.com/cf-convention/cf-conventions/issues/357

MOM6 avoids this by just not providing a "coordinate" or "auxiliary coordinate" attribute. 

Its common to provide the grid in a seperate file I believe ? (e.g. some CMIP models do this I thought ?)

apcraig:
OK, I have updated the code.  The latest changes
- add a global external_variables attribute for grid variables that are not on the history files
- add date_created global attribute
- remove variable attributes for time and grid bounds variables.  these are inherited.
- remove some checks that were forcing grid variables to be on
- remove use of cstream in grid output file

I have confirmed that netcdf and pio files are generating the same data.  I have also passed some files thru CF compliance checkers with success.  However, CF compliance will fail if grid variables are removed from history files AND they are defined as coordinate or other variable attributes for history fields.  See https://github.com/cedadev/cf-checker/issues/116, https://github.com/cf-convention/cf-conventions/issues/357, and https://cfconventions.org/cf-conventions/cf-conventions.html#external-variables.

apcraig:
This should be ready for final review.  Let me know if anything else needs revision.

dabail10 (approved):

apcraig:
@anton-seaice, would you have a quick review before I merge.  Thanks!

cicecore/cicedyn/infrastructure/io/io_netcdf/ice_history_write.F90
```diff
@@ -436,6 +439,8 @@ subroutine ice_write_hist (ns)
                   call ice_check_nc(status, subname// ' ERROR: defining bounds for '//var_coord(i)%short_name, &
                                     file=__FILE__, line=__LINE__)
                endif
+            else
```
anton-seaice, line 442:
```suggestion
            else #grid_outfile
```

?

cicecore/cicedyn/infrastructure/io/io_netcdf/ice_history_write.F90
```diff
@@ -662,7 +675,16 @@ subroutine ice_write_hist (ns)
          call ice_check_nc(status, subname// ' ERROR: global attribute history', &
                            file=__FILE__, line=__LINE__)
 
-         status = nf90_put_att(ncid,nf90_global,'io_flavor','io_netcdf')
+         write(start_time,1001) current_date(1:4), current_date(5:6), &
+                                current_date(7:8), current_time(1:2), &
+                                current_time(3:4), current_time(5:8)
+1001     format(a,'-',a,'-',a,' ',a,':',a,':',a)
+
+         status = nf90_put_att(ncid,nf90_global,'date_created',start_time)
```
anton-seaice, line 683: Do we need to clarify this is model date not real world ?

anton-seaice (approved):
Looks good @apcraig - I didn't rerun but the changes look good by code review.

A couple of very minor comments

cicecore/cicedyn/infrastructure/io/io_netcdf/ice_history_write.F90
apcraig: This is the real world date, the variables come from a call to "date_and_time".

apcraig:
Thanks @anton-seaice.  Will merge now.

PR #1006, opened 2025-02-05
Update rectangular (box) configurations

### LINKED ISSUE #1004: Rectangular grid problems
As a part of some other debugging, I've identified a number of issues with the rectangular grid setup.  For a default 128x128 box, we get 30km grid spacing that wraps over the top of the pole with lower left corner at -156 degE, 71 degN.  ULAT values go over 90 with units degN which makes no sense.  In addition, the i=1 TLON and TLAT values are incorrect.  I think when the box implementation was updated in 2022, it may have only been tested on the 12x12 box configuration.  And nobody has looked at the details since.

I doubt this is impacting our results much if at all for our testing, I think functionally just the LON and LAT assignment is bad.  I guess this might come into play for some physics, forcing, and diagnostics, but may not really the box testing in practice.

If anyone has any suggestions about the location and grid spacing for our box grids, please offer it.  The box configurations we have are 180, 128, 80, and 12.  As a starting point, I'll probably just update the box grids so we use the same lower left reference LON/LAT and change the grid spacing so each grid covers the same basic area.  Other suggestions welcome.  I will also verify all the grid fields as part of this effort.

apcraig:
## PR checklist
    Update rectangular (box) configurations
- [X] Developer(s): 
    apcraig
    Several box tests change answers, as expected.  This fixes a few bugs in the box setup.  Non-box configurations pass and are bit-for-bit as expected.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#c3429c9a8f34869800e2ef8d06a2096be89d1d0e
    - [ ] bit for bit
    - [X] more substantial, but just for some box configurations 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Update the rectangular (box) grid configurations.
- Fix bug where left hand side of rectangular grids had incorrect longitudes and latitudes computed.  This changes answers in some box configurations.
- Update the box resolutions.  dxrect and dyrect were hardwired to 30km for all box resolutions.  Now adjust dxrect and dyrect depending on the number of gridcells so it always covers an area of about 3 degrees by 3 degrees.  This means gbox12 is 30km resolution, gbox80 is 5km, gbox128 is 3 km, and gbox180 is 2 km.  Reduce the gbox180 timestep to 1800s for stability.  This changes answers for some box configurations.
- Remove u masking from box2001 ocean forcing, bit-for-bit.
- Remove 50W check associated with box2001 initial conditions, bit-for-bit.
- Add rectangular grid diagnostics
- Minor update to the box2001 documentation
    
Explicitly set npt_unit in all set_nml files that have npt set.  Improves robustness where one option sets npt and another sets npt_unit.

These changes do not fundamentally change how the model behaves with simple box test configurations.  This just fixes a few inconsistencies in how the box grids are setup.

Closes #1004

### DIFF
--- a/cicecore/cicedyn/general/ice_forcing.F90
+++ b/cicecore/cicedyn/general/ice_forcing.F90
@@ -5202,7 +5202,6 @@
       use ice_domain, only: nblocks, blocks_ice
       use ice_blocks, only: block, get_block, nx_block, ny_block, nghost
       use ice_flux, only: uocn, vocn
-      use ice_grid, only: uvm
 
       ! local parameters
 
@@ -5234,9 +5233,6 @@
                             / real(ny_global,kind=dbl_kind) - p1
          vocn(i,j,iblk) = -p2*real(iglob(i), kind=dbl_kind) &
                             / real(nx_global,kind=dbl_kind) + p1
-
-         uocn(i,j,iblk) = uocn(i,j,iblk) * uvm(i,j,iblk)
-         vocn(i,j,iblk) = vocn(i,j,iblk) * uvm(i,j,iblk)
 
          enddo
          enddo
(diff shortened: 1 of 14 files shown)


dabail10 (approved):
This looks good. Thanks for chasing this down.

eclare108213 (approved):

Issue #1044, opened 2025-07-31
missingvalue in netcdf reads


anton-seaice:
In [ice_read_write.F90](https://github.com/CICE-Consortium/CICE/blob/227862610a87c6595de7ed6ea7583be228ca320b/cicecore/cicedyn/infrastructure/ice_read_write.F90#L1221), there are a few times where `_FillValue` is read from netcdf files.

If _FillValue is not set, this returns a nan, which causes `floating invalid` errors when doing _Debug_ builds (e.g. `-fpe0` set in intel compiler flags).

I suggest we put explicit handling in for this, e.g.:

```
         if (isnan(missingvalue)) then
            !  _FillValue not defined for resetting to spval_dbl
            missingvalue =  spval_dbl
         endif
```

 _FillValue seems to only be used in some logging statements, so i don't there's any impact on the model as such?

PR #1045, opened 2025-08-04
Debug failures

anton-seaice:
This PR address #1044 

There are a couple of spots in the code when loading files which fail when building with checks for floating point exceptions (`-fpe0` in intel land). This addresses those issues

## PR checklist
       Resolve issues with nans / accessing arrays out of range
- [ ] Developer(s): 
@anton-seaice 
@apcraig 
Adhoc testing, these changes don't impact any existing tests
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [ ] Yes
        - [x] No 

This change max sure that CICE is not trying to use array indeces which don't have values associated with them, and puts a value in _FillValue when it's not set (to avoid a nan). When creating ULAT, don't create the `work_mom` array on tasks other than master task because it is not used. When loading `_FillValue`, reset the value to `spval_dbl` if `_FillValue` is a nan.

### DIFF
--- a/cicecore/cicedyn/infrastructure/ice_grid.F90
+++ b/cicecore/cicedyn/infrastructure/ice_grid.F90
@@ -369,27 +369,26 @@
 
                if (my_task == master_task) then
                   allocate(work_mom(nx_global*2+1, ny_global*2+1), stat=ierr)
-               else
-                  allocate(work_mom(1, 1), stat=ierr)
-               endif
-               if (ierr/=0) call abort_ice(subname//' ERROR: Out of memory', file=__FILE__, line=__LINE__)
+                  if (ierr/=0) call abort_ice(subname//' ERROR: Out of memory', file=__FILE__, line=__LINE__)
 
-               fieldname='y'                ! use mom y field to fill cice ULAT
-               call ice_open_nc(grid_file,fid_grid)
-               call ice_read_global_nc(fid_grid,1,fieldname,work_mom,.true.)
-               call ice_close_nc(fid_grid)
-               im = 3
-               do i = 1, nx_global
-                  jm = 3
-                  do j = 1, ny_global
-                     work_g1(i,j) = work_mom(im, jm)
-                     jm = jm + 2
+                  fieldname='y'                ! use mom y field to fill cice ULAT 
+                  call ice_open_nc(grid_file,fid_grid)
+                  call ice_read_global_nc(fid_grid,1,fieldname,work_mom,.true.)
+                  call ice_close_nc(fid_grid)
+                  im = 3
+                  do i = 1, nx_global
+                      jm = 3
+                      do j = 1, ny_global
+                         work_g1(i,j) = work_mom(im, jm)
+                         jm = jm + 2
+                      enddo
+                      im = im + 2 
                   enddo
-                  im = im + 2
-               enddo
 
-               deallocate(work_mom, stat=ierr)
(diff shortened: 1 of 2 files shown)


apcraig:
Interesting, I've been working around this by setting -fpe1.  I understood this to be a bug in an hdf library that is getting fixed.  Are these workarounds something we should be doing (ie. something we should fix) or just a workaround for the errors in the latest versions of the libraries?

anton-seaice:
Oh - so 

`nf90_get_att(fid, varid, "_FillValue", missingvalue)` won't return a nan if build without hdf5 ?

Is there an issue in the hdf5 repo ?

My interpretation of this was that using a NaN for _FillValue was valid, but that `work_g1 /= missingvalue` gave a runtime error when comparing to a nan?

Probably a better fix is to use nan for the mask 

e.g. 

```
         if (isnan(missingvalue)) then
            amax = maxval(work_g1, mask = .not. ieee_is_nan(work_g1) )
         else
            amax = maxval(work_g1, mask = work_g1 /= missingvalue)
         endif
!         write(nu_diag,*) subname,' missingvalue= ',missingvalue
         amin = minval(work_g1)
```

Or i guess we could skip the diagnostic printing when _FillValue is a nan?

apcraig:
If we're talking about the same thing, my understanding is that an update in hdf5 (1.14.3) introduced some bugs.  I have had to update several machine ports to switch -fpe0 to -fpe1 for them to continue to work.  I have also had to turn off -check uninit in some inteloneapi builds, that's a separate issue.

https://github.com/HDFGroup/hdf5/issues/3831

Google "hdf5 float trap bug", the AI info seems to provide a reasonable summary.

If this PR fixes these problems, maybe that's a good thing and maybe it's something we should have in our code to make it more robust.  On the other hand, if our code is OK and we can wait for the hdf5 update and work around with compiler flags in the short term, maybe we should do that.  Thoughts?

anton-seaice:
Oh that is a different issue :)

The issue here is if grid/bathymetry netcdf files use `Nan` as the _FillValue, when CICE tries to mask using the Nan fillvalue it gives a floating point exception runtime error.

apcraig:
I tested these changes on derecho with 4 compilers.  The nvhpc compiler fails to build most cases with the following error,

NVFORTRAN-S-0038-Symbol, isnan, has not been explicitly declared (/glade/work/tcraig/cice-consortium/cice.anton.debugf/cicecore/cicedyn/infrastructure/ice_read_write.F90)
  0 inform,   0 warnings,   1 severes, 0 fatal for ice_read_nc_xy

The gnu compiler is also unhappy when building with the debug flags,

/glade/work/tcraig/cice-consortium/cice.anton.debugf/cicecore/cicedyn/infrastructure/ice_read_write.F90:1604:13:
 1604 |          if (isnan(missingvalue)) then
Error: Function 'isnan' at (1) has no IMPLICIT type

It looks like isnan may not be a standard Fortran intrinsic.  There does seem to be an "ieee_is_nan" function in the standards, maybe?  I'm not totally clear on this.

apcraig:
I've reviewed the code mods, tested on derecho with quick_suite, io_suite, unittest_suite, and base_suite using intel, gnu, cray, and nvhpc.  I've also run the quick_suite on the nag compiler and everything seems to work well.  We'll merge this week unless there are any other thoughts.

apcraig (approved):

anton-seaice:
Thanks @apcraig - sounds good

Issue #1038, opened 2025-07-18
SIMIP history output implementation


eclare108213:
I believe several of the SIMIP history variables are not implemented according to the SIMIP definition, e.g. **sitemptop**, **sitempsnic** and **sitempbot**.  Other variables might have similar issues and all should be reviewed.  This issue arose from the discussion in https://github.com/CICE-Consortium/CICE/issues/1033.

There are several different problems with the implementation of these variables:
1. misplaced parentheses
2. inappropriate scaling by `aice` and/or `aice_init`
3. using fields that aren't aggregated across categories
4. using fields that incorporate nonzero open water values

Regarding the ice area scaling:

A fundamental concept in the model is that the thermodynamics calculations are for a single point on an ice category and assume that the same value is assigned to every other point on the ice in that category, i.e. the value is constant across the category.  If the area changes, the vertical temperature profile stays the same -- there is no associated area scaling for state variables!  For fluxes, there is an area scaling, but not for state variables like temperature.

Averaging across categories is a slightly different concept.

The [SIMIP document](https://gmd.copernicus.org/articles/9/3427/2016/) clearly states that it wants quantities averaged over only the ice area, not the grid cell area, e.g.

> … we specifically request all fluxes over the ice-covered part of the grid cell.
> ...
> For all variables that are not proportional to area fraction, i.e. intensive variables such as albedo, temperature, or heat flux, SIMIP requests the area-weighted average. Hence, all time samples with non-zero sea-ice fraction are first multiplied by area fraction, then summed, and then divided by the sum of the area fractions.
> ...
> C1.8 Surface temperature (sitemptop)
> Mean surface temperature of the sea-ice covered part of the grid cell. 

When tracers are aggregated and averaged in `icepack_aggregate` and `compute_new_tracers`, the pseudocode is
```
aice = sum(aicen)
if (aice > puny) then
    trcr = [sum (trcrn * aicen)] / aice
else
    trcr = 0
    if (tracer = Tsfc) trcr = Tf
endif
```
where the sums are for `n=1,ncat`.  This average is over the ice, not over the grid cell.  Multiplying by `aice` produces the grid-cell average IF the quantity being averaged is zero for open water.

If you want to average using `aice_init` instead of `aicen` because the ice area was `aice_init` during the thermo calculations, then you would have to recalculate the whole equation, not just scale the end result by `aice/aice_init`:
```
    aice_init = sum(aicen_init)
[...]

Issue #1033, opened 2025-05-23
aice vs. aice/aice_init factor in ice_history


NickSzapiro-NOAA:
In coupled UFS (GFS/GEFS/SFS/RTOFS test runs), EMC folks are noticing large negative values in sitempbot (like <-1000 K) in ice margins like this quick ncview from cpld_control_gfsv17 regression test (note -300<T<300 K colorbar):
![Image](https://github.com/user-attachments/assets/9e0a7f81-a311-4d33-9600-95c3bbe55de9)

As sitempbot is diagnostic based on Tbot, one suspect is the accumulation in ice_history where sitemptop, sitempsnic, and sitempbot are all treated differently. In particular the aice/aice_init factor, which I imagine relates to calculating variables at beginning vs. end of timestep. I tried a quick fix of: 
```
diff --git a/cicecore/cicedyn/analysis/ice_history.F90 b/cicecore/cicedyn/analysis/ice_history.F90
index a0313c5..8f5c5ac 100644
--- a/cicecore/cicedyn/analysis/ice_history.F90
+++ b/cicecore/cicedyn/analysis/ice_history.F90
@@ -2775,8 +2773,8 @@ subroutine accum_hist (dt)
            worka(:,:) = c0
            do j = jlo, jhi
            do i = ilo, ihi
-              if (aice_init(i,j,iblk) > puny) &
-                 worka(i,j) = aice(i,j,iblk)*(Tbot(i,j,iblk)/aice_init(i,j,iblk)+Tffresh)
+              if (aice(i,j,iblk) > puny) &
+                 worka(i,j) = aice(i,j,iblk)*(Tbot(i,j,iblk)+Tffresh)
            enddo
            enddo
            call accum_hist_field(n_sitempbot, iblk, worka(:,:), a2D)
```

This fixes the sitempbot field (note 271<T<273 K colorbar): 
![Image](https://github.com/user-attachments/assets/ee3bd0fb-b719-45de-9370-e0da232e879b)

There is a similar issue with sitempsnic and Tsnice.
 
Apologies if my figure artistry is less than clear...happy to share the actual output if it helps.

As it looks like the current implementation involves beginning vs. end of timestep variables, is removing the `1/aice_init(i,j,iblk)` right?

dabail10:
[...]

PR #1054, opened 2025-09-24
Fixes for sitemptop, sitempbot, and sitempsnic.

dabail10:

## PR checklist
This is the first in a series of fixes for SIMIP variables. For now, it is just the sitemp variables.
- [X] Developer(s): 
 dabail10 (D. Bailey)
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#ae5b8fd81f597724f1d804902cbb7df56f83f725
    - [X] bit for bit except gnu and cray due to changes in Icepack
    - [ ] more substantial 
    - [X] Yes
    - [ ] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

This PR will also require an icepack commit later on. For now, this is the changes for CICE only. Start of addressing issue #1038 and #1033 

Changes in CICE:

- move conversion to K into the define_hist_field call
- fix the accumulation of sitemptop, sitempsnic, and sitempbot to reflect the "intrinsic" versus "extrinsic" definitions.
- redo the avg_ice_present division by aice, so that conb is added at the end
- also undo the ravgct multiplication and division for the case of avg_ice_present
- add cell_methods as "area: time: mean where sea ice (mask=siconc)"
- Update icepack hash to bug fix version of "Tsnice".

### DIFF
--- a/cicecore/cicedyn/analysis/ice_history.F90
+++ b/cicecore/cicedyn/analysis/ice_history.F90
@@ -3848,7 +3848,7 @@
                           a3Dc(i,j,k,n,iblk) = spval_dbl
                        else                            ! convert units
                           a3Dc(i,j,k,n,iblk) = avail_hist_fields(nn)%cona*a3Dc(i,j,k,n,iblk) &
-                                         * ravgipn(i,j,n) + avail_hist_fields(nn)%conb
+                                         * ravgipn(i,j,k) + avail_hist_fields(nn)%conb
                        endif
                     enddo             ! i
                     enddo             ! j


NickSzapiro-NOAA (approved):

anton-seaice (commented):
This looks ok - but https://github.com/CICE-Consortium/Icepack/pull/542 has changed Tsnice (in Celsius) to already be aice weighted, so in combination with this change , it would get weighted by aice twice (sortof)

dabail10:
> The category and time averaging are confusing enough for everyone (including me) that I think it's worth writing it out mathematically and making that available in the docs. I'll work on that.
> 
> My question for @dabail10 is whether the CMIP data request asks for Tsnice over the grid cell or over just the ice. I think it's the latter, and the confusing part is why you're only computing a grid-cell value in Icepack.
> 
> I think of the time averaging this way:
> 
> Average volume `V_avg = sum(V) / M`, where the sum is over all time steps in the averaging period and `M` is the number of time steps with ice present. Average content, e.g. `salt_avg = sum(V*S) / M` where `S` is salinity i.e. salt tracer Average tracer, e.g. `salinity_avg = sum(V*S) / sum(V) = salt_avg / V_avg`
> 
> The quantities being averaged in time here have already been integrated or averaged over categories, and it's always better to use 'content' quantities for conservation and/or accuracy. I think that's what @dabail10 is doing but I want to write it out with units to check. The temperatures are surface quantities so the integrals would be over the surfaces rather than the volumes.
> 
> Let me know if you see errors in my thinking here!

The SIMIP request only asks for the temperature within / over / under the ice. However, you point is valid in terms of the definition of intensive versus extensive. I think the point is we want to avoid averaging in zeroes that are not physical. So, the aggregate Tsnice is:

```
Tsnice = sum(aicen(n)*Tisnicen(n)), for n=1,5
```

From the Notz et al. 2016 paper:

"For all variables that are proportional to area fraction, i.e. extensive variables such as volume, mass, or area fraction, a zero should always be averaged in for all time steps where no sea ice is present. This is because the extensive variables naturally approach zero as area fraction approaches zero."

So in this sense, this definition of Tsnice is indeed extensive. It becomes intensive if we do:

Tsnice_cell = sum(aicen(n)*Tsnicen(n)) / aice(n), n=1,5

this is a weighted mean of Tsnicen, again only over the ice. However, it does not go to zero as aice goes to zero. So, I agree that I should just do a straight time average of Tsnice as it is defined. This is similar for trcr(:,:,nt_Tsfc) and Tbot. For Tbot though since it is all the same for each category, then Tbot_cell = Tbot.

dabail10:
I just tested the this with direct time averaging for sitemptop, sitempbot, and sitempsnic. I am getting some sitempsnic values above 0C which doesn't make sense to me. Does ktherm=2 ever allow internal snow and ice temperatures above 0C?

dabail10:
@eclare108213 @anton-seaice @NickSzapiro-NOAA Hopefully I have addressed your concerns on this as well.

cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -1536,17 +1536,17 @@ subroutine init_hist (dt)
          call define_hist_field(n_sitemptop,"sitemptop","K",tstr2D, tcstr,    &
              "sea ice surface temperature", &
              "none", c1, c0,           &
-             ns1, f_sitemptop, avg_ice_present=.true., mask_ice_free_points=.true.)
+             ns1, f_sitemptop, avg_ice_present=.false., mask_ice_free_points=.true.)
```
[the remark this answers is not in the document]
anton-seaice: Nevermind, tmask is applied seperately

anton-seaice:
Sorry @dabail10 - I am still confused, well probably need to have a meeting !

In my very simply example, with one ice thickness category:


If ncat = 1 and aicen = aice = 0.5
And there is no snow
And surface temperature is 2°C

Then icepack produces a grid cell mean temperature (Tsnice) of 1°C (as sum(aicen*Tsf))

If we assume that there is no change during a day, then this implementation calculates the sitemptop daily average as 274.15K (for either setting of `avg_ice_present`)

But the correct answer is 275.15K ? (As its a diagnostic of surface temperature over sea ice)

dabail10:
Ok. I am happy with the averaging when ice present for sitemptop, sitempsnic, and sitempbot. I also added the updated cell_methods for variables when avg_ice_present = .true. I will work on the aice, aice_mid, aice_init stuff in the next PR.

cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -2764,11 +2763,7 @@ subroutine accum_hist (dt)
...
            do i = ilo, ihi
-              if (vsno(i,j,iblk) > puny .and. aice_init(i,j,iblk) > puny) then
-                 worka(i,j) = aice(i,j,iblk)*(Tsnice(i,j,iblk)/aice_init(i,j,iblk)+Tffresh)
-              else
-                 worka(i,j) = aice(i,j,iblk)*(trcr(i,j,nt_Tsfc,iblk)+Tffresh)
-              endif
+              worka(i,j) = Tsnice(i,j,iblk)
            enddo
            enddo
            call accum_hist_field(n_sitempsnic, iblk, worka(:,:), a2D)
```
anton-seaice:
```suggestion
if (f_sitempsnic(1:1) /= 'x') &
call accum_hist_field(n_sitempsnic, iblk, Tsnice(:,:,iblk), a2D)
```

although this might get undone again when averaging by aice_mid ?

is it useful to add a comment here,

e.g. sitempsnic is intensive, and Tsnice is already weighted by aice in icepack

cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -3705,31 +3699,31 @@ subroutine accum_hist (dt)
...
+              ! Only average when/where ice present
               if (avail_hist_fields(n)%avg_ice_present) then
                  do j = jlo, jhi
                  do i = ilo, ihi
                     if (tmask(i,j,iblk)) then
-                          a2D(i,j,n,iblk) = &
-                          a2D(i,j,n,iblk)*avgct(ns)*ravgip(i,j)
+                       a2D(i,j,n,iblk) = avail_hist_fields(n)%cona*a2D(i,j,n,iblk) &
+                                      * ravgip(i,j) + avail_hist_fields(n)%conb
                     endif
```
anton-seaice: Does there need to an else case here?

e.g 

```
else
    a2D(i,j,n,iblk) = spval_dbl

File: cicecore/cicedyn/analysis/ice_history.F90
Comment (anton-seaice) [inline cicecore/cicedyn/analysis/ice_history.F90]: This section needs re-ordering the match the changes to the 2d vars

File: cicecore/cicedyn/infrastructure/io/io_netcdf/ice_history_write.F90 [hunk 4]
```diff
@@ -1287,9 +1287,15 @@ subroutine ice_hist_field_def(ncid, hfield, lprecision, dimids, ns)
...
            .and.TRIM(hfield%vname(1:4))/='sigP') then
-             status = nf90_put_att(ncid,varid,'cell_methods','time: mean')
-             call ice_check_nc(status, subname// ' ERROR: defining cell methods for '//hfield%vname, &
-                               file=__FILE__, line=__LINE__)
+             if (hfield%avg_ice_present) then
+                status = nf90_put_att(ncid,varid,'cell_methods','area: time: mean where sea ice (mask=siconc)')
+                call ice_check_nc(status, subname// ' ERROR: defining cell methods for '//hfield%vname, &
+                                  file=__FILE__, line=__LINE__)
+             else
+                status = nf90_put_att(ncid,varid,'cell_methods','time: mean')
```
anton-seaice:
```
status = nf90_put_att(ncid,varid,'cell_methods','time: mean')
```

so in theory we could be adding `area:` method here to? (for the case where `avg_ice_present` is .false.)

but we'd need a way to distinguish between diagnostics using ice area vs grid cell area, so i don't think is easily possible at this point ?

cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -3705,31 +3699,31 @@ subroutine accum_hist (dt)
...
+              else
+                 do j = jlo, jhi
+                 do i = ilo, ihi
+                    if (.not. tmask(i,j,iblk)) then ! mask out land points
+                       a2D(i,j,n,iblk) = spval_dbl
+                    else                            ! convert units
+                       a2D(i,j,n,iblk) = avail_hist_fields(n)%cona*a2D(i,j,n,iblk) &
+                                      * ravgct + avail_hist_fields(n)%conb
+                    endif
+                 enddo             ! i
```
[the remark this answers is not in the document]
anton-seaice: Possibly this suggestion would work fine without the just `:` instead of `ilo:ihi` (both forms seem to be used in this file)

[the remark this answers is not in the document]
dabail10: Great question. Not sure why. I'm going to leave it for now.

cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -2778,8 +2774,7 @@ subroutine accum_hist (dt)
            worka(:,:) = c0
            do j = jlo, jhi
            do i = ilo, ihi
-              if (aice_init(i,j,iblk) > puny) &
-                 worka(i,j) = aice(i,j,iblk)*(Tbot(i,j,iblk)/aice_init(i,j,iblk)+Tffresh)
+               worka(i,j) = aice(i,j,iblk)*Tbot(i,j,iblk)
```
apcraig: could shift this line 1 character left

apcraig (changes requested):
Why is Icepack updated in this PR?   The PR says Icepack is not being updated.  Does this PR need the latest Icepack and/or https://github.com/CICE-Consortium/Icepack/pull/542?

Are there any changes needed for io_binary (noticed only io_netcdf and io_pio2) were updated?

I have not checked the science of the averaging, just the technical details.

dabail10:
We need the Icepack fix for the computation of Tsnice. I assumed this woul come in later when the Icepack PR has been merged. Am I missing something?

cicecore/cicedyn/analysis/ice_history.F90
dabail10: This is just metadata in the netcdf files. We don't have this for binary files?

apcraig:
> We need the Icepack fix for the computation of Tsnice. I assumed this woul come in later when the Icepack PR has been merged. Am I missing something?

That's fine, so we do need to merge the Icepack PR first.  Then you will want to update this PR, right?

dabail10:
There is still the outstanding issue of using aice, aice_init, or aice_mid for averaging. That will be in the next PR.

cicecore/cicedyn/analysis/ice_history.F90
dabail10: > In this section of code, the Tsfc tracer and Tbot are multiplied by aice, but Tsnice is not. I think it would be less confusing if all three were sent out of Icepack in the same form. I know they are computed differently in Icepack, but shouldn't the end result that is sent to the driver be something like "a surface temperature (top, ice-snow interface, or bottom) that is averaged only over the ice that's present" for all three? At the moment it looks like Tsfc and Tbot are averaged over the ice and Tsnice is averaged over the grid cell when they are sent out of Icepack.

I feel like I am going around in circles on this one. The idea was that we did not want Tsnice as:

```
Tsnice = sum(aicen(n)*Tsnicen(n)) / sum(aicen(n))
```

and then multiply by aice again in the history averaging. So, Icepack is just sending:

```
Tsnice = sum(aicen(n)*Tsnicen(n))
```

thus we do not need to mutilply again by aice when accumulating in ice_history.F90. However Tsfc and Tbot do require multiplication by aice.

eclare108213:
The associated Icepack PR has been merged.

apcraig:
My testing reveals some issues.  I did

./cice.setup --suite quick_suite,base_suite,io_suite,unittest_suite -m derecho -e intel,gnu --testid dh01 --bgen cice.dh01 --bcmp cice.bd30d63f31.251101

Several tests are failing.  For instance,

derecho_intel_restart_gx3_32x1_debug_histall_iocdf1_ionetcdf_precision8

```
forrtl: severe (408): fort: (2): Subscript #3 of the array RAVGIPN has value 12 which is greater than the upper bound of 5

cice               0000000001130B91  ice_history_mp_ac        3851  ice_history.F90
cice               0000000000415D9D  cice_initmod_mp_c         199  CICE_InitMod.F90
cice               0000000000413F33  cice_initmod_mp_c          54  CICE_InitMod.F90
cice               0000000000413D2B  MAIN__                     43  CICE.F90
cice               0000000000413CFD  Unknown               Unknown  Unknown
libc-2.31.so       0000146CEBED829D  __libc_start_main     Unknown  Unknown
cice               0000000000413C2A  Unknown               Unknown  Unknown

```

You can see the results here if it's helpful,

  derecho:  /glade/work/tcraig/cice-consortium/cice.db.history/testsuite.dh01

cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -3838,30 +3840,33 @@ subroutine accum_hist (dt)
...
-                                   * ravgct + avail_hist_fields(nn)%conb
+                 if (avail_hist_fields(nn)%avg_ice_present) then
+                    do k = 1, ncat_hist
+                    do j = jlo, jhi
+                    do i = ilo, ihi
+                       if (.not. tmask(i,j,iblk)) then ! mask out land points
+                          a3Dc(i,j,k,n,iblk) = spval_dbl
+                       else                            ! convert units
+                          a3Dc(i,j,k,n,iblk) = avail_hist_fields(nn)%cona*a3Dc(i,j,k,n,iblk) &
+                                         * ravgipn(i,j,n) + avail_hist_fields(nn)%conb
```
anton-seaice:
```suggestion
                                         * ravgipn(i,j,k) + avail_hist_fields(nn)%conb
```

I think there is a typo here @dabail10

dabail10: Good catch. Fixed.

dabail10:
Running the quick suite on izumi.

dabail10:
I am getting answer changes on izumi. It is in the restart files, but these should not be changed. I wonder if it is that I don't have read permission?

ncdiff: ERROR User does not have read permission for /scratch/cluster/tcraig/CICE_BASELINES_MASTER/cice.bd30d63f31.251101-120105/izumi_gnu_smoke_gx3_4x1_debug_diag1_run2day_thread/restart/iced.2005-01-03-00000.nc

dabail10:
> > I wasn't seeing failed tests. I will look into this though. Derecho is down today and tomorrow for maintenance.
> 
> A full test suite on derecho is something like
> 
> `--suite first_suite,base_suite,travis_suite,decomp_suite,reprosum_suite,unittest_suite,omp_suite,gridsys_suite,perf_suite,io_suite,quick_suite `
> 
> As noted above, I ran
> 
> `--suite quick_suite,base_suite,io_suite,unittest_suite `
> 
> The io suite does things like turn on all history variables; test all flavors of netcdf, pio, and binary; and tests several history options. This suite is pretty useful when modifying history stuff. FYI, currently pio support in standalone mode is only available on derecho, so other machines will fail the pio tests. So the io_suite is only run on derecho at this point.

Ah ok. I will rerun when derecho comes back up.

dabail10:
I have rerun the io_suite on derecho compared to the baselines. Everything is good except for the following test. Is this a known issue?

PEND derecho_intel_restart_gx3_32x1_alt02_histall_iohdf5_iopio1 run
PEND derecho_intel_restart_gx3_16x2_alt06_histall_iohdf5_iohdf5opts_iopio2 run
PEND derecho_intel_restart_gx3_8x4_histall_iohdf5_iopio2_isotope_precision8 run
FAIL derecho_intel_restart_gx3_30x1_bgcz_histall_iohdf5_iopio1_precision8 run

 (ice_pio_init) create file ./history/iceh_ic.2005-01-01-00000.nc
 NetCDF: Problem with HDF5 dimscales.
 NetCDF: Problem with HDF5 dimscales.

 (abort_ice)ABORTED:
 (abort_ice) called from ice_history_write.F90
 (abort_ice) line number          689
 (abort_ice) error =
 (ice_pio_check), (ice_write_hist) ERROR: ending pio definitions
 NetCDF: Problem with HDF5 dimscales.

apcraig:
> I think everything is committed now.

The last push seems to be from 3 days ago if I'm looking push history above correctly, you might check whether you need to push to the repo.

apcraig:
The nvhpc compiler issues are unrelated to this PR.  I tested last weeks Icepack main which worked last week and now it's not running.  I will reach out to CISL and update the port in a separate PR if needed.  We can ignore the nvhpc issues for now.

dabail10:
> It's a little concerning to me that the answers change even without any optimization. The Icepack changes are minimal -- I can see why Tsnice itself would not be BFB, but it's only diagnostic in both Icepack and CICE, so shouldn't be changing the BFBness in the test suites. Is this a problem with the code, or is it a problem with the gnu and cray compilers? Since intel is BFB, I'd suspect the latter, but at -O0?

I have rechecked the code and the only changes are related to sitemptop, sitempsnic, and sitempbot which are all diagnostic. There was a derecho update last week that looks to have changed answers.

apcraig (approved):

anton-seaice:
> @anton-seaice Have your questions and concerns above been adequately addressed? Some of them were postponed to the next PR. I mainly want to make sure that the answers for all three variables now reflect what you'd expect as a CMIP/SIMIP user.

I think what this change does is ok by the definition:

`sitempbot`, `sitemptop` and `sitempsnic` are defined in Kelvin, and are being calculated as a mean weighted by sea ice area fraction. Because the output is by aice>puny, users can be reasonably confident its a area-weighted mean. Cell_methods are defined correctly for these variables (although they seem to be still in a state of flux in a general sense in the data request, so it's probably hard to get them correct straight out of the model).

I guess users may want to know that sitempbot and sitempsnic are not tracers - so are approximate?

In a general sense, applying consistent notes and cell methods to the output would help but its not really specific to this change. (e.g. the metadata for all variables could state what time averaging & spatial averaging was applied). Similarly - internally labelling variables consistently based on their area fraction would help developers a lot (but also unrelated to this change) (e.g. Tsnice is a grid cell aggregate and Tbot is an true temperature). The other unresolved part is obviously applying aice weightings consistent - Tsnice/Tbot are from the middle of the timestep, but then the area weighting in time uses aice at the end of the timestep.

Anyway - I don't think i've said anything new - happy to merge!

eclare108213 (approved):
This work has highlighted how much more there is to do, to improve the history output as well as more general conservation issues.
@dabail10 thanks for the huge effort to get these issues fixed and for your patience with the rest of us.
@anton-seaice and @nick thanks for your thoughtful reviews.
@apcraig thanks for your extensive testing.
There will likely be more refactoring, but for now I think this PR is ready to merge.

PR #1055, opened 2025-09-25
add Icepack minimum area and mass parameters

eclare108213:
Initializes Icepack's minimum area and mass parameters to dynamics values
- [x] Developer(s): 
@eclare108213 @DeniseWorthen 
    ENTER INFORMATION HERE 
    - [ ] bit for bit
    - [x] more substantial 
    - [x] Yes
    - [ ] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [x] Yes
        - [ ] Yes
        - [ ] No 
Addresses https://github.com/CICE-Consortium/CICE/issues/645. The required Icepack changes are in https://github.com/CICE-Consortium/Icepack/pull/543.

Removes residual amounts of ice that are not otherwise handled by the numerics.  The controlling parameters (`itd_area_min` and `itd_mass_min`, implemented in Icepack) set minimum ice area and mass values below which all ice is removed following the thermodynamics and ridging calculations. For the B-grid, these parameters are currently set to the dynamics stability minima, which are being reduced to extremely small values based on testing in multiple modeling systems. If needed, users can revert these parameters to the original, larger values by adding them to ice_in. Setting them to 0 turns off the new zapping completely. These parameters are set to the larger, original values in the C-grid test scripts, pending further work.

Closed due to conflicts.  The replacement PR is https://github.com/CICE-Consortium/CICE/pull/1067.


apcraig:
So zap_residual_ice = .false. preserves the current icepack zapping implementation?

It seems like there are several considerations.
- Do we intend to preserve use of the zap_residual logical as well as the independent min_area and min_mass parameters indefinitely?  Or is the ultimate goal to have zap_residual always on and min_area=dyn_area_min and min_mass=dyn_mass_min?
- Are there multiple zap features in dynamics or thermodynamics that we might want to control?
- If we keep the new parameters, do we want to have a mode where a user can specify min_area=dyn_area_min and min_mass=dyn_mass_min without having to set all 4 parameters manually.
- Is turning off the zap feature in dynamics a valid option?
- Can any or all of the 4 area/mass parameters be set to zero and produce a result (even if not ideal)?

And some other feedback.
- Would it make sense for zap_residual to be a character string with options "off", "dyn-only", "therm-only", "dyn+therm" and "dyn=therm" or something like it?  In other words, merge the control of the dyn and therm zap features somehow.
- The namelist variable names could be cleaner?  dyn_area_min and min_area seems a bit awkward?  how about dyn_area_min and therm_area_min?  Same dyn_mass_min and therm_mass_min or something like it?

I think we should be moving towards whatever we envision as the final implementation, even if that means changing answers.  I do think it would be great to change answers only once if we can.  If that involves more testing of the area/mass parameters then I understand that may not be ideal at the moment.  Can we setup the namelist names and options in final form and make an initial adjustment to those parameters now, then keep on the table a more formal tuning exercise TBD?

eclare108213:
@apcraig I like that solution.  The impact on users who do not want their answers to change (yet) is that they will have to revert the two dynamics parameters to the original values and also set the new zapping parameters to 0, all in namelist.  And since they're having to change the dynamics parameters anyhow, the `zap_residual_ice` flag is redundant. @NickSzapiro-NOAA requested a flag to turn off the new zapping without having to mess with the parameters, and so I'd like his feedback on this.  Nick? Or @DeniseWorthen?

apcraig:
OK, testing continues.  A full test suite of standalone Icepack on Derecho shows everything is bit-for-bit and passes.  On the other hand, all CICE tests are NOT bit-for-bit except some box and unit tests.  There is another problem, about 10% of the C-grid or CD-grid tests are aborting with "bad departure points" or "negative area (open water)" errors.   So, should we increase the area/mass min values for the C-grid to make sure those tests all pass again?  I will do some tests to see what values might allow all the C and CD tests to pass in case we want to go that way.

apcraig:
> Is it a gx1 or gx3 test that fails then?

Both grids have cases that pass and fail.  But I think most of the tests are too short to really assess the sensitivity in detail.  Only the C-grid tests fail in our test suite, and the longer gx1 test results above suggest the C-grid requires zap parameters that are several orders of magnitude larger to be stable.

JFLemieux73:
I am trying to set up the gx3 test with your code @eclare108213. I am using residual_dynmin for CICE and iresidual2 for icepack. The compilation fails. I am probably doing something wrong with the git submodule. @eclare108213 could you please help me with that?

eclare108213:
Thanks for looking at this, @JFLemieux73.  While the C-grid issues are being worked out, I'm inclined to set the min values back to the current (large) defaults just for C-grid tests, document that, and move on with this PR.  Is everyone okay with that?

JFLemieux73:
The same problem can happen with open water replacing the land cell:

C2

There could be a relatively large shear stress (sigma12) causing the instability.

JFLemieux73:
If we look at the same case for the B-grid we have:

B2

The concentration at the U point is large compared to the C-grid case as the 4 values around it are used for the averaging.

eclare108213:
The icepack update didn't work... not sure why there are conflicts showing up here, or even what they are, but the icepack update seems to be bringing in a bunch of old stuff.  I'll close this PR and create a new one, as recommended in the git workflow guide.
@JFLemieux73 please move the C-grid discussion to a separate issue.  We'll fix that in a different PR when you're ready. Glad you have a good lead on it!

Issue #1068, opened 2025-11-28
Update frz_onset and mlt_onset restart and history capability


apcraig:
frz_onset and mlt_onset do not restart properly and they are unable to restart bit-for-bit in history files.  This is a long standing issue that we should correct.  I believe the main solution is to add these fields to the restart file, but additional assessments may be needed.

dabail10:
Actually, frz_onset is on the restart, but mlt_onset is not. Not sure why it was done this way. We should definitely add it.

PR #1069, opened 2025-11-28
Add history restart capability

### LINKED ISSUE #777: Some history file additional functionality.
- [ ] I have been asked to add the capability to for restart history files. This means that the model would write "interim" average files, so that if say the model stops in the middle of the month then one can restart the monthly history averaging exactly. It is fairly complicated, but should be straightforward.

- [ ] Another functionality is to change some of the time metadata names. This might need a CESMCOUPLED ifdef.

- [X] Finally, adding the option to have the timestamp of the history files to be the middle of the averaging period instead of the end. (see #839)

apcraig:
## PR checklist
    Add history restart capability
- [X] Developer(s): 
    apcraig
    Results are bit-for-bit, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#5766dbd0c59523d58d267d0502cfe5fa6789d5bb, ran a quick_suite, base_suite, io_suite, and unittest_suite on 6 derecho compilers.
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes, histall10d option that tests history restart
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Add history restart to netcdf and pio IO options.  Binary was not included due to the complexity of having to track history fields in binary files.  History restart files are written automatically for history streams that are averaged and when a restart is written during the middle of a history accumulation period.  There is one history restart file per history stream.  File are written in the restart directory using the history name, an appended "_r[histfreq]", and the model date.  An ice_read_hist subroutine was added to the ice_history_write.F90 file.  For binary, calling this returns with a warning message that history restarts are not implemented.  When history restarts are read, the model will only read files and fields that are found and continue with the accumulator initialized to zero for fields that are not found.  For production runs, this should work fine.  If a user modifies the history streams in the middle or a run, then an assessment should be made of which fields are valid on the first restart run.

The history restart files are basically history files, written at double precision, writing the accumulated fields.  In addition, some additional fields are written including time_beg, avgct, albcnt, and snwcnt which represent accumulation counters for time average history output.

A new histall10d set_nml option was added that turns on 3 averaged history streams and all history fields.  When used in a restart test, the scripts will verify bit-for-bit history files and history restart files across the restart.  Several tests were added to the io_suite to include formal testing of bit-for-bit history restarts.  Two fields, mlt_onset and frz_onset and not turned on with histall10d because they do not restart properly and they are unable to restart bit-for-bit on the history file, see #1068.

Several history fields have a bug in them and have been written out incorrectly, and these bugs were fixed.  The bug in these cases was that the fields were accumulated during the timestep across categories but were not zeroed out at the start of the timestep.   As a result, those fields were accumulating over the entire run incorrectly.  The fields that had to be zeroed out were evaps and evaps plus upNO, upNH, bTiz, bphi, iDi, and iki associated with bgc.  The bit-for-bit history restart test discovered these errors.

Add a new namelist, write_histrest, to turn off history restart writing.  The default is that history restarts are on.

Update set_nml.cmip to fix an error in f_apond_ai setting.

Closes #777.

### DIFF
--- a/cicecore/cicedyn/analysis/ice_history.F90
+++ b/cicecore/cicedyn/analysis/ice_history.F90
@@ -2206,7 +2206,7 @@
       use ice_grid, only: tmask, lmask_n, lmask_s, dxU, dyU, grid_ice
       use ice_calendar, only: new_year, write_history, write_restart, &
                               write_ic, timesecs, histfreq, nstreams, mmonth, &
-                              new_month
+                              new_month, write_histrest
       use ice_dyn_eap, only: a11, a12, e11, e12, e22, s11, s12, s22, &
           yieldstress11, yieldstress12, yieldstress22
       use ice_dyn_shared, only: kdyn, principal_stress
@@ -4041,9 +4041,9 @@
       ! write history restarts
       !---------------------------------------------------------------
 
-      if (write_restart == 1) then
+      if (write_histrest .and. write_restart == 1) then
         ! turn on histrest features
-        write_histrest = .true.
+        write_histrest_now = .true.
 
         ! write history restarts
         call ice_timer_start(timer_readwrite)  ! reading/writing
@@ -4056,7 +4056,7 @@
         call ice_timer_stop(timer_readwrite)  ! reading/writing
 
         ! turn off histrest features
-        write_histrest = .false.
+        write_histrest_now = .false.
       endif
 
       end subroutine accum_hist
--- a/cicecore/cicedyn/analysis/ice_history_shared.F90
+++ b/cicecore/cicedyn/analysis/ice_history_shared.F90
@@ -39,7 +39,7 @@
       integer (kind=int_kind), public :: history_precision
 
       logical (kind=log_kind), public :: &
-         write_histrest = .false. , & ! true when writing history restarts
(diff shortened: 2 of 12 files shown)


cicecore/cicedyn/analysis/ice_history.F90
```diff
@@ -4245,6 +4245,28 @@ subroutine accum_hist (dt)
       enddo                     ! iblk
       !$OMP END PARALLEL DO
 
+      !---------------------------------------------------------------
+      ! write history restarts
+      !---------------------------------------------------------------
+
+      if (write_restart == 1) then
+        ! turn on histrest features
+        write_histrest = .true.
```
NickSzapiro-NOAA: If I understand @apcraig, can there be a way for write_histrest to be optional?

UFS may write several prognostic restarts where history restarts are not needed ... to not incur associated I/O cost

apcraig: Lots of good points, thanks @NickSzapiro-NOAA.  I will add a flag to turn off the history restarts, that's easy enough to do and is a good idea.

PR #1114, opened 2026-06-18
plot output from box tests

eclare108213:
script modifications for plotting output from box tests
- [x] Developer(s): 
    @eclare108213 
    ENTER INFORMATION HERE 
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [x] Yes
        - [ ] No 

Added a python script to make simple 2-dimensional plots of output from box tests, in conjunction with the slotted cylinder tests shown in https://github.com/CICE-Consortium/CICE/issues/1072.  The new python script closely mimics the original one, which only made plots on global lat/lon grids.  I also altered the filenames and logic in `ciceplots.csh`.  The current version here shows what was needed for my box plots in the user-defined`files`, `notes`,  and `fstr` data.  I'd be happy to comment mine out in favor of the original, since (I think) it's used to create plots when we make a release, but there must be a better way to do this.  Suggestions are most welcome...

I added the `advection` namelist option to `set_nml.boxslotcyl`, since that test is aimed at advection and I wanted to check upwind too.  The code itself should still be BFB, including slotted cylinder tests.

### DIFF
--- a/configuration/scripts/ciceplots.csh
+++ b/configuration/scripts/ciceplots.csh
@@ -2,14 +2,14 @@
 
 # User defined stuff
 
-#source ${MODULESHOME}/init/csh
+source ${MODULESHOME}/init/csh
 
 # Set plot type (box, global, timeseries)
 # Set case and case directory
 # Set files, notes, fstr, and fields
 
-#set plotgrid = "global"
-set plotgrid = "box"
+set plotgrid = "global"
+#set plotgrid = "box"
 #set plotgrid = "none"
 echo "plotgrid = ${plotgrid}"
 
--- a/configuration/scripts/ciceplots.csh
+++ b/configuration/scripts/ciceplots.csh
@@ -1,6 +1,13 @@
 #!/bin/csh -f
 
-# User defined stuff
+# This script can be modified as needed.
+# By default, the script is set up to create March and September plots for the global case:
+#   ./cice.setup -m derecho -e intel -g gx1 -p 128x1 -s gx1prod,long,run8year --case casename
+# The script can be edited to plot output from box cases, e.g.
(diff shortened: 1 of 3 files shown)


apcraig (changes requested):
I would move gridtype and plottimeseries (lines 56-62) settings up to the top of the script. Then I would add an if block after that for global or box to set the case, casedir, files, notes, and fstrs. And could we set the default to 'global' and 'true'. These are needed for making release plots, I think it would be good to have that be the default. Otherwise, this seems fine overall. Are you happy with the box plots or need any help otherwise?

eclare108213:
I rearranged `ciceplots.csh` so that the settings for both global and box grids are available, and set the defaults for the global grids.  In my setup, `${MODULESHOME}` is not defined (should it be?), so I still have to comment that out.  Does that source command need to be at the top, or can I also move it into the 'global' block?  

The only remaining weirdness for box plots is that `timeseries.py` labels the box output as 'Arctic' (the code uses NH longitudes), and the 'Antarctic' lines are all zero.  That's not optimal, but for this purpose it's good enough in my opinion, rather than adding yet another script specific for labeling box timeseries.

@apcraig please, would you test this version in your regular setup to make sure I haven't broken something?

apcraig:
Sorry for the delay, will try to find some time to review and test next week.

apcraig (commented):
This looks good.  My only comment would be to add some info about how to setup the "standard" global and box cases in the ciceplots.csh script.  Somewhere near the top, add comments like

```
# script can be modified as needed.  by default, the script is setup to create plots for the following cases:
# global case 
#  ./cice.setup -m derecho -e intel -g gx1 -p 128x1 -s gx1prod,long,run8year --case casename
# box case
#  ./cice.setup (fill this part in)
```

apcraig (approved):

PR #1116, opened 2026-06-19
Use /usr/bin/env -S csh -f instead of /bin/csh -f

nsf-name:
    Use /usr/bin/env -S csh -f instead of /bin/csh -f
- [x] Developer(s): 
    @nsf-name
    (No need for a test, doesn't change any Fortran)
    - [x] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
    - [x] No
    - [ ] Yes
        - [x] Yes
        - [ ] No 

This commit changes the `csh` interpreter path from `/bin/csh` to `/usr/bin/env -S csh` in all cases where it's not unambiguously clear what machine CICE is running on. This makes all scripts more portable, as [`env` loads the first `csh` that appears in the path](https://en.wikipedia.org/wiki/Env), rather than assuming it exists at `/bin/csh` which might not be true on all systems. 

Note that `env -S` is not in POSIX; however, Linux, FreeBSD, and macOS have had the nonstandard `-S` option to `env` since it was added to the [GNU coreutils in July of 2018](https://savannah.gnu.org/forum/forum.php?forum_id=9187) and to [FreeBSD (and macOS, since they share coreutils) in November 2005](https://www.freebsd.org/releases/6.0R/relnotes-i386/). This breaks compatibility with systems shipping an old version of the coreutils, but Repology indicates that basically all systems with extant package repos [have this minimum version by now](https://repology.org/project/coreutils/versions). This also breaks out of the box compatibility with OpenBSD and Linux distributions that use BusyBox instead of the coreutils (like Alpine Linux), although users of those systems can install the GNU coreutils through a package or by compiling it themselves, and those OSes are likely rare in HPC environments where CICE is run. You'd also need the GNU coreutils for building `gfortran` anyways on those machines, so it's not an unreasonable dependency. Also, just in case any of the machines don't have this version, I didn't change any of the machine-specific scripts. 

Given that HPC environments don't let users install arbitrary system packages, I believe this change ensures forward compatibility for future distros that may not ship `csh` by default (for example, the AlmaLinux 10 HPC cluster at Williams College doesn't), as it is an uncommon shell for daily use. This change makes it easier for users to build their own version of `csh` and use it with CICE, without breaking backwards compatibility for existing callers (except in the case of old/uncommon systems like I mentioned above).

### DIFF
--- a/.github/workflows/write_logfiles.csh
+++ b/.github/workflows/write_logfiles.csh
@@ -1,4 +1,4 @@
-#!/usr/bin/env csh -f
+#!/usr/bin/env -S csh -f
 
 #echo "hello"
 
--- a/.github/workflows/write_logfiles.csh
+++ b/.github/workflows/write_logfiles.csh
@@ -1,4 +1,4 @@
-#!/bin/csh 
+#!/usr/bin/env csh -f
 
 #echo "hello"
 
--- a/cice.setup
+++ b/cice.setup
@@ -1,4 +1,4 @@
-#!/usr/bin/env csh -f
+#!/usr/bin/env -S csh -f
 
 #set pd0 = `date -u "+%s%N"`
 
@@ -501,7 +501,7 @@ else
   cp -f ${ICE_SCRIPTS}/tests/poll_queue.csh ${tsdir}
 
   cat >! ${tsdir}/suite.submit << EOF0
-#!/usr/bin/env csh -f
+#!/usr/bin/env -S csh -f
 
 set nonomatch && rm -f ciceexe.* && unset nonomatch
 rm -f suite.jobs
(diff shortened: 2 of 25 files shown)


apcraig:
Thanks @nsf-name.  I think this is probably a good idea.  I want to test on several platforms just to gain some confidence before we merge.

anton-seaice (approved):
Works good for us (on gadi) - we also seem to prefer `/usr/bin/env` in other contexts

eclare108213 (commented):
I don't have a way to test this across platforms, so if @apcraig's tests do not turn up any problems, then these changes are fine with me.

mhrib:
Works fine at DMI (aurora+boreas).

Consider doing similar for all BASH scripts - also works at DMI:
`#!/usr/bin/env -S bash -f`

NickSzapiro-NOAA:
Tagging @rgrumbine , if want to check on additional platforms

PR #1118, opened 2026-06-24
Minor fixes in variable truncation and uninitialized value from code analyses

NickSzapiro-NOAA:
For detailed information about submitting Pull Requests (PRs) to the CICE-Consortium,
please refer to: <https://github.com/CICE-Consortium/About-Us/wiki/Resource-Index#information-for-developers>

## PR checklist
- [x] Short (1 sentence) summary of your PR: 
    Minor fixes for variable truncation and uninitialized value identified by code analyses/review
- [x] Developer(s): 
    Nick Szapiro
- [x] Suggest PR reviewers from list in the column to the right.
- [ ] Please copy the PR test results link or provide a summary of testing completed below.
    ufs-weather-model RTs.  Standalone CICE tests, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#998957ffd4d09cc4ef8695b8faed3d98987b2745
- How much do the PR code changes differ from the unmodified code? 
    - [x] bit for bit
    - [ ] different at roundoff level
    - [ ] more substantial 
- Does this PR create or have dependencies on Icepack or any other models?
    - [ ] Yes
    - [x] No
- Does this PR update the Icepack submodule?  If so, the Icepack submodule must point to a hash on Icepack's main branch.
    - [ ] Yes
    - [x] No
- Does this PR add any new test cases?
    - [ ] Yes
    - [x] No
- Is the documentation being updated? ("Documentation" includes information on the wiki or in the .rst files from doc/source/, which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/. A test build of the technical docs will be performed as part of the PR testing.)
    - [ ] Yes
    - [x] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [x] No 
- [x] Please document the changes in detail, including _why_ the changes are made.  This will become part of the PR commit log.
Close https://github.com/CICE-Consortium/CICE/issues/1108 https://github.com/CICE-Consortium/CICE/issues/1112 https://github.com/CICE-Consortium/CICE/issues/1119

PR #1122, opened 2026-07-07
Fix uninitialized fill values in ice_read_write.F90

### LINKED ISSUE #1112: Minor possibly uninitialized variable in ice_read_nc_xy diagnostic
In testing memory related issues with valgrind in a coupled UFS configuration, @DusanJovic-NOAA got traceback including
[bout_124411_cice.txt](https://github.com/user-attachments/files/28601462/bout_124411_cice5.txt) like
```
==2366748== Conditional jump or move depends on uninitialised value(s)
==2366748==    at 0x51254E0: __ice_read_write_MOD_ice_read_nc_xy (ice_read_write.F90:1234)
==2366748==    by 0x50B7938: __ice_grid_MOD_kmtmask (ice_grid.F90:926)
==2366748==    by 0x50B89CA: __ice_grid_MOD_init_grid2 (ice_grid.F90:616)
==2366748==    by 0x4C74283: __ice_comp_nuopc_MOD_initializeadvertise (ice_comp_nuopc.F90:793)
```

These are all related to:
https://github.com/CICE-Consortium/CICE/blob/8e3ef7c4cb657705ceff5bfec3e12b49dec4973e/cicecore/cicedyn/infrastructure/ice_read_write.F90#L1228-L1243

Digging into netcdf, it seems attribute is undefined if it does not exist (as intent(out) that is not set). See
https://github.com/Unidata/netcdf-fortran/blob/f8dc3e9ffdc3d94ae04b8a320fd8ac312c83663f/fortran/netcdf_attributes.F90#L357
https://github.com/Unidata/netcdf-c/blob/6b0fc10e1249ba927ea6b95ac44c1bb0ae73bdd0/libdispatch/dattget.c#L453-L460

Seems like should only set missingvalue="_FillValue" iff the attribute exists

Also noting that GNU compiler doesn't catch all uninitialized variables, maybe particularly into "library" codes.

apcraig:
## PR checklist
    Fix uninitialized fill values in ice_read_write.F90
- [X] Developer(s): 
    apcraig
    Testing on derecho with 4 compilers looks fine, https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#efbbbe2e0c2792ce5ebb5056bbe8c59a222f0ae6
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [ ] Yes
        - [ ] Yes
        - [X] No 

Closes #1112 

This should complete the fixes to the uninitialized fill variables.  A first fix was applied with #1118.  This fixes a couple other instances.

### DIFF
--- a/cicecore/cicedyn/infrastructure/ice_read_write.F90
+++ b/cicecore/cicedyn/infrastructure/ice_read_write.F90
@@ -1416,10 +1416,10 @@
 !            write(nu_diag,*) subname,' Dim name = ',trim(dimname),', size = ',dimlen
 !         enddo
          ! optional
-         missingvalue = spval_dbl
          status = nf90_get_att(fid, varid, "_FillValue", missingvalue)
 !          call ice_check_nc(status, subname//' ERROR: Missing _FillValue', &
 !                            file=__FILE__, line=__LINE__)
+         if (status /= nf90_noerr) missingvalue = spval_dbl
          allocate(mask(nx,ny))
          do n=1,ncat
             if ( ieee_is_nan(missingvalue) ) then
@@ -1619,10 +1619,10 @@
             write(nu_diag,*) subname,' Dim name = ',trim(dimname),', size = ',dimlen
          enddo
          ! optional
-         missingvalue = spval_dbl
          status = nf90_get_att(fid, varid, "_FillValue", missingvalue)
 !          call ice_check_nc(status, subname//' ERROR: Missing _FillValue', &
 !                            file=__FILE__, line=__LINE__)
+         if (status /= nf90_noerr) missingvalue = spval_dbl
          allocate(mask(nx,ny))
          do n=1,ncat
             if ( ieee_is_nan(missingvalue) ) then
--- a/cicecore/cicedyn/infrastructure/ice_read_write.F90
+++ b/cicecore/cicedyn/infrastructure/ice_read_write.F90
@@ -1215,21 +1215,19 @@
       !-------------------------------------------------------------------
 
       if (my_task==master_task .and. diag) then
-           write(nu_diag,'(2a,i8,a,i8,2a)') &
-             subname,' fid= ',fid, ', lnrec = ',lnrec, &
-             ', varname = ',trim(varname)
-!          status = nf90_inquire(fid, nDimensions=ndim, nVariables=nvar)
-!          write(nu_diag,*) subname,' ndim= ',ndim,', nvar= ',nvar
-!          do id=1,ndim
+         write(nu_diag,'(2a,i8,a,i8,2a)') &
+            subname,' fid= ',fid, ', lnrec = ',lnrec, &
+            ', varname = ',trim(varname)
+!         status = nf90_inquire(fid, nDimensions=ndim, nVariables=nvar)
+!         write(nu_diag,*) subname,' ndim= ',ndim,', nvar= ',nvar
+!         do id=1,ndim
 !            status = nf90_inquire_dimension(fid,id,name=dimname,len=dimlen)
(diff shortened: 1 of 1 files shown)


apcraig:
I will merge this once it's reviewed and I've completed and documented the test results which are currently underway.

eclare108213 (approved):

cicecore/cicedyn/infrastructure/ice_read_write.F90
```diff
@@ -1416,10 +1416,10 @@ subroutine ice_read_nc_xyz(fid,  nrec,  varname, work,  diag, &
 !            write(nu_diag,*) subname,' Dim name = ',trim(dimname),', size = ',dimlen
 !         enddo
          ! optional
-         missingvalue = spval_dbl
          status = nf90_get_att(fid, varid, "_FillValue", missingvalue)
 !          call ice_check_nc(status, subname//' ERROR: Missing _FillValue', &
 !                            file=__FILE__, line=__LINE__)
+         if (status /= nf90_noerr) missingvalue = spval_dbl
```
anton-seaice, line 1418: My view is we should be checking against `nf90_enotatt`/_Attribute not found_ rather than `noerr` but i've not won that argument before. So assuming I don't, I think the commented out 

`!          call ice_check_nc(status...` 

can be removed as we are checking the status code using a different method

apcraig: Regarding nf90_enotatt, I agree we could (and maybe should) use that check, but I think it also comes with some questions.  What if status doesn't return nf90_noerr and it doesn't return nf90_enotatt.  In other words, the get_att fails, but not with an nf90_enotatt.  Would we need some logic like

```
  if (status == nf90_enotatt) then
    missingvalue = spval_dbl
  else
    call ice_check_nc(status....   ! which would abort if the get_att had a different error code?
  endif
```

Should the code fail if get_att fails but not with nf90_enotatt?

anton-seaice: I think your suggestion is correct, I can't think of any corner cases where that won't work

anton-seaice: Oh - are the error codes fully backward compatible with old versions of netcdf ?

apcraig: Looks like nf_enotatt was added in netcdf3.0 around 1996 and then was included from the start in the nf90 standard.  So I think that error code is safe to use.

I guess the question is still.  Do we want get_att for the fill value to revert to spval_dbl only when the attribute can't be found and abort on any other error or do we want get_att for the fill to revert to spval_dble on any error?

apcraig:
Testing is complete and looks fine.

cicecore/cicedyn/infrastructure/ice_read_write.F90
anton-seaice: I guess my argument for doing the first isn't very strong. Most of the other errors in https://github.com/Unidata/netcdf-fortran/blob/f1d8fa58acb440f39732f1f199ce0e36f48596f7/fortran/netcdf_constants.F90#L85 would be unlikely unless someone was doing code changes. Doing the first does prevents an error having flow on impacts, e.g. it would be impossible that the data set to FillValue ends up getting interpreted as real data. But there's a pretty limited set of scenarios where that's possible anyway.

Now that i've written it down, it all sounds fairly moot so go with what's convenient I guess.

PR #1126, opened 2026-07-31
Remove some unused variables flagged by compiler warnings

NickSzapiro-NOAA:
For detailed information about submitting Pull Requests (PRs) to the CICE-Consortium,
please refer to: <https://github.com/CICE-Consortium/About-Us/wiki/Resource-Index#information-for-developers>

## PR checklist
- [x] Short (1 sentence) summary of your PR: 
    Remove some unused variables flagged by compiler warnings
- [x] Developer(s): 
    Nick Szapiro
- [ ] Suggest PR reviewers from list in the column to the right.
- [ ] Please copy the PR test results link or provide a summary of testing completed below.
    CICE and ufs-weather-model CI
- How much do the PR code changes differ from the unmodified code? 
    - [x] bit for bit
    - [ ] different at roundoff level
    - [ ] more substantial 
- Does this PR create or have dependencies on Icepack or any other models?
    - [ ] Yes
    - [x] No
- Does this PR update the Icepack submodule?  If so, the Icepack submodule must point to a hash on Icepack's main branch.
    - [ ] Yes
    - [x] No
- Does this PR add any new test cases?
    - [ ] Yes
    - [x] No
- Is the documentation being updated? ("Documentation" includes information on the wiki or in the .rst files from doc/source/, which are used to create the online technical docs at https://readthedocs.org/projects/cice-consortium-cice/. A test build of the technical docs will be performed as part of the PR testing.)
    - [ ] Yes
    - [x] No, does the documentation need to be updated at a later time?
        - [ ] Yes
        - [x] No 
- [ ] Please document the changes in detail, including _why_ the changes are made.  This will become part of the PR commit log.

Unused variables in code can obscure what the code is doing (like declaring local i,ilo,... block loop variables w/o an actual loop in the code) and lead to compiler warnings (that can block pull requests in host models). But, compiler warnings are not errors and "fixing" them should do no harm. Here we remove a number of local variables that are obviously unused, for clearer code and ~300 fewer warnings. Variables that are unused but mentioned in comments are still kept

Issue #1061, opened 2025-10-24
PIO support for restart_ext feature


apcraig:
The PIO restart implementation (cicecore/cicedyn/infrastructure/io/io_pio2/ice_restart.F90) does not seem to support the restart_ext feature.  That option is ignored and haloUpdates are done when a field is read.

This is inconsistent with the netcdf and binary implementations where no halo updates are ever done and the restart_ext is supported.

Very few test cases (and probably users) use the restart_ext feature, and it's never tested with PIO in our test suites.  As we implement regional boundary conditions, this may become more important.  We should also evaluate the restart_ext test cases to make sure we have a rigorous test.  While we are testing that feature with netcdf, it's not clear there is any significant data on the outer boundary in those tests.

anton-seaice:
I had guessed the the `restart_ext` was initially related to the orca grid implementation which we've deprected. I have a feeling restar_ext is used when coupled with NEMO ? (Although i'm fairly confident ECCC don't use PIO) @phil-blain

dabail10:
I agree that we might need to use this. This was implemented as a restart option for tripole grids where we would not have to worry about synchronizing across the tripole seam by explicitly saving the ghost cells.

apcraig:
I'm going to submit a PR that will refactor some of the grid/index/halo operations to support regional boundary conditions to better support the ability to set values on the outer boundary.  The restart_ext feature may be needed or at least useful to allow exact restart for those types of cases.

phil-blain:
> I had guessed the the `restart_ext` was initially related to the orca grid implementation which we've deprected. I have a feeling restar_ext is used when coupled with NEMO ? (Although i'm fairly confident ECCC don't use PIO) [@phil-blain](https://github.com/phil-blain)

Hi Anton, sorry I was on parental leave oct-nov. 

From what I understand, no, the `restart_ext` is not related to the orca grid, but to the tripole grid in general. `restart_ext` is not used in our NEMO-CICE systems. And indeed we no dot use PIO.

apcraig:
Had a quick look to see if this fix is easy.  I think there are actually several changes that have to be made to get extended restarts to work in pio including the io descriptors.  Will see if I can get this to work.

PR #1127, opened 2026-08-10
Merge infrastructure/comm mpi and serial code

apcraig:

## PR checklist
    Merge infrastructure/comm mpi and serial code
- [X] Developer(s): 
    apcraig
    All results are bit-for-bit and all results pass as expected.  https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#7050dac59fd0cbccb8534ed4461d0267db9009e4
https://github.com/CICE-Consortium/Test-Results/wiki/cice_by_hash_forks#a9ffc866ea4a9bd27d59c7d1de85c7d397170c45
    - [X] bit for bit
    - [ ] more substantial 
    - [ ] Yes
    - [X] No
    - [ ] Yes
    - [X] No
    - [X] Yes
    - [ ] No
    - [X] Yes
        - [ ] Yes
        - [ ] No 

Merged the comm/serial and comm/mpi directories to the comm directory.  This may impact CICE builds in coupled systems.  Add CPP, NO_MPI, to turn off MPI.

Synchronized the comm/serial and comm/mpi directories moving all files to the comm directory and removing the serial and mpi directories.  Added a CPP, NO_MPI, to turn off MPI in the comm code.  The scripts build and run works the same, instead of choosing a serial or mpi directory, it always grabs the code in the comm directory and sets -DNO_MPI on the compile if serial is being used.

Refactored the ice_gather_scatter.F90 file.  

- Removed the _ext methods by adding an optional argument, grid_ext, to the baseline gather and scatter methods to indicate use of an extended grid field in the call.
- Added a generate_GatherScatter.sh script to generate several overloaded methods quickly.  Updated the README.generate_code in the comm directory to provide documentation of script usage.
- Eliminated scatter_global_stress.  Where this was used, the global read and scatter_global_stress was replaced by an ice_read_global that scatters plus a call to ice_HaloUpdate_stress.
- Updated the i_global (i_glob) and j_global (j_glob) arrays in ice_blocks.F90 to reflect the grid indexing WITHOUT boundary conditions.  The extra logic and special tripole values have been removed.  The halo information is now computed directly in the gather/scatter infrastructure and the tripole logic is triggered by the value of ns_boundary_type.
- Ensured all results were bit-for-bit.  Some issues and bugs are documented separately and may need to be addressed at a later time.

Updated comm infrastructure to consistently use fillValue as the optional argument to specify fill values.

Update comm and IO infrastructure to consistently use grid_ext as the optional argument to specify use of an extended grid field.

Added a gather/scatter unittest and updated the unittest_suite.

Updated some of the halochk unittest implementation.

Added and updated some of the namelist option files to support various mixed boundary conditions better.

Moved ns_boundary_type and ew_boundary_type from ice_domain to ice_blocks to provide access to these variables to the CICE infrastructure.

Cleaned up some unused variables in the comm files.

Fixed an error in the namelist settings in set_nml.bdyrestornest required for the full/nest testing.  This is unrelated to the gather/scatter comm cleanup of this PR.

Updated the documentation.

A significant amount of work was put in to understand the gather/scatter logic and to try to clean up the code (for instance, replace an extended grid scatter with a grid scatter plus a haloUpdate), but problems were constantly created.  The model seems to depend on some of the undocumented features (for instance, the extended scatter uses a fill value of zero by default and if this is changed, the code breaks), and there are other issues that could be fixed as well.

Getting the gather/scatter unit test to work was incredibly challenging and that unit test has a huge number of special cases to validate the checking.  In particular, the gather/scatter must validate for any combination of

- double, real, integer, logical datatypes
- all grids
- all boundary conditions including cyclic, open, closed, tripole, tripoleT, zero_gradient, linear_extrap
- center, north, east, and northeast points
- scaler, vector, and angle fields
- with and without land block elimination especially at the boundary edge or in the corner
- with fill values specified or not
- in cases where the outer halo is specifically not updated
- for extended and non-extended global grids (both for gather and scatter)

The grid scatter updates the halo values.  In the interior, this is relatively trivial.  On the outer boundary, this should behave like a haloUpdate, but it doesn't match perfectly.  Attempts to replace the scatter with a scatter and a haloUpdate caused havoc.

Several issues were identified.  Attempts were made to fix some of them, but this generally led to more model failures.  So for now, the gather/scatter update is bit-for-bit with the prior version even if there has been significant cleanup.  I decided that keeping the code bit-for-bit for this PR was important.  Fixing issues and introducing non bit-for-bit changes can be added incrementally in the future as needed and documented carefully.  This cleanup should make later fixes more tangible.  The gather/scatter unit tester also provides a comprehensive testing capability that can be used as future changes are made.  Some of the issues are

- The haloUpdate fills the local halos including the outside halo depending on the value of the boundary conditions (ew/ns_boundary_type).  The haloUpdate also performs a tripole line synchronization on the zipper which has nothing to do with the haloUpdate.  The scatter method does NOT perform this tripole line synchronization at ny_global.  Ideally, the tripole line synchronization would be removed from the haloUpdate and be called as a separate infrastructure method.  Then the scatter and the haloUpdate should be updated to perform the same computations or even better, the haloUpdate aspect of the scatter should be removed from the scatter and a haloUpdate should be called when needed with a scatter call.
- The grid scatter has a default fill value of zero.  If this is changed to spval (or similar), the code breaks.  I suspect this has to do with the grid scatters that occur in the grid initialization, but it was difficult to identify exactly which fields and which calls are problematic.
- There is some ambiguity whether gridpoints that are not set in the infrastructure (i.e. open and closed bcs) should take on the fill value or keep the values that were passed into the infrastructure.  Different methods may behave differently, this should be clarified.
- Some features of the haloUpdate and scatter methods do not make sense for logicals on the tripole grid (i.e. averaging on the zipper).  This should be clarified.
- The haloUpdate_stress call updates the zipper for the tripoleT but not for the tripole.  I think the tripoleT implementation is incorrect and is a bug in the code.
- If there is land block elimination at the top of the grid for tripole grids, the scatter sets the value to the fill value but then flips the sign of that fill value if the field is "angle" or "vector".  This should not happen for fill values and could cause problems later if that specific fill value is used to check for a fill.
- It's not clear the NW and NE outer halo corners with a tripole grid are handled properly.  I think they are automatically doing an east-west cyclic boundary condition (which is normally how the code is run with tripole grids), but this is not strictly required and is a gap in the ability of the implementation to handle tripole north-south boundary conditions with an arbitrary east-west boundary condition.


cicecore/cicedyn/infrastructure/comm/generate_HaloUpdate.sh
```diff
@@ -90,9 +90,9 @@ cat <<EOFF > $file
                                fillValue, tripoleOnly)
 
 !  Generated by ${0} on ${cdate}
-!  This is autogenerated so may have some extra code, like k and l
+!  This is generated from a script so may have some extra code, like k and l
 !  loops of length 1 or extra size 1 dimensions in arrays.  This is
-!  done to simply code generation and does not seem to impact performance.
+!  done to simpifycode generation and does not seem to impact performance.
```
eclare108213: typo - needs a space

cicecore/cicedyn/infrastructure/comm/README.generate_code
```diff
@@ -0,0 +1,70 @@
...
+   ./generate_HaloUpdate.sh
+   ./generate_GatherScatter.sh
+
+  ! Verify generate_HaloUpdate.sh produces code in HaloUpdate_all.subr that matches the current 
+  ! implementation in ice_boundary.F90.  If it doesn't, it suggests someone made manual changes to 
+  ! the subroutines in ice_boundary.F90.  Try to reconcile those differences and update 
+  ! generate_HaloUpdate.sh first, then move forward.
+  ! Do the same for generate_GatherScatter, GatherScatter_all.subr, and ice_gather_scatter.F90.
+
+  ! Now copy the current generate output.
```
eclare108213: Suggestion:  remove the word "generate" here.

cicecore/cicedyn/infrastructure/ice_blocks.F90
```diff
@@ -140,7 +149,8 @@ subroutine create_blocks(nx_global, ny_global, ew_boundary_type, &
    integer (int_kind) :: &
       i, j, n              ,&! loop indices
       iblock, jblock       ,&! block loop indices
-      is, ie, js, je         ! temp start, end indices
+      iindg, jindg         ,&! i,j global index
+      is, js                 ! temp start, end indices
```
eclare108213: typo:  remove the word "end"

cicecore/cicedyn/infrastructure/ice_restart_driver.F90
```diff
@@ -991,52 +988,66 @@ subroutine restartfile_v4 (ice_ic)
...
+
+         call ice_HaloUpdate_stress(stress12_1, stress12_3, halo_info, &
+                                    field_loc_center,  field_type_scalar)
+         call ice_HaloUpdate_stress(stress12_3, stress12_1, halo_info, &
+                                    field_loc_center,  field_type_scalar)
+         call ice_HaloUpdate_stress(stress12_2, stress12_4, halo_info, &
+                                    field_loc_center,  field_type_scalar)
+         call ice_HaloUpdate_stress(stress12_4, stress12_2, halo_info, &
+                                    field_loc_center,  field_type_scalar)
+         ! TODO: CD-grid
```
eclare108213, line 1049: This is a major refactoring!
Does the code still run properly with the CD grid?

doc/source/developer_guide/dg_infra.rst
```diff
@@ -47,17 +47,14 @@ The time manager was updated in early 2021.  Additional information about the ti
...
-serial directories are compiled with CICE, not both.
-
-**cicedyn/infrastructure/comm/mpi/**
-is based on MPI and provides various methods to do halo updates, global sums, gather/scatter, broadcasts
-and similar using some fairly generic interfaces to isolate the MPI calls in the code.
-
-**cicedyn/infrastructure/comm/serial/** support the same interfaces, but operates
-in shared memory mode with no MPI.  The serial library will be used, by default in the CICE scripts,
-if the number of MPI tasks is set to 1.  The serial library allows the model to be run on a single
+A low-level communications package, that supports both mpi and serial, are provided as part of CICE.  This software
```
eclare108213: fix grammar:  A low-level communications package, which supports both mpi and serial, is provided as part of CICE.

doc/source/developer_guide/dg_infra.rst
```diff
@@ -47,17 +47,14 @@ The time manager was updated in early 2021.  Additional information about the ti
...
+provides a middle layer between the model and the underlying libraries.  
+
+**cicedyn/infrastructure/comm/**
+provides  various methods to do halo updates, global sums, gather/scatter, broadcasts
+and similar feature using some fairly generic interfaces to isolate the MPI calls in the code.
```
eclare108213: replace "feature" with "tasks"

cicecore/drivers/unittest/halochk/halochk.F90
```diff
@@ -524,10 +759,19 @@ program halochk
...
+               cjchk_bas = cjchk
+
+               if (debugpts .and. this_block%j_glob(j) == jc .and. k1==k1c .and. k2==k2c) then
+                  write(100+my_task,'(a,4i4,2f12.3)') 'dp11',i,j,k1,k2,aichk,ajchk
+                  write(100+my_task,'(a,6i4,2f12.3)') 'dp12',i,j,k1,k2,this_block%i_glob(i),this_block%j_glob(j)
+                  write(100+my_task,'(a,4i4,2f12.3)') 'dp13',i,j,k1,k2,cichk,cjchk
+               endif
 
                ! halo special cases
+               ! this has to be done after cidata_gl2 is used because they are specific values and don't want to add k1 and k2
```
eclare108213: some compilers used to complain about apostrophes in comment statements.  I don't understand the comment following the "and" -- is that part necessary/helpful?

eclare108213 (commented):
Just reading through the code changes, this all looks great to me.  Impressive!  I'll wait to approve until someone running in a coupled configuration has a chance to test/review it.

cicecore/cicedyn/infrastructure/ice_restart_driver.F90
apcraig: The full test suite is bit-for-bit.  This includes a large set of tests focused on the C and CD grids.

apcraig:
I fixed the suggested changes proposed so far.

dabail10 (approved):
I tested this within the regional CESM configuration and aside from having to fix the comm path it went smoothly. I did not test for bfb as this was already done extensively.

apcraig:
Thanks @dabail10, great to hear.

NickSzapiro-NOAA (approved):
Coupled ufs-weather-model regression tests for https://github.com/CICE-Consortium/CICE/pull/1127 https://github.com/CICE-Consortium/CICE/pull/1126 are bit-for-bit on Ursa

apcraig:
@NickSzapiro-NOAA, thanks for running some tests.  That's super helpful.

cicecore/cicedyn/infrastructure/io/io_netcdf/ice_restart.F90
anton-seaice, line 1: is there a seperate issue already for restart_ext support with pio ?

cicecore/cicedyn/infrastructure/ice_blocks.F90
```diff
@@ -179,32 +189,45 @@ subroutine create_blocks(nx_global, ny_global, ew_boundary_type, &
 !----------------------------------------------------------------------
 
    if (.not.allocated(all_blocks)) allocate(all_blocks(nblocks_tot))
-   if (.not.allocated(i_global)) allocate(i_global(nx_block,nblocks_tot))
-   if (.not.allocated(j_global)) allocate(j_global(ny_block,nblocks_tot))
+   if (.not.allocated(i_glob_data)) allocate(i_glob_data(nx_block,nblocks_tot))
+   if (.not.allocated(j_glob_data)) allocate(j_glob_data(ny_block,nblocks_tot))
```
anton-seaice, line 193: sorry if this is covered elsewhere. What are these (`i/j_glob_data`) used for ? Or are these to support new features ?

cicecore/cicedyn/infrastructure/ice_blocks.F90
```diff
@@ -226,88 +249,42 @@ subroutine create_blocks(nx_global, ny_global, ew_boundary_type, &
...
-               case ('cyclic')
-                  j_global(j,n) = j_global(j,n) + ny_global
-               case default
-                  ! lower to upper
-               end select
+            jindg = js - nghost + j - 1
+            if (jindg >= 1-nghost .and. jindg <= ny_global+nghost) then
+                j_glob_data(j,n) = j_global(jindg)  ! simple lower to upper counting
+            else
+                j_glob_data(j,n) = spval_int  ! padding
```
anton-seaice, line 256: So this change means that the ghost / padded cells no longer have global indeces, just fill values.

( I guess these indeces weren't used for anything ?)

anton-seaice (approved):
Looks great Tony, thanks for adding the unit test.

This is bit for bit with ACCESS-OM3 too.

do the test cases cover tripoleT and coupled wind stress (calc_strair=.false.) ?

cicecore/cicedyn/infrastructure/io/io_netcdf/ice_restart.F90
apcraig: Yes, #1061

cicecore/cicedyn/infrastructure/ice_blocks.F90
apcraig: It's a good question.  They are used to define the global index that each gridpoint is associated with in the i and j direction.  Previously, that partly included some info about the bcs, but the implementation was built up over time and was not really extensible.  With cyclic, the outer halo reflected the cyclic indices, for tripole, the index was negative, padded gridcells were set to zero.  Clearly, this was built over time and became problematic in terms of providing a general extensible implementation.  The main place this was used was in the scatter routines.  The halo update routines do not use these arrays.

Anyway, I redefined the indices to just represent the actual global index with the reference point (1,1) being the lower left corner of the active grid.  So the lower and west outer halo have indices <= 0 and the right and upper outer halo have indices > nx/ny_global.  The padded indices are set to spval_int now and no info is contained in the indices about the boundary conditions.

Then the scatter was refactored to NOT use the global arrays to scatter the halo at all, especially to the outer boundary.  It now uses the ew/ns_boundary_type values and fills in the outer halo with new logic that is much more general and extensible, but that also produces bit-for-bit results with the prior version.

Where these global indices are used for the active gridpoints (not the outer halo or padding), the values are the same.  Again, the "special" values for the global arrays were on the outer halo only and that was used primarily in the scatter routine.

cicecore/cicedyn/infrastructure/ice_blocks.F90
apcraig: They actually never had global values.  The prior implemented set padded values to zero explicitly.  Now they are set to spval_int.  See some further discussion above.

apcraig:
I'd like to merge this by this weekend unless there are any other comments or concerns.

eclare108213 (approved):

cicecore/cicedyn/infrastructure/ice_blocks.F90
anton-seaice: Thanks - whats the distinction between `i/j_glob_data` and `i/j_global` ?

apcraig: Good question.

Short answer, they contain the same info, i_global is an array that is global in size and i_glob_data is an array associated with a "CICE" block of data.  Read long answer below if you want to know more.

i_global is now a 1d "global" array that holds all the indices in the i direction from 1-nghost to nx_global+nghost.  The value of that array is exactly 1-nghost to nx_global+nghost.  it's a fairly trivial array at this point.

```
   if (.not.allocated(i_global)) allocate(i_global(1-nghost:nx_global+nghost))
   do i = 1-nghost, nx_global+nghost
      i_global(i) = i
   enddo
```

i_glob_data is an array associated with each block and the value of it matches the value of i_global at the same point.  It's i_global "copied" to an array that's associated with each block.  So in the code block below, loop over every block (iblock, jblock), count blocs (n), copy i_global to i_glob_data for that block.  Then, that data is pointed to from inside the "all_blocks" datatype.

```
   if (.not.allocated(i_glob_data)) allocate(i_glob_data(nx_block,nblocks_tot))
   do jblock=1,nblocks_y
      do iblock=1,nblocks_x
         n = n + 1  ! global block id                                                                                           

         do i=1,nx_block
            iindg = is - nghost + i - 1
            if (iindg >= 1-nghost .and. iindg <= nx_global+nghost) then
                i_glob_data(i,n) = i_global(iindg)  ! simple lower to upper counting                                            
            else
                i_glob_data(i,n) = spval_int  ! padding                                                                         
            endif
         enddo
         all_blocks(n)%i_glob => i_glob_data(:,n)

     enddo
  enddo
```

All of those arrays (and their equivalent j arrays) exist on all pes for the entire global grid.  This is all computed after the blocks are created but before they are distributed.  Again, all pes have all the same info about all blocks and all gridcells.  In addition, i_glob_data is private to the module, but all_blocks is public.  So i_glob_data is just a temporary so all_block(n)%i_glob has allocated associated with it.

With this update, these arrays and a lot simpler to understand.  Whether they are in a global or block array, they are just the index of the global array.

Eventually, the blocks get distributed and each pe has a hidden list of global block ids that it owns.  Each pe then gets the block info by querying the all_blocks datatype.  So, the block memory usage does not scale, but this is how it's been since CICEv0.  And it's why the model is constantly doing 

```
         do iblk = 1, nblocks
            this_block = get_block(blocks_ice(iblk),iblk)
            ilo = this_block%ilo
            ...
```

get block returns a pointer to all_blocks(globalID(iblk)) where globalID(iblk) returns a different value on each pe.

anton-seaice: Oh awesome, thankyou !

anton-seaice: I don't find the names super clear, but also don't have good suggestions for them. The concept makes sense though,

Feel free to go ahead with the merge :)

